PR #3, opened 2015-04-13
Oracle database integration test with Docker

sheikhsameer1609:
Hi Simon, 
Please find the added functionality for database integration testing with Docker and Type safe configuration library.

Regards,
Sameer Sheikh


io/plugins/eu.esdihumboldt.hale.io.jdbc.test/src/eu/esdihumboldt/hale/io/jdbc/test/AbstractDBTest.java
```diff
@@ -0,0 +1,277 @@
...
+
+	public static final String DB_UPTIME = ".dbUPTime";
+	private DBImageParameters dbi;
+	private DBDockerClient tdc;
+	public static final String ORACLE_DB = "ORCL";
+
+	/**
+	 * Setup host and database.
+	 */
+	public void setupDB(DBImageParameters dbi) {
```
stempler: Why not pass the DBImageParameters in the test constructor?
Then you would not have to create a specific setupDB method in every test, but could define it in AbstractDBTest (including the Before annotation)

io/plugins/eu.esdihumboldt.hale.io.jdbc.test/src/eu/esdihumboldt/hale/io/jdbc/test/AbstractDBTest.java
```diff
@@ -0,0 +1,277 @@
...
+			ri.close();
+		}
+
+		return count;
+	}
+
+	/**
+	 * Shutdown database and host.
+	 */
+	public void tearDownDocker() {
```
stempler: When is this method executed?

io/plugins/eu.esdihumboldt.hale.io.jdbc.test/src/eu/esdihumboldt/hale/io/jdbc/test/DBDockerClient.java
```diff
@@ -0,0 +1,174 @@
...
+
+			dc.startContainer(containerId, hostConfig);
+			
+			final ContainerInfo info = dc.inspectContainer(containerId);
+			portMapper = info.networkSettings().ports();
+			
+
+		} catch (DockerException | InterruptedException e) {
+			
+			e.printStackTrace();
```
stempler: Doing this kind of default exception handling is a bad practice (this is the same for other similar code in the pull request as well). Here is not the place where to decide what to do in case of failure.

Starting the container is something that may fail, and the tests using the container should not be executed as they rely on it.

io/plugins/eu.esdihumboldt.hale.io.jdbc.test/src/eu/esdihumboldt/hale/io/jdbc/test/DBImageParameters.java
```diff
@@ -0,0 +1,96 @@
+package eu.esdihumboldt.hale.io.jdbc.test;
+
+import java.net.URI;
+
+
+public class DBImageParameters {
```
stempler: I would much prefer to have an interface instead of this class for this purpose.

For the following reasons:
* instead of just providing configuration data, methods can be used to allow more flexible configuration (e.g. for building the JDBC URL, in some cases the "Start URL" may not be sufficient)
* the configuration object should not be mutable
* if we decide later to do other tests than database tests with Docker, we can extract part of the interface that is relevant for Docker as well a the corresponding test logic and use it for both kind of tests

io/plugins/eu.esdihumboldt.hale.io.jdbc.test/src/eu/esdihumboldt/hale/io/jdbc/test/DockerConfig.java
```diff
@@ -0,0 +1,92 @@
+package eu.esdihumboldt.hale.io.jdbc.test;
+
+import java.io.File;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import com.typesafe.config.Config;
+import com.typesafe.config.ConfigFactory;
+
+public class DockerConfig {
+	public static final String DOCKER_CONF = "docker.conf";
```
stempler: I would prefer to not use a filename like `docker.conf` - this is too prone to conflicts with something else related to Docker.
I would suggest something that is more specific, like `hale-docker.conf` or something like that. An option could also be sticking to `overcast.conf` if we keep a format that is compatible to their configuration.

io/plugins/eu.esdihumboldt.hale.io.jdbc.test/src/eu/esdihumboldt/hale/io/jdbc/test/DBImageParameters.java
```diff
@@ -0,0 +1,96 @@
...
+	
+	/**
+	 * The database user name
+	 */
+	private String user ;
```
stempler: If we now have a custom configuration for the Docker images it would be good if we could configure things like user, password and database also through the configuration file - as these are settings that are closely related to the Docker image itself, not the type of database.

stempler:
Created Gerrit change from pull request:
http://gerrit.esdi-community.eu/#/c/3814/

Will be merged if it passes the automated and manual testing. May need changes to the build server to be able to execute the DB integration tests.

PR #4, opened 2015-04-24
Test for checking sql type and binding

sheikhsameer1609:
Hello Simon,

Created the unit test case for testing sql type and expected binding of data types for postgresql.

Regards,
Sameer Sheikh

### DIFF
--- a/common/plugins/eu.esdihumboldt.hale.common.instance/src/eu/esdihumboldt/hale/common/instance/model/InstanceUtil.java
+++ b/common/plugins/eu.esdihumboldt.hale.common.instance/src/eu/esdihumboldt/hale/common/instance/model/InstanceUtil.java
@@ -16,6 +16,7 @@
 
 package eu.esdihumboldt.hale.common.instance.model;
 
+import java.math.BigDecimal;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -87,7 +88,6 @@ public final class InstanceUtil {
 			if (o1.getClass().isArray() && o2.getClass().isArray()) {
 				return arrayToList(o1).equals(arrayToList(o2));
 			}
-
 			// special case: geometry properties
 			if (o1 instanceof GeometryProperty<?> && o2 instanceof GeometryProperty<?>) {
 				GeometryProperty<?> g1 = (GeometryProperty<?>) o1;
@@ -251,6 +251,11 @@ public final class InstanceUtil {
 		else if (a instanceof Group) {
 			if (b instanceof Group && groupEqual((Group) a, (Group) b, propertyOrderRelevant))
 				return true;
+		}
+		else if (a instanceof BigDecimal && b instanceof BigDecimal) {
+			BigDecimal x = (BigDecimal) a;
+			BigDecimal y = (BigDecimal) b;
+			return x.compareTo(y) == 0;
 		}
 		else if (equals(a, b))
 			return true;
--- a/common/plugins/eu.esdihumboldt.hale.common.instance/src/eu/esdihumboldt/hale/common/instance/model/InstanceUtil.java
+++ b/common/plugins/eu.esdihumboldt.hale.common.instance/src/eu/esdihumboldt/hale/common/instance/model/InstanceUtil.java
@@ -252,6 +252,8 @@ public final class InstanceUtil {
 			if (b instanceof Group && groupEqual((Group) a, (Group) b, propertyOrderRelevant))
 				return true;
 		}
+		// Two BigDecimal objects that are equal in value but have a different
+		// scale (like 2.0 and 2.00) should consider as equal.
 		else if (a instanceof BigDecimal && b instanceof BigDecimal) {
 			BigDecimal x = (BigDecimal) a;
 			BigDecimal y = (BigDecimal) b;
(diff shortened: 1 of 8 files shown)


common/plugins/eu.esdihumboldt.hale.common.instance/src/eu/esdihumboldt/hale/common/instance/model/InstanceUtil.java
```diff
@@ -252,6 +252,11 @@ else if (a instanceof Group) {
 			if (b instanceof Group && groupEqual((Group) a, (Group) b, propertyOrderRelevant))
 				return true;
 		}
+		else if (a instanceof BigDecimal && b instanceof BigDecimal) {
+			BigDecimal x = (BigDecimal) a;
+			BigDecimal y = (BigDecimal) b;
+			return x.compareTo(y) == 0;
```
stempler: Would be nice if you could provide s short comment here why you do a special case handling for BigDecimal and how it makes a difference.

io/plugins/eu.esdihumboldt.hale.io.jdbc.test/src/eu/esdihumboldt/hale/io/jdbc/test/AbstractDBTest.java
```diff
@@ -74,8 +86,8 @@ public AbstractDBTest(DBImageParameters imageParams) {
 	/**
 	 * Setup host and database.
 	 * 
-	 * @throws InterruptedException
-	 * @throws DockerException
+	 * @throws InterruptedException InterruptedException
+	 * @throws DockerException DockerException
```
stempler: Please provide a meaningful documentation for throws and param or leave it empty (also in other places). Just repeating the Exception type or the parameter class name does not provide any benefit.

For throws it should reflect the reason(s) why a call would fail with that exception, e.g. "if the configuration of the docker client failed" or "if the docker image could not be started".

io/plugins/eu.esdihumboldt.hale.io.jdbc.test/src/eu/esdihumboldt/hale/io/jdbc/test/AbstractDBTest.java
```diff
@@ -185,6 +201,38 @@ protected Schema readSchema() throws Exception {
 	}
 
 	/**
+	 * @param map binding map
+	 * @param schema schema
+	 * @throws Exception exception
+	 */
+	protected void checkBidingAndSqlType(Schema schema, Map<String, Class<?>> map) throws Exception {
```
stempler: Please fix the spelling of the method name.

For the binding map, please provide a more detailed documentation (what is mapped to what?), so users of this method will know exactly what they have to provide.

stempler:
Created Gerrit change from pull request, pending build and tests:
http://git.esdi-community.eu/3855

io/plugins/eu.esdihumboldt.hale.io.jdbc.test/src/eu/esdihumboldt/hale/io/jdbc/test/AbstractDBTest.java
```diff
@@ -94,15 +113,20 @@ public void setupDB() throws DockerException, InterruptedException {
 			Thread.currentThread().setContextClassLoader(cl);
 		}
 
-		jdbcUri = URI.create(dbi.getJDBCURL(client.getHostPort(dbi.getDBPort()),
-				client.getHostName()));
-
+		jdbcUri = URI.create(dbi.getJDBCURL(
+				client.getHostPort(dbi.getDBPort()), client.getHostName()));
+		jdbcUri = URI.create("jdbc:oracle:thin:@//localhost:1521/demoDB");
```
stempler: This seems to be testing code that should not be in the commit.

io/plugins/eu.esdihumboldt.hale.io.jdbc.test/src/eu/esdihumboldt/hale/io/jdbc/test/AbstractDBTest.java
```diff
@@ -94,15 +113,20 @@ public void setupDB() throws DockerException, InterruptedException {
...
+		TestUtil.startConversionService();
+		TestUtil.startService(
+				Arrays.asList("eu.esdihumboldt.hale.io.jdbc.oracle"),
```
stempler: This would have to be done in an Oracle specific test class.
Also, it fails if the bundle is not available (which is the case during tests on the build server).

io/plugins/eu.esdihumboldt.hale.io.jdbc.postgresql.test/src/eu/esdihumboldt/hale/io/jdbc/postgresql/test/PostDataTypesTest.groovy
```diff
@@ -0,0 +1,137 @@
...
+	void testDataTypes(){
+		Sql s = new Sql(waitForConnection());
+		try{
+			s.execute query;
+		}
+		finally{
+			s.close();
+		}
+		Schema schema = readSchema()
+		checkBidingAndSqlType(schema,map);
```
stempler: You fixed the spelling at one point, but not here - which results in the test failing.

PR #5, opened 2015-05-04
Reproject geometry function

mbarto:
Adds a ReprojectGeometry transformation. This allows an alignment project to declare a specific SRS and reproject a source geometry if its SRS is different from the declared one.


stempler:
Thanks a lot. I have a few comments on the changes which I will add inline.

In general I understand that you added this function to be able to reproject the geometry in cases where it is not supported on export.

Regarding the way the user specifies the CRS, did you consider using a control similar to the `TargetCRSConfigurationPage` that is used for instance on the XML/GML export? It allows to specify the CRS as WKT as well, and provides slightly better feedback if the user enters an unrecognized code.

common/plugins/eu.esdihumboldt.hale.common.test/Tests.product
```diff
@@ -203,6 +211,7 @@
       <plugin id="org.apache.commons.io"/>
       <plugin id="org.apache.commons.jxpath"/>
       <plugin id="org.apache.commons.lang"/>
+      <plugin id="org.apache.commons.logging"/>
```
stempler: The only plugin that needs to be added here is `eu.esdihumboldt.cst.functions.geometric.test`. It does not introduce any new dependencies, so the other additions should be removed.

The neat option in Eclipse to *Add required plugins* is a bit of a pitfall here, as it usually will add plugins and fragments that are not desired - as with `org.apache.commons.logging`, which adds a second Apache Commons Logging implementation in addition to the existing one.

cst/plugins/eu.esdihumboldt.cst.functions.geometric.test/src/eu/esdihumboldt/cst/functions/geometric/TransformationReproject.java
```diff
@@ -0,0 +1,227 @@
...
+import eu.esdihumboldt.hale.common.instance.model.InstanceReference;
+import eu.esdihumboldt.hale.common.instance.model.ResourceIterator;
+import eu.esdihumboldt.hale.common.instance.model.ext.InstanceIterator;
+import eu.esdihumboldt.hale.common.instance.model.impl.FilteredInstanceCollection;
+import eu.esdihumboldt.hale.common.instance.model.impl.PseudoInstanceReference;
+import eu.esdihumboldt.hale.common.schema.model.Schema;
+import eu.esdihumboldt.hale.common.schema.model.TypeDefinition;
+import eu.esdihumboldt.hale.common.test.TestUtil;
+
+public class TransformationReproject implements TransformationExample{
```
stempler: As far as I see the functionality in here is more or less the same as in the existing `TransformationExample` implementation `InternalExample`. As it is not reusable like that in another plugin, I extracted a base class (`TransformationExampleImpl`) from it, that you should be able to use and thus get rid of most of this file's content, so we avoid any duplication.

The class mentioned was added recently in 4a349f770727e11097fbc867572d316cda8b15b0.

cst/plugins/eu.esdihumboldt.cst.functions.geometric/plugin.xml
```diff
@@ -275,6 +275,51 @@
...
+                  minOccurrence="1"
+                  name="target">
+               <geometryOrParentCondition></geometryOrParentCondition>
+            </property>
+         </targetProperties>
+         <sourceProperties>
+            <property
+                  description="The property that shall hold the input geometry"
+                  eager="false"
+                  label="Source geoemtry"
```
stempler: minor misspelling

ui/plugins/eu.esdihumboldt.hale.ui.functions.geometric/src/eu/esdihumboldt/hale/ui/functions/geometric/CoordinateReferenceSystemParameterPage.java
```diff
@@ -0,0 +1,136 @@
...
+		
+		String crsAsString = null;
+		if (crsDef != null) {
+			if (crsDef instanceof CodeDefinition) {
+				crsAsString = ((CodeDefinition)crsDef).getCode();
+			} else if (crsDef instanceof WKTDefinition) {
+				crsAsString = ((WKTDefinition)crsDef).getWkt();
+			}
+		}
+		conf.put(PARAMETER_REFERENCE_SYSTEM, new ParameterValue(crsAsString));
```
stempler: Just a minor improvement suggestion here - you can easily create a String representation for a `CRSDefinition` using the `CRSDefinitionManager`. Advantage is that you don't have to care about the encoding or any possible additional kinds of CRSDefinitions that might be added in the future. A corresponding call would look like this:

```
String crsAsString = CRSDefinitionManager.getInstance().asString(crsDef))
```

When reading the parameter in the function implementation, you can use it as well:

```
CRSDefinition crsDef = CRSDefinitionManager.getInstance().parse(crsString)
```

ridethepenguin: Thanks for the suggestion, I wasn't aware of the existence and purpose of `CRSDefinitionManager`. I implemented the change and the code runs fine in the HALE product, but for some reason `ReprojectGeometryTest` fails when the Tests product is run. After debugging, I noticed that `CRSDefinitionManager.getInstance().parse(crsString)` returns null, even if `crsString = "EPSG:4326"`. I guess there is something wrong with the lookup of OSGi services in `CRSDefinitionManager.getDefinitions()`, but since I'm a complete beginner in OSGi development I can't figure out what... any ideas?

stempler: That's because through this the syntax of the string representation changes slightly. The `CRSDefinitionManager` will also add a marker to be able to identify which kind of `CRSDefinition` it should create when parsing.

In this case, the parameter value in the Alignment file must be `code:EPSG:4326` - because the Alignment file was created with the old UI, it does not use this new representation. So it's good to switch to the `CRSDefinitionManager` now while we don't have to deal with backwards compatibility issues.

ridethepenguin: Of course! So shameful I didn't realize this earlier... Anyway, I fixed the test data and committed the changes: I hope everything is ok now! :-)

stempler:
Created Gerrit change from pull request
http://git.esdi-community.eu/3872

Let's see what the build server has to say :smile: 

Are there any others apart from you that I should add to the contributors list in the HALE help after this is merged?

ridethepenguin:
> Are there any others apart from you that I should add to the
> contributors list in the HALE help after this is merged?


Yes, please include also Sandro Salari, the original author of the code.
Thanks! :-)

PR #6, opened 2015-05-12
Add SpatiaLite r/w support

ridethepenguin:
SpatiaLite specific implementations of `SchemaReader`, `InstanceReader` and `InstanceWriter` interfaces, wrapping and customising the generic JDBC implementations.


stempler:
What prerequisites are needed to load the SpatialLite extension? Does it need to be present on the Java library path?
Currently I'm getting this error:
`[SQLITE_ERROR] SQL error or missing database (mod_spatialite.so: cannot open shared object file: No such file or directory)`

ridethepenguin:
You need to copy SpatiaLite's native library (with dependencies) in a folder where the dynamic linker can find them (e.g. `C:\Windows\System32` on Windows, `/usr/lib` on Linux). For Windows, you can find binaries [here](http://www.gaia-gis.it/gaia-sins/windows-bin-x86-test/mod_spatialite-4.2.1-rc1-win-x86.7z) (32-bit) and [here](http://www.gaia-gis.it/gaia-sins/windows-bin-amd64-test/mod_spatialite-4.2.1-rc1-win-amd64.7z) (64-bit). Would be wonderful to find a way to package them with the plugin...

thorsten-reitz:
It should be possible to distribute the dlls with HALE without requiring to put them in the Windows/System directory:
https://msdn.microsoft.com/en-us/library/7d83bc18.aspx

stempler:
@ridethepenguin thanks for the hint - I managed to get it running in Windows and Linux both. It would be really nice to be able to package the libraries, but I am not sure how that should be done.

*Another question:*

We did some tests with other SpatialLite files and encountered problems with the geometries - is this maybe due to the SpatialLite version?
Please find a sample here: http://download.osgeo.org/livedvd/data/spatialite/

The error that I get is
```
[SQLITE_ERROR] SQL error or missing database (no such column: geometry_type)
```
which results in the geometry property not being properly classified by the schema reader and no CRS information to be available for the data.

ridethepenguin:
@thorsten-reitz so, having the DLLs in the same dir as the HALE executable should work on Windows... but on Linux? AFAIK we would need to play with the `LD_LIBRARY_PATH` envirnoment variable to make it work...

@stempler yes, the current implementation is based off SpatiaLite 4.+, so this is indeed a wrong version issue: https://www.gaia-gis.it/fossil/libspatialite/wiki?name=switching-to-4.0. With a bit of extra effort, I guess I could make the code work also with older versions of SpatiaLite.

thorsten-reitz:
@ridethepenguin You are right about Linux. So you wouldn't recommend doing something like adding the containing directory to LD_LIBRARY_PATH before launching the application?

    export LD_LIBRARY_PATH = $LD_LIBRARY_PATH:/some/pathOfContainingDirectory

I don't have a lot of experience with spatialite, but noticed that most samples that are online are of older databases. Do you think missing support for older version is a deal-breaker for customers?

@stempler The least we have to do is to properly handle older databases (give user feedback on limitations or deny loading).

ridethepenguin:
@thorsten-reitz Absolutely, re-defining `LD_LIBRARY_PATH` as you showed is perfectly fine, but it's still something that the user should do "manually", so IMHO not a huge improvement over instructing him/her to copy she required *.so somewhere. Or, we could provide a bash script that would: 1) define LD_LIBRARY_PATH properly; 2) call HALE's executable, and tell users to use that instead of the native launcher on Linux... or is there a way to have the native launcher itself set the environment library?

Regarding SpatiaLite versions, I have no statistical basis to claim older (i.e. < 4.0) versions of SpatiaLite are rare: I focused on version 4.+ because that's what our current customer is using. Anyway, as I said, making the plugin compatible with older SL versions doesn't seem to be a big deal: I'll take a shot at it ASAP.

ridethepenguin:
@thorsten-reitz @stempler Could you please have a look at the updated code? It *should* work with any SpatiaLite database, no matter which version.

stempler:
Thanks a lot @ridethepenguin !
I'm afraid we will only be able to take a closer look next week when we are back from the INSPIRE conference.

ridethepenguin:
@stempler no problem at all! I won't have much time to work on it 'til next week either. Thanks for taking the time to review my code!

stempler:
Thanks again @ridethepenguin for your contribution.
Your changes have been merged into master. Only exception is 36799dc because we had a fix for that applied already in the meantime (see 9c2def660be79a96a9fd5d106d5efcbda325d701).

We plan to release a version 2.9.3 that includes the SpatialLite support by the end of the month. For Windows, the DLLs needed for mod_spatialite will be packaged with the application, Linux and Mac users will have to provide/build the libraries themselves.

stempler:
A short summary on adaptions/changes we did after merging the code.
@ridethepenguin Any comments are welcome.

* Added basic help (11842d0fc6302bbcbbd09e3d07a2a68f41f87e40)
* Don't include the file path in the namespaces for type definitions representing the tables (217388f11b6d10df83c1c244d36dbebd7da4f3e6)
* Notify the user when files with geometries are used but **mod_spatialite** could not be loaded (804214cfb2f89778f9ad37efc60bd3bee86ddfb4)
* Changed the way geometries are retrieved form the database to correctly support reading 3D geometries (88d6923f0b8003505c5a75ddb03607c8a96b1099)
* Don't use a java.sql.Blob binding for SQLite blobs (1b21782601b307d678434c43f283f8892270383c)

ridethepenguin:
@stempler It's fantastic to see SpatiaLite support in HALE! :smile: Thanks for the time you spent in merging and testing.

I've added a few comments, hope you will find them useful.

PR #7, opened 2015-07-14
GeoServer app-schema configurator plugin

ridethepenguin:
The pull request contains an extension to translate HALE alignments to an app-schema configuration, which can be added to a GeoServer instance to serve the transformed data via standard OGC services (WMS, WFS, etc...). See [this blog post](http://www.geo-solutions.it/blog/inspire-support-in-geoserver-made-easy-with-hale/) for a more detailed explanation.

The current status of the plugin is experimental and not all features have been implemented. Other features will be hopefully added later on.

Features currently implemented include:
- Transformation functions:
 1. Retype
 2. Merge (partial)
 3. Join (partial)
 4. Assign
 5. Date extraction
 6. Formatted string
 7. Mathematical expression
 8. Rename
 9. Classification
- GeoServer compatibility mode
- Wizard page to configure a single  PostGIS datastore
- Three export formats:
 1. app-schema mapping file only (requires additional configuration in GeoServer)
 2. app-schema configuration archive (includes ancillary configuration files, no additional configuration required)
 3. direct upload via GeoServer's REST API (requires a recent GeoServer build)

The plugin is self-contained and does not require external dependencies. 

A second commit adding the possibility to create a custom alignment export configuration and save it with the project is included in the PR, as it is most useful to avoid entering datastore configuration settings each time the app-schema mapping is exported.


stempler:
@ridethepenguin thanks for the pull request. Looks like a lot of work.

Any chance you will be at the FOSS4GE in Como the next days? So we can discuss it and - that would be great - maybe you could even show to me how it works together with GeoServer.

I only did a quick glance through the code right now. One thing that I noticed is that there is quite a big amount of generated code included. We would like to avoid including generated code in the code base if possible (I have to admit there was a time when we did that - also with Jaxb generated code). If the classes are not available in a (Maven-published?) library already, an option would be adding the Jaxb build to [our existing Jaxb-Bundle builds](https://github.com/igd-geo/hale-platform/tree/master/jaxb)  in [hale-platform](https://github.com/igd-geo/hale-platform) so it can be included in the target platform easily.
Something else related to that: Will the net.opengis... package names of the generated classes interfere with any libraries from the Geotools context?

Another question that comes to my mind is how we should further proceed with your contribution regarding HALE releases (i.e. should we merge it into master or use a feature branch for now). If it shall be included in HALE, even as an experimental feature it should add some documentation to the help, showing how to use it and documenting what the limitations are.

ridethepenguin:
@stempler No, unfortunately I won't be at FOSS4GE, but I'm about to prepare some documentation and a working example showing the app-schema configurator in action: I'll share it with you as soon as it is ready. That may serve also as a starting point for the documentation to be added to HALE's help.

About the generated code, I had a hard time configuring JAXB bindings and I finally managed to do it using the tools from [this project](http://www.ogcnetwork.net/jaxb4ogc). Perhaps the easiest way to avoid adding generated code to HALE's codebase would be to package it in a JAR file and publish it in a Maven repository (perhaps HALE's?), instead of integrating the code generation step in the build process... but I'm open to discussion. Regarding possible conflicts with the net.opengis.* packages: there shouldn't be any so far, since the target platform only includes classes from org.opengis.* packages. However, I can investigate the feasibility of renaming all packages in a way that removes even any remote possibility of conflict (i.e. eu.esdihumboldt.net.opengis.*).

In the next few days, I'll discuss these points with the rest of the team and get back to you, so we can decide how to proceed further.

Thanks, as usual, for your feedback!

stempler:
@ridethepenguin to bad you can't come to the FOSS4GE - would have liked to meet you and the conference being in Italy is probably the best opportunity for that in some time. At least promise me to let me know when you come by Darmstadt or Frankfurt so I can buy you a :beer: 

**About the generated code:**
I was wondering which schema it is that you are using. Peeking in the classes I found a lot of GML stuff, but I'm not sure if that's not just a dependency. I saw the project you mentioned seems to have quite a lot of schemas already as JARs in Maven - so if you did not see that before, maybe [check it out](http://repo1.maven.org/maven2/org/jvnet/ogc/) to see if your schema is already there.

Taking a look at what classes are in HALE already, there actually are `net.opengis...` classes (for Example `net.opengis.wfs20.DescribeFeatureTypeType`) - probably from the same project - or maybe not, as it seems those are (at least partly) EMF models, not Jaxb.

I would be OK with publishing a JAR to the HALE artifactory for the purpose of making it available. However, there should nonetheless be a reproducable way to create that library, to recreate or update it. As an alternative you could also publish it to Bintray (which is probably easier than going for Maven Central directly).

ridethepenguin:
@stempler of course, I'll let you know if I travel to Germany, I could really use some (h)ale :wink: 

**About the generated code (again)**:
I noticed the artifacts published by jaxb4ogc project and for sure classes in `the net.opengis.*` packages can be found there, but what I really need to do here is to map [this schema](https://github.com/geosolutions-it/hale/blob/appschema/io/plugins/eu.esdihumboldt.hale.io.appschema.test/src/data/AppSchemaDataAccess.xsd), which is specific to GeoTools/GeoServer (not an official OGC standard schema, so mappings for it are not readily available from jaxb4ogc project). The `AppSchemaDataAccess.xsd` schema includes a couple of OGC schemas, `expr.xsd` and `feature.xsd`, which is where the `net.opengis.*` classes come from.

Anyway, I will try and make the generation process automatic, reproduceable and conflict-free. Then, I'll probably need your help to integrate it in hale-platform :wink:

ridethepenguin:
@stempler I managed to setup an automated build for the generation of `AppSchemaDataAccess.xsd` Java bindings, it was not as difficult as I initially thought.

I updated the code in this PR and submitted a [second PR](https://github.com/igd-geo/hale-platform/pull/5) against the hale-platform repo. I also renamed the packages of the generated classes, so now name conflicts should be impossible.

stempler:
@ridethepenguin Great, without the Jaxb classes, now I can even inspect the whole diff in Github :smile: 

I hope you don't mind - I made two comments inline on things which sprang to my mind while quickly going over the code.
Looks very good to me so far (at least what I am able to understand), though I may have a few more comments when I get to actually run it.

About the custom Alignment export configurations, I'm not quite sure yet if I am comfortable with that. Originally they were added just for the instance export, specifically for convenient use from the command line by referencing such a preset. So far, for other alignment exports, there was no real need. Saving something like user and password to the project may be critical - this is done already for JDBC schemas and data and that's not ideal as well. Have to think about that.

ridethepenguin:
@stempler Of course I don't mind, your feedback is most welcome!

Regarding the custom alignment export configuration, if I understand you correctly, your real concern is about storing plain text passwords in the project configuration, rather than the functionality itself, am I right? Do you think introducing in HALE a password encryption scheme, [along the lines of what GeoServer does](http://docs.geoserver.org/latest/en/user/security/passwd.html#password-based-encryption), could be a workable solution? Ultimately, I would let the user decide whether passwords should be stored as encrypted or not, to allow for easy project sharing when security is not a concern.

What do you think?

stempler:
Regarding the custom alignment export configuration:

One thing is the plain password storage, the other thing is that export configurations were initially created exclusively for instance export, which is reflected for instance in the `ProjectTransformationEnvironment` class.

A key question is if these settings are more appropriate to be stored on the machine or with the project.

Password encryption is already possible, though not for configurations included in the project file. Eclipse brings its own mechanism for password-based encryption (using a master password). This is in place currently for the proxy user password in the settings.
An alternative approach could be allowing the user to specify a set of GeoServer connections in the application settings, which he can choose from in a wizard page on export.

stempler:
Hi @ridethepenguin ,

after a bit of thinking I came to the conclusion that I would prefer to keep this in a feature branch for now and not merge it directly into master. This is because we may do a release of HALE before the appschema extension is ready for it (e.g. a bugfix release). Also like this we can postpone the discussion on the Alignment export configurations and provide proper integration tests with GeoServer (Docker image?).
When we plan a release for HALE we can still talk about the state of the appschema extension and see if we get it ready in time.

The current state of the pull request is now merged into the [feature/appschema](https://github.com/igd-geo/hale/tree/feature/appschema) branch. Atop of that a few small fixes for bugs I stumbled upon.

I hope this approach is OK for you, please let me know if you have any doubts.

Simon

ridethepenguin:
Hi @stempler,
your approach is sound and allows both of us to move further with the developments without stepping on each other's toes. I will probably need your assistance soon to setup those integration tests.

Thanks for merging!

Stefano

PR #9, opened 2015-10-15
Docker-based integration tests with GeoServer App-Schema

ridethepenguin:
I split this in two commits: the first one adds a method to check the server availability to the `HaleDockerClient` class (you may want to cherry-pick and merge this to master); the second one is the actual integration test code.


stempler:
Thanks for the contribution, @ridethepenguin . I think having these kind of integration tests is important and this can be the base for adding more extensive tests later on, that for instance test the GML that is produced by app-schema with a created mapping.
We separated unit tests (xxxTest classes) and integration tests (xxxIT classes) a while ago (at least for the more expensive kind of integration tests). So for later integration into master I might make this *ResourceManagerIT* and remove the docker availability check, as I think if the tests cannot be run (even if that's probably a build server problem), that should not go unnoticed.

ridethepenguin:
Thanks for merging, @stempler.

Regarding the renaming, no problem, I'll do it immediately. I guess this also removes the need for the availability check: I put it in place as a way to avoid running docker based tests every time I do `build.sh commitStage`, but, if I understand correctly, test classes not ending in *Test are not run, right? Out of curiosity, how did you instruct your build server to run integration tests automatically? I would like to do the same on our own Jenkins ;-)

One last note about checking the GML produced by app-schema. I'm not sure this is something we should test in the app-schema plugin for HALE: I believe our job here is just to make sure the generated app-schema configuration is correct, and trust app-schema to do its job. GeoServer's code base contains a huge test suite checking app-schema's generated GML, we don't want to replicate that in HALE I guess.

stempler:
Regarding the test execution - yes, you understand correctly. Running `build.sh integrationStage` ~~will~~ should run both unit tests and integration tests (just saw that they don't seem to be run currently :warning: )

Regarding the app-schema testing - you are right, most of that should be done in app-schema. I will have to take some time someday and take a closer look at the tests there.

PR #11, opened 2015-10-23
Fix integration tests naming and server availability check

ridethepenguin:
Fixed according to discussion on [previous PR](https://github.com/igd-geo/hale/pull/9).


stempler:
Thanks!
I also merged from master where the fix is included that the integration tests actually run with the integrationStage build :wink:

PR #16, opened 2015-10-30
Semi-automated handling of nillable elements

ridethepenguin:
Implemented compromise solution for the handling of nillable mandatory elements, as [discussed](https://github.com/ridethepenguin/hale/issues/16).

Pending docs update, hope to make it before release time.


ridethepenguin:
Added docs on the mapping of nillable elements, the limitation on the namespace of target types and the restriction to mapping relevant types in feature chaining configuration.

stempler:
Awesome, thanks. I tested with an example I prepared based on an extended INSPIRE Building 2D schema. Simple cases are handled quite well with the changes from this PR, though there remain some issues - I will expand on them in ridethepenguin#16.

PR #17, opened 2015-10-31
Fixed encodeIfEmpty handling for nillable mandatory elements

ridethepenguin:
@stempler This should fix the issue with nested nillable mandatory elements we discussed this morning, am I too late to see it in the release (after your confirmation that it really works... I made a few tests myself but I could really use a double check here)?


stempler:
You changes yielded for me a better result in my test, at least for that aspect. So it's in.

I did a small addendum to the help with things I found to be not supported (or at least only partially supported):

* Priorities for relations (e.g. using a low priority Assign as default value that is overridden by a higher priority mapping)
* Instance contexts for target properties for creating additional instances of properties that can appear multiple times

PR #31, opened 2016-06-30
Join: use soft reference delegates for instances

stempler:


stempler:
Requires #52 to be fixed. Otherwise transformation may fail as a symptom (because instance references are used more often and dynamically here).

PR #35, opened 2016-07-04
Winding order handling for geometry objects

arunverma87:
Winding Order handling for Polygon, MultiPolygon, GeometryCollection, LinearString

### DIFF
--- a/cst/plugins/eu.esdihumboldt.cst/.classpath
+++ b/cst/plugins/eu.esdihumboldt.cst/.classpath
@@ -3,6 +3,5 @@
 	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.7"/>
 	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
 	<classpathentry kind="src" path="src"/>
-	<classpathentry exported="true" kind="lib" path="C:/dev/eclipse/plugins/org.codehaus.groovy_2.3.10.xx-201604081629-e45/lib/groovy-all-2.3.10.jar"/>
 	<classpathentry kind="output" path="bin"/>
 </classpath>
--- a/cst/plugins/eu.esdihumboldt.cst/.classpath
+++ b/cst/plugins/eu.esdihumboldt.cst/.classpath
@@ -3,5 +3,6 @@
 	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.7"/>
 	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
 	<classpathentry kind="src" path="src"/>
+	<classpathentry exported="true" kind="lib" path="C:/dev/eclipse/plugins/org.codehaus.groovy_2.3.10.xx-201604081629-e45/lib/groovy-all-2.3.10.jar"/>
 	<classpathentry kind="output" path="bin"/>
 </classpath>
--- a/util/plugins/eu.esdihumboldt.util.test/src/eu/esdihumboldt/util/geometry/WindingOrderTest.java
+++ b/util/plugins/eu.esdihumboldt.util.test/src/eu/esdihumboldt/util/geometry/WindingOrderTest.java
@@ -41,30 +41,29 @@ public class WindingOrderTest {
 
 	private static GeometryCollection clockWise4;
 
+	private static LinearRing r1, r2, h1, h2;
+
 	@BeforeClass
 	public static void setUp() {
 		GeometryFactory factory = new GeometryFactory();
 
-		LinearRing r1 = factory.createLinearRing(new Coordinate[] { new Coordinate(10, 30),
+		r1 = factory.createLinearRing(new Coordinate[] { new Coordinate(10, 30),
 				new Coordinate(20, 0), new Coordinate(0, 0), new Coordinate(10, 30) });
 
+		r2 = factory.createLinearRing(new Coordinate[] { new Coordinate(49.87445, 8.64729),
+				new Coordinate(49.87582, 8.65441), new Coordinate(49.87095, 8.65694),
+				new Coordinate(49.86978, 8.65032), new Coordinate(49.87197, 8.64758),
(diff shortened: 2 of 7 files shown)


cst/plugins/eu.esdihumboldt.cst/.classpath
```diff
@@ -3,5 +3,6 @@
 	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.7"/>
 	<classpathentry kind="con" path="org.eclipse.pde.core.requiredPlugins"/>
 	<classpathentry kind="src" path="src"/>
+	<classpathentry exported="true" kind="lib" path="C:/dev/eclipse/plugins/org.codehaus.groovy_2.3.10.xx-201604081629-e45/lib/groovy-all-2.3.10.jar"/>
```
stempler: This change should be reverted. It's specific for your system and I don't think you need that anyway to compile the project.

util/plugins/eu.esdihumboldt.util.test/src/eu/esdihumboldt/util/geometry/WindingOrderTest.java
```diff
@@ -0,0 +1,130 @@
...
+		assertFalse(clockWise1.equalsExact(result));
+		assertTrue(WindingOrder.isCounterClockwise(((Polygon) result).getExteriorRing()));
+	}
+
+	@Test
+	public void testUnifyCCWWithHoles() {
+		Geometry result = WindingOrder.unifyWindingOrder(clockWise2, true);
+		assertTrue(result instanceof Polygon);
+		assertFalse(clockWise2.equalsExact(result));
+		assertTrue(WindingOrder.isCounterClockwise(((Polygon) result).getExteriorRing()));
```
stempler: You could check the winding order of the holes here as well.

util/plugins/eu.esdihumboldt.util/src/eu/esdihumboldt/util/geometry/WindingOrder.java
```diff
@@ -0,0 +1,205 @@
...
+import com.vividsolutions.jts.geom.Coordinate;
+import com.vividsolutions.jts.geom.Geometry;
+import com.vividsolutions.jts.geom.GeometryCollection;
+import com.vividsolutions.jts.geom.GeometryFactory;
+import com.vividsolutions.jts.geom.LinearRing;
+import com.vividsolutions.jts.geom.MultiPolygon;
+import com.vividsolutions.jts.geom.Polygon;
+
+/**
+ * TODO Type description
```
stempler: Add a short type description please.

util/plugins/eu.esdihumboldt.util/src/eu/esdihumboldt/util/geometry/WindingOrder.java
```diff
@@ -0,0 +1,205 @@
...
+ * 
+ * @author Arun
+ */
+public class WindingOrder {
+
+	private static GeometryFactory factory = new GeometryFactory();
```
stempler: This should be `final` as it is only set once.

util/plugins/eu.esdihumboldt.util/src/eu/esdihumboldt/util/geometry/WindingOrder.java
```diff
@@ -0,0 +1,205 @@
...
+			return false;
+
+		// Get coordinates of geometry
+		Coordinate[] coordinates = geometry.getCoordinates();
+		// Using algorithm get order of geometry
+		boolean orientation = false;
+		try {
+			orientation = CGAlgorithms.isCCW(coordinates);
+		} catch (IllegalArgumentException ex) {
+			System.err.println("Illegal Argument");
```
stempler: It makes sense to propagate the exception instead of silently ignoring it. Otherwise you will return information that was not checked in any way. Thus a caller would never be sure if the result from the `isCounterClockwise` is correct or if there was an internal error.

util/plugins/eu.esdihumboldt.util/src/eu/esdihumboldt/util/geometry/WindingOrder.java
```diff
@@ -0,0 +1,205 @@
...
+		// Checking and reversing Holes
+		LinearRing holes[] = new LinearRing[poly.getNumInteriorRing()];
+		for (int i = 0; i < poly.getNumInteriorRing(); i++) {
+			holes[i] = unifyWindingOrderForLinearRing((LinearRing) poly.getInteriorRingN(i),
+					!counterClockWise);
+		}
+
+		// Create New Polygon using unified shell and holes
+		Polygon tempPoly = factory.createPolygon(shell,
+				poly.getNumInteriorRing() > 0 ? holes : null);
```
stempler: If you do the check `poly.getNumInteriorRing() > 0` anyway, you could do it earlier and if the condition is met not create an array in the first place.

util/plugins/eu.esdihumboldt.util/src/eu/esdihumboldt/util/geometry/WindingOrder.java
```diff
@@ -0,0 +1,205 @@
...
+			GeometryCollection geoCollection, boolean counterClockWise) {
+
+		// get no of geometries in GeometryCollection
+		int noOfGeoms = geoCollection.getNumGeometries();
+
+		Geometry[] revGeoms = new Geometry[noOfGeoms];
+		// Unify each geometry one by one
+		for (int i = 0; i < noOfGeoms; i++) {
+			Geometry geometry = geoCollection.getGeometryN(i);
+			if (geometry instanceof MultiPolygon) {
```
stempler: Instead of having the conditions for the different geometry types duplicated here, you could just call `unifyWindingOrder(geometry, counterClockWise)` for the individual geometries in the collection.

util/plugins/eu.esdihumboldt.util.test/src/eu/esdihumboldt/util/geometry/WindingOrderTest.java
stempler: You added new tests for that. What I actually meant was that in this test, in addition to testing the shell of the processed polygon you could verify the winding order of the holes of the processed polygon.

stempler:
LGTM

PR #37, opened 2016-07-05
Feature/windingorder

arunverma87:
Made changes in Geo Instance Writer for unifying WindingOrder.


common/plugins/eu.esdihumboldt.hale.common.instance/src/eu/esdihumboldt/hale/common/instance/io/GeoInstanceWriter.java
```diff
@@ -41,6 +43,11 @@
 	public static final String PARAM_CRS_CODE_FORMAT = "crs.epsg.prefix";
 
 	/**
+	 * Name of the parameter specifying a Winding order of Geometry.
+	 */
+	public static final String PARAM_UNIFY_WINDING_ORDER = "counterclockwise";
```
arunverma87: First I took it as a Boolean parameter so i gave that name.. Forgot to change..

common/plugins/eu.esdihumboldt.hale.common.instance/src/eu/esdihumboldt/hale/common/instance/io/impl/AbstractGeoInstanceWriter.java
```diff
@@ -65,6 +67,16 @@ public String getCustomEPSGPrefix() {
...
 	}
 
+	@Override
+	public void setWindingOrder(EnumWindingOrderTypes windingOrderType) {
+		setParameter(PARAM_UNIFY_WINDING_ORDER, Value.of(windingOrderType));
+	}
+
+	@Override
+	public EnumWindingOrderTypes getWindingOrder() {
+		return getParameter(PARAM_UNIFY_WINDING_ORDER).as(EnumWindingOrderTypes.class);
```
stempler: I don't think that will work. I think you have to use the string value (`as(String.class)`) and then use the `valueOf()` method of the enum type.

common/plugins/eu.esdihumboldt.hale.common.instance/src/eu/esdihumboldt/hale/common/instance/io/impl/AbstractGeoInstanceWriter.java
```diff
@@ -65,6 +67,16 @@ public String getCustomEPSGPrefix() {
 		return getParameter(PARAM_CRS_CODE_FORMAT).as(String.class);
 	}
 
+	@Override
+	public void setWindingOrder(EnumWindingOrderTypes windingOrderType) {
+		setParameter(PARAM_UNIFY_WINDING_ORDER, Value.of(windingOrderType));
```
stempler: Note that you can only do that here because the enum value returns its name in `toString()`.

common/plugins/eu.esdihumboldt.hale.common.instance/src/eu/esdihumboldt/hale/common/instance/io/impl/AbstractGeoInstanceWriter.java
stempler: This would be the ideal place to provide a default value, in case the parameter is not set. This should be something that classes extending the writer are able to adapt.
The default should be not doing anything, but for GML writers it should be a unification to CCW.

common/plugins/eu.esdihumboldt.hale.common.instance/src/eu/esdihumboldt/hale/common/instance/io/impl/AbstractGeoInstanceWriter.java
```diff
@@ -139,6 +170,63 @@ else if (value instanceof GeometryProperty<?>) {
...
+			switch (windingOrder) {
+
+			case COUNTERCLOCKWISE:
+				unifiedGeometry = WindingOrder.unifyWindingOrder(geom, true);
+				break;
+			case CLOCKWISE:
+				unifiedGeometry = WindingOrder.unifyWindingOrder(geom, false);
+				break;
+			default:
+				unifiedGeometry = WindingOrder.unifyWindingOrder(geom, true);
```
stempler: This case should never happen, other than the enum is extended. Thus I would suggest one of two things here:

1. Log an error message and don't touch the geometry *or*
2. Throw an exception

No 1 would be the more robust solution, so I would prefer that.

common/plugins/eu.esdihumboldt.hale.common.instance/src/eu/esdihumboldt/hale/common/instance/io/util/EnumWindingOrderTypes.java
```diff
@@ -0,0 +1,25 @@
...
+
+package eu.esdihumboldt.hale.common.instance.io.util;
+
+/**
+ * TODO Type description
+ * 
+ * @author Arun
+ */
+public enum EnumWindingOrderTypes {
+	CLOCKWISE, COUNTERCLOCKWISE, DONTCHANGE;
```
stempler: As the enum names are also the parameter values to be provided to a writer I would prefer names that are not all uppercase. The parameter values for similar settings usually use camel case.
So the settings could be `clockwise`, `counterClockwise` and `noChanges`.

When you create a pull request, please try to eliminate all compiler warnings in Eclipse that are introduced by your changes. I'm quite sure that there are warnings here about missing documentation.
This would be the perfect spot to document the different modes and state that the winding order is applied in reverse to polygon holes.

io/plugins/eu.esdihumboldt.hale.io.gml.test/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriterTest.java
```diff
@@ -670,8 +745,34 @@ private IOReport fillFeatureTest(String elementName, URI targetSchema,
...
+	 * @param skipValueTest if the check for equality shall be skipped
+	 * @param expectWriteFail if the GML writing is expected to fail
+	 * @param windingOrderParam winding order parameter or <code>null</code>
+	 * @return the validation report or the GML writing report if writing
+	 *         expected to fail
+	 * @throws Exception if any error occurs
+	 */
+	private IOReport fillFeatureTest(String elementName, URI targetSchema,
+			Map<List<QName>, Object> values, String testName, String srsName, boolean skipValueTest,
+			boolean expectWriteFail, String windingOrderParam) throws Exception {
```
stempler: As an enum is used for the parameter now, it would be better to use the enum type here, instead of a string.

common/plugins/eu.esdihumboldt.hale.common.instance/src/eu/esdihumboldt/hale/common/instance/io/GeoInstanceWriter.java
stempler: Thought so :smiley:

common/plugins/eu.esdihumboldt.hale.common.instance/src/eu/esdihumboldt/hale/common/instance/io/util/EnumWindingOrderTypes.java
arunverma87: I remember to add JavaDocs on Function but with class/ interface/ Enum. I will remember it now onward. :)

common/plugins/eu.esdihumboldt.hale.common.instance/src/eu/esdihumboldt/hale/common/instance/io/impl/AbstractGeoInstanceWriter.java
```diff
@@ -212,14 +230,17 @@ protected Geometry unifyGeometry(Geometry geom) {
...
 				unifiedGeometry = WindingOrder.unifyWindingOrder(geom, true);
 				break;
-			case CLOCKWISE:
+			case clockwise:
 				unifiedGeometry = WindingOrder.unifyWindingOrder(geom, false);
 				break;
 			default:
-				unifiedGeometry = WindingOrder.unifyWindingOrder(geom, true);
+				if (report != null) {
+					report.error(new IOMessageImpl("WindingOrder is not set", null));
```
stempler: Actually, in that case the winding order parameter would be set, but with a enum value that is not handled here. So it would be good if the message would state that it encountered a parameter it does not know. It could also state the enum value's name.

common/plugins/eu.esdihumboldt.hale.common.instance/src/eu/esdihumboldt/hale/common/instance/io/util/EnumWindingOrderTypes.java
```diff
@@ -16,10 +16,24 @@
...
  * @author Arun
  */
 public enum EnumWindingOrderTypes {
-	CLOCKWISE, COUNTERCLOCKWISE, DONTCHANGE;
+
+	/**
+	 * Unified geometry by Clockwise Winding order to the shell and Counter
+	 * Clockwise to the holes.
+	 */
+	clockwise, /**
```
stempler: A line break before the comment will probably make for a nicer formatting of the comment with the auto-formatter.

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/writer/GmlInstanceWriter.java
```diff
@@ -45,4 +46,12 @@ protected boolean requiresDefaultContainer() {
 		return true; // requires a FeatureCollection being present
 	}
 
+	/**
+	 * @see eu.esdihumboldt.hale.common.instance.io.impl.AbstractGeoInstanceWriter#getDefaultWindingOrder()
+	 */
+	@Override
+	protected EnumWindingOrderTypes getDefaultWindingOrder() {
+		return EnumWindingOrderTypes.counterClockwise;
```
stempler: `StreamGmlWriter` is the base class of GML writers, please move the method there.

util/plugins/eu.esdihumboldt.util/src/eu/esdihumboldt/util/geometry/WindingOrder.java
```diff
@@ -76,7 +77,8 @@ else if (geometry instanceof Polygon) {
 		else if (geometry instanceof LinearRing) {
 			return unifyWindingOrderForLinearRing((LinearRing) geometry, counterClockWise);
 		}
-		else if (geometry instanceof GeometryCollection) {
+		else if (geometry instanceof GeometryCollection
+				&& geometry.getClass() == GeometryCollection.class) {
```
stempler: If you do the class check you can omit the `instanceof` check. But I would prefer if you use equals instead of `==`, because in OSGi you might actually have the same class provided by different class loaders.
So something like this should work in all cases:
`GeometryCollection.class.equals(geometry.getClass())`

io/plugins/eu.esdihumboldt.hale.io.gml.test/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriterTest.java
```diff
@@ -809,7 +922,35 @@ private IOReport fillFeatureTest(String elementName, URI targetSchema,
 					Object value = geoms.iterator().next().getGeometry();
 
 					if (expected instanceof Geometry && value instanceof Geometry) {
-						matchGeometries((Geometry) expected, (Geometry) value);
+
+						// I have to comment below line to test Winding Order.
+						// Below method got assertion failed error.
+						if (windingOrderParam == null) {
```
stempler: Maybe also if the setting for the winding order is to leave things unchanged? The geometry match check should still succeed then.

io/plugins/eu.esdihumboldt.hale.io.gml.test/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriterTest.java
```diff
@@ -809,7 +922,35 @@ private IOReport fillFeatureTest(String elementName, URI targetSchema,
...
+							matchGeometries((Geometry) expected, (Geometry) value);
+						}
+						// Winding Order Test.
+						if (windingOrderParam != null) {
+							if (windingOrderParam == EnumWindingOrderTypes.counterClockwise) {
+								assertFalse(((Geometry) expected).equalsExact((Geometry) value));
```
stempler: This checks makes assumptions about the geometries passed in (that they are not counter-clockwise). I would remove it so there are no unnecessary (and undocumented) restrictions on this method. Just checking if the result complies with the rules is enough.

io/plugins/eu.esdihumboldt.hale.io.gml.test/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriterTest.java
arunverma87: That's true. in CCW it is getting failed. but It is working when Winding Order set to unChanged.

common/plugins/eu.esdihumboldt.hale.common.instance/src/eu/esdihumboldt/hale/common/instance/io/impl/AbstractGeoInstanceWriter.java
arunverma87: Does it make any difference? Because I am getting the Winding Order as "getParameter(PARAM_UNIFY_WINDING_ORDER)
				.as(EnumWindingOrderTypes.class);".

stempler: No, just a remark about the inner workings of Value :smile:

arunverma87: Yes, I just checked it, Simple Value and Complex Value. Thanks.

stempler: Yes, and complex values only work if they is a class registered that handles it serialization to XML (and deserialization from XML)

stempler:
LGTM

PR #43, opened 2016-07-07
CRS flip handling on geometry

arunverma87:


util/plugins/eu.esdihumboldt.util.test/META-INF/MANIFEST.MF
```diff
@@ -7,3 +7,5 @@ Fragment-Host: eu.esdihumboldt.util;bundle-version="2.5.0"
 Bundle-RequiredExecutionEnvironment: JavaSE-1.7
 Import-Package: com.vividsolutions.jts.geom;version="1.13.0",
  org.junit;version="4.8.1"
+Require-Bundle: com.google.guava,
```
stempler: For the test bundle it doesn't matter much but usually I would prefer that you use package imports rather than require bundle, especially in cases where only a small part of a library is actually used.

util/plugins/eu.esdihumboldt.util.test/src/eu/esdihumboldt/util/geometry/WindingOrderTest.java
```diff
@@ -77,14 +98,56 @@ public static void setUp() {
...
+			}
+		}
+	}
+
+	/**
+	 * Test of Flip state of CRS
+	 */
+	@Test
+	public void testCRSFlipped() {
+		assertTrue(WindingOrder.isCRSFlip(crs));
```
stempler: You should also add at least a test with a CRS where the winding order does not have to be reversed.
Otherwise this could also be a function returning `true` all the time :wink:

thorsten-reitz:
If the CRS can't be determined, I would fall back to Cartesian. Currently the process would fail, I assume?

    throw new IllegalStateException("Invalid CRS code", e);

io/plugins/eu.esdihumboldt.hale.io.jdbc.msaccess/src/eu/esdihumboldt/hale/io/jdbc/msaccess/MsAccessJdbcIOSupplier.java
```diff
@@ -0,0 +1,63 @@
+/*
+ * Copyright (c) 2015 Data Harmonisation Panel
```
thorsten-reitz: We need to change the default (c) header. The (c) lies with wetransform for all new files.

thorsten-reitz:
Does it make sense to convolute the MS Access changes into the CRS flip pull request? I would prefer to keep them separate.

stempler:
LGTM

@thorsten-reitz Anything open from your side?

thorsten-reitz:
LGTM.

PR #51, opened 2016-07-11
Schema reader for MS Access

arunverma87:
…library


stempler:
As a general note, when you create new projects, please run the `apply-preferences.groovy` script to apply the common project settings and auto-formatting rules (see also the [Contribution guidelines](https://github.com/halestudio/hale/blob/master/CONTRIBUTING.md#project-preferences-and-code-formatting)).
It' usually best to do that directly after creating the project. Make sure to refresh the project afterwards (closing and opening it again works also I think).

io/plugins/eu.esdihumboldt.hale.io.jdbc.msaccess/src/eu/esdihumboldt/hale/io/jdbc/msaccess/MsAccessDataReader.java
```diff
@@ -0,0 +1,50 @@
...
+import java.io.File;
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+
+public class MsAccessDataReader {
+
+	public static String getFirstData() {
```
stempler: This is intended for testing, right?
So I would move it to the test bundle. Also, you will need some additional logic to make it work when your test database is located inside the JAR.

stempler:
Location of the updated p2 repository to use in the target platform definition is
https://gitlab.wetransform.to/hale/hale-build-support/raw/2c063edeef29187d74badaadec0313ddbb033949/updatesites/platform

See https://github.com/halestudio/hale-platform/pull/6

common/plugins/eu.esdihumboldt.hale.common.schema/src/eu/esdihumboldt/hale/common/schema/io/impl/AbstractSchemaReader.java
```diff
@@ -32,6 +32,8 @@
 
 	private SchemaSpaceID schemaSpace;
 
+	private Boolean isSchemaNameQuoted;
```
stempler: This configuration for adapting the behavior of the `JDBCSchemaReader` should be done in that class, and not in the `AbstractSchemaReader`. The `AbstractSchemaReader` has no association with JDBC or databases in general and is used as base for any schema reader class.

Also, please use a `boolean` with the default value representing the current behavior. Please think about the choice of name as well, because the quoting not only affects the schema name, but at least also table names. Maybe something more generic like `useQuotes` or similar.

io/plugins/eu.esdihumboldt.hale.io.jdbc.msaccess.test/.settings/org.eclipse.jdt.core.prefs
```diff
@@ -0,0 +1,7 @@
+eclipse.preferences.version=1
```
stempler: You did not run the `apply-preferences` script yet for the new projects, please do so.

io/plugins/eu.esdihumboldt.hale.io.jdbc.msaccess.test/src/eu/esdihumboldt/hale/io/jdbc/msaccess/reader/MsAccessSchemaReaderTest.java
```diff
@@ -0,0 +1,52 @@
...
+ * or (at your option) any later version.
+ * 
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this distribution. If not, see <http://www.gnu.org/licenses/>.
+ * 
+ * Contributors:
+ *     wetransform GmbH <http://www.wetransform.to>
+ */
+
+package eu.esdihumboldt.hale.io.jdbc.msaccess.reader;
```
stempler: Your test project should be added to the test product (and any dependencies so the product validates). Otherwise the tests are not executed within the CI.

io/plugins/eu.esdihumboldt.hale.io.jdbc.msaccess.test/src/eu/esdihumboldt/hale/io/jdbc/msaccess/test/MsAccessDataReaderTestSuit.java
```diff
@@ -0,0 +1,250 @@
...
+	 * 
+	 * @throws Exception
+	 *             if an error occurs
+	 */
+	public void schemaReaderTest() throws Exception {
+		MsAccessSchemaReader schemaReader = new MsAccessSchemaReader();
+		schemaReader.setSource(new FileIOSupplier(getSourceTempFilePath()));
+		schemaReader.setParameter(JDBCSchemaReader.PARAM_USER, Value.of(USER_NAME));
+		schemaReader.setParameter(JDBCSchemaReader.PARAM_PASSWORD, Value.of(PASSWORD));
+		schemaReader.setIsSchemaNameQuoted(false);
```
stempler: Setting this behavior should be done in your `MsAccessSchemaReader` class. After all you want to customize the behavior of the class, not make it adaptable by the user.

io/plugins/eu.esdihumboldt.hale.io.jdbc.msaccess/plugin.xml
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<?eclipse version="3.4"?>
+<plugin>
+   <extension
+         point="eu.esdihumboldt.hale.io.jdbc.config">
+      <connection
+            configurer="eu.esdihumboldt.hale.io.jdbc.msaccess.MsAccessConnectionConfigurer"
+            id="eu.esdihumboldt.hale.io.jdbc.msaccess"
+            type="java.sql.Connection">
```
stempler: Here the specific connection type used by UCanAccess should be specified. This class is then used by HALE to detect  that it is a UCanAccess connection.

stempler: Also, right now it would make any connection attempt to other databases fail, because for your connection configurer all connections would be tried to cast to the UCanAccess connection type.

io/plugins/eu.esdihumboldt.hale.io.jdbc.msaccess.test/.settings/org.eclipse.jdt.core.prefs
arunverma87: I applied it before, Now again I run the groovy file but in that msaccees.test project is not there.

arunverma87:
Sorry for late pull request. Actually Hale-platform update site need to be updated in hale project as some dependency added in it. But at that time eclipse download site was not working to download eclipse plugins. So could not able to build.

stempler:
retest this please

io/plugins/eu.esdihumboldt.hale.io.jdbc.msaccess/src/eu/esdihumboldt/hale/io/jdbc/msaccess/MsAccessSchemaReaderAdvisor.java
```diff
@@ -0,0 +1,72 @@
...
+		int index = path.lastIndexOf("/");
+		String name;
+		if (index >= 0 && index + 1 < path.length()) {
+			name = path.substring(index + 1).toLowerCase();
+		}
+		else {
+			name = path.toLowerCase();
+		}
+		// remove extension
+		if (name.endsWith(".mdb")) {
```
stempler: Probably makes sense to handle the other extension you registered for the content type as well, what do you think?

io/plugins/eu.esdihumboldt.hale.io.jdbc.msaccess/src/eu/esdihumboldt/hale/io/jdbc/msaccess/MsAccessSchemaReaderAdvisor.java
```diff
@@ -0,0 +1,72 @@
...
+
+	@Override
+	public void configureSchemaCrawler(SchemaCrawlerOptions options) {
+		options.setTableInclusionRule(new InclusionRule() {
+
+			private static final long serialVersionUID = -1559715487368953641L;
+
+			@Override
+			public boolean test(String t) {
+				final Pattern uca_metadata_TablePattern = Pattern.compile("\\w*.UCA_METADATA.\\w*");
```
stempler: Ah, so that excludes all tables in the `UCA_METADATA` schema?
I think we should also exclude the tables in the `SYS` schema, or do you see a reason why they should be mappable in hale?

arunverma87: Here table names are coming as SchemaName.TableName format. So by pattern compilation it bacame easy. and regarding SYS tables, which can be handled in Connection String also.  jdbc:ucanacceess:///<Database path>;showschema=true;sysschema=false; It will not include Sys tables schema..  I thought , that will be extra thing if we put in Inclusion rule.  What do you suggest?

io/plugins/eu.esdihumboldt.hale.io.jdbc.msaccess/src/eu/esdihumboldt/hale/io/jdbc/msaccess/MsAccessSchemaReaderAdvisor.java
arunverma87: Oh you are talking about .accdb extension. That I completely forgot. I was testing on .mdb only. Thanks.

io/plugins/eu.esdihumboldt.hale.io.jdbc.msaccess/src/eu/esdihumboldt/hale/io/jdbc/msaccess/MsAccessSchemaReaderAdvisor.java
stempler: I thought SchemaCrawler would need the system tables. If it doesn't, we can also hide it by changing the connection string.

PR #78, opened 2016-07-18
Finishing the MsAccess support in Hale

arunverma87:


io/plugins/eu.esdihumboldt.hale.io.jdbc/src/eu/esdihumboldt/hale/io/jdbc/constraints/DatabaseTable.java
```diff
@@ -36,14 +36,30 @@
...
+		this.isTable = false;
+	}
+
 	/**
 	 * Create a default constraint. The default table name is the given type
 	 * local name.
 	 * 
 	 * @param type the type definition the table is associated to
 	 */
 	public DatabaseTable(TypeDefinition type) {
```
stempler: This constructor should be removed. It does not make sense to create constraints that state that a type is not a table and at the same time provide a table name.

io/plugins/eu.esdihumboldt.hale.io.jdbc/src/eu/esdihumboldt/hale/io/jdbc/constraints/DatabaseTable.java
```diff
@@ -36,14 +36,30 @@
 	// Database support.
 	private final boolean useQuote;
 
+	private final boolean isTable;
```
stempler: The `DatabaseTableFactory`should be adapted to properly persist and restore the constraint information, with the new information added to the constraint class.

stempler:
Is the file `CDDA_v13.mdb` actually allowed to be included here?
How big is it?

arunverma87:
I was testing with real time data. So I kept it in bundle for testing purpose only. Its size is 37.8 MB.

stempler:
@arunverma87 I would vote against adding a file that big. The fact if it is allowed to add it here, where it is publicly accessible, is another matter.

stempler:
LGTM

PR #83, opened 2016-07-22
Filters handling on Command Line

arunverma87:
…-Types Filter. Right now common to all source.


app/plugins/eu.esdihumboldt.hale.app.transform.test/src/eu/esdihumboldt/hale/app/transform/test/ExecuteTest.groovy
```diff
@@ -38,26 +38,41 @@ class ExecuteTest extends GroovyTestCase {
 
 	// XXX Doesn't work -> private static final String METADATA_PATH = "platform:/plugin/$PLUGIN_NAME/projects/gmdMD_Metadata.xml"
 	// works, same as absolute path does
+
+	private static final String CDDA_DATA = "C://Users/Arun/Downloads/Databases/CDDA_v13.mdb"
```
stempler: This should not make it in here.

app/plugins/eu.esdihumboldt.hale.app.transform.test/src/eu/esdihumboldt/hale/app/transform/test/ExecuteTest.groovy
```diff
@@ -38,26 +38,41 @@ class ExecuteTest extends GroovyTestCase {
...
-	 * 
-	 * Test transformation of an example project.	 * 
+	 *
+	 * Test typed filter of an example project.	 *
 	 */
-	void ignore_testTransformXml() {
+	void testExcludeTypeTransformXml() {
 		File targetFile =  File.createTempFile('transform-hydro', '.gml')
-		targetFile.deleteOnExit()
+		//targetFile.deleteOnExit()
```
stempler: Please re-enable the cleanup for the tests.

app/plugins/eu.esdihumboldt.hale.app.transform/src/eu/esdihumboldt/hale/app/transform/ExecApplication.groovy
```diff
@@ -133,6 +133,9 @@ $baseCommand
          [-preset <name-of-export-preset>]
          [-providerId <ID-of-target-writer>]
          [<setting>...]
+	 [-filter <filter-expression>]
+	 [-filter-on <Type>:<filter-expression>]
+ 	 [-exclude-type <Type>:...]
```
stempler: What does `:...` mean here?

app/plugins/eu.esdihumboldt.hale.app.transform/src/eu/esdihumboldt/hale/app/transform/ExecApplication.groovy
```diff
@@ -150,6 +153,8 @@ $baseCommand
   control which files to load.
   If you do not specify -include, it defaults to "**", i.e. all files being
   included, even if they are in sub-directories.
+  If you specify more than one filter ( e.g. -filter or -filter-on) will be applied as 'OR' on each instance.
```
stempler: You inserted that in the middle of the section explaining the usage of directories as sources. Please create a new paragraph with your content. Maybe add a small intro that filters are applied to source data, before stating what happens with multiple filters.

app/plugins/eu.esdihumboldt.hale.app.transform/src/eu/esdihumboldt/hale/app/transform/ExecApplication.groovy
```diff
@@ -215,6 +221,30 @@ $baseCommand
...
+			case '-filter-on':
+			//<Type>:CQL:<EXPRESSION>
+				if(value.findIndexValues {it = ':'}.size() >=2 ){
+					int typeNameIndex = value.indexOf(':')
+					if(typeNameIndex>0){
+						String typeName = value.take(typeNameIndex)
+						String expression = value.substring(typeNameIndex+1)
+						executionContext.filters.addTypeFilter(typeName, expression)
+					}
+				}else {
```
stempler: I thought you support using CQL as default for a filter? Then something like `River:width > 10` would be OK, or not?

app/plugins/eu.esdihumboldt.hale.app.transform/src/eu/esdihumboldt/hale/app/transform/ExecApplication.groovy
```diff
@@ -215,6 +221,30 @@ $baseCommand
...
+					warn("Illegal parameter -filter-on should be applied with Type")
+				}
+
+				break
+			case '-exclude-type':
+			// <Type1>:<Type2>:<Type3>
```
stempler: OK, I see now what you hint at in the usage documentation. The colon is used with different semantics in the other places (not as list separator), so I would prefer not to use it here in a different manner. I think it is alright if `-exclude-type` just takes one type as argument.

common/plugins/eu.esdihumboldt.hale.common.headless/src/eu/esdihumboldt/hale/common/headless/transform/filter/InstanceFilterDefinition.java
```diff
@@ -0,0 +1,202 @@
...
+
+import com.google.common.collect.ArrayListMultimap;
+import com.google.common.collect.Multimap;
+
+import eu.esdihumboldt.hale.common.instance.extension.filter.FilterDefinitionManager;
+import eu.esdihumboldt.hale.common.instance.model.Filter;
+import eu.esdihumboldt.hale.common.instance.model.Instance;
+
+/**
+ * TODO Type description
```
stempler: Please add a short description to get rid of the TODO.

common/plugins/eu.esdihumboldt.hale.common.headless/src/eu/esdihumboldt/hale/common/headless/transform/filter/InstanceFilterDefinition.java
```diff
@@ -0,0 +1,202 @@
...
+	private boolean applyFilters(Instance instance) {
+
+		if (this.unconditionalFilters.size() == 0 && this.typeFilters.size() == 0
+				&& this.excludedTypes.size() == 0)
+			return true;
+
+		// applying all filters one by one (also handling lazy evaluation)
+		return (this.unconditionalFilters.size() == 0 ? false : applyUnconditionalFilter(instance))
+				|| (this.typeFilters.size() == 0 ? false : applyTypedFilter(instance))
+				|| (this.excludedTypes.size() == 0 ? false : applyExcludeFilters(instance));
```
stempler: I don't think there should be an "or" used for the excluded types. As that check should be the least expensive, it should be done first, and false returned on a failed check.

common/plugins/eu.esdihumboldt.hale.common.headless/src/eu/esdihumboldt/hale/common/headless/transform/filter/InstanceFilterDefinition.java
```diff
@@ -0,0 +1,202 @@
...
+			if (filter.match(instance))
+				return true;
+		// Instance does not match any of the unconditional filter
+		return false;
+	}
+
+	private boolean applyTypedFilter(Instance instance) {
+		// Loop through all the type filters
+		for (String type : this.typeFilters.keySet()) {
+			if (instance.getDefinition().getName().getLocalPart().equals(type + "Type")) {
```
stempler: Matching against a type is not that easy. I think it would be a good idea to (1) add a method dedicated to check a type expression against a given type and (2) discuss again shortly how the expression should be matched.

app/plugins/eu.esdihumboldt.hale.app.transform/src/eu/esdihumboldt/hale/app/transform/ExecApplication.groovy
```diff
@@ -153,11 +153,13 @@ $baseCommand
...
   If you do not specify -include, it defaults to "**", i.e. all files being
   included, even if they are in sub-directories.
-  If you specify more than one filter ( e.g. -filter or -filter-on) will be applied as 'OR' on each instance. 
-  If instance matches any of the filter will be included for transformation.
   Patterns use the glob pattern syntax as defined in Java and should be quoted
   to not be interpreted by the shell, see
   http://docs.oracle.com/javase/8/docs/api/java/nio/file/FileSystem.html#getPathMatcher-java.lang.String-
+
+  -filter, -filter-on and -exclude-type are filtration, applied on source data.
+  If you specify more than one filter ( e.g. -filter or -filter-on) will be applied as 'OR' on each instance.
```
stempler: `-exclude-type` is handled differently, so the description should reflect that.

app/plugins/eu.esdihumboldt.hale.app.transform/src/eu/esdihumboldt/hale/app/transform/ExecApplication.groovy
```diff
@@ -227,23 +229,18 @@ $baseCommand
...
-			//<Type>:CQL:<EXPRESSION>
-				if(value.findIndexValues {it = ':'}.size() >=2 ){
-					int typeNameIndex = value.indexOf(':')
-					if(typeNameIndex>0){
-						String typeName = value.take(typeNameIndex)
-						String expression = value.substring(typeNameIndex+1)
-						executionContext.filters.addTypeFilter(typeName, expression)
-					}
-				}else {
+				int typeNameIndex = value.indexOf(':')
```
stempler: Thinking further about it we have a problem here with the colon. It's not a rare case that there is a colon used in a namespace. In the example I think the fully fledged QName would be represented like this:

`{eu:esdihumboldt:hale:example}RiverType`

That means that this approach to separating the type name from the other arguments does not work.

common/plugins/eu.esdihumboldt.hale.common.headless/src/eu/esdihumboldt/hale/common/headless/transform/filter/InstanceFilterDefinition.java
```diff
@@ -183,20 +186,25 @@ private boolean applyTypedFilter(Instance instance) {
 				return false;
 			}
 		}
-		// it reaches here that means Instance does not match any type of type
-		// filters.
-		// Method should return true.
+		// it reaches here that means Instance does not match any type of
+		// type filters. Method should return true.
 		return true;
```
stempler: But returning true here will result in the OR to yield always true, meaning that when an instance's type does not match the type of a type filter it will always be included, no matter what the result of the unconditional filters would be.

common/plugins/eu.esdihumboldt.hale.common.headless/src/eu/esdihumboldt/hale/common/headless/transform/filter/InstanceFilterDefinition.java
```diff
@@ -183,20 +186,25 @@ private boolean applyTypedFilter(Instance instance) {
...
+		// instance does not match any excluded types.
+		return false;
+	}
 
-		// instance does not match any excludeed types.
-		return true;
+	// Checking type of instance with givenType.
+	private boolean checkType(TypeDefinition instanceType, String givenType) {
+		// Here giveType is supplied by user which should be trailed by "Type"
+		givenType += "Type";
```
stempler: There should be different ways to compare the type, based on the given type expression:

1. Check for equality with the fully qualified type name (including the namespace information)
2. If no namespace is provided in the given expression, check for equality with the type's display name
3. Check for equality with the fully qualified XML element names (including the namespace information)

app/plugins/eu.esdihumboldt.hale.app.transform/src/eu/esdihumboldt/hale/app/transform/ExecApplication.groovy
arunverma87: I thought user will provide only simple name like "River". He will not supply whole QName as per above. In Hale Studio also, we can see only name, not whole QName.

app/plugins/eu.esdihumboldt.hale.app.transform/src/eu/esdihumboldt/hale/app/transform/ExecApplication.groovy
```diff
@@ -134,7 +135,8 @@ $baseCommand
          [-providerId <ID-of-target-writer>]
          [<setting>...]
 	 [-filter <filter-expression>]
-	 [-filter-on <Type>:<filter-expression>]
+	 [-filter-on <Type> <filter-expression>]
+ 	 [-args-file <Path>]
```
stempler: Maybe put that at the very beginning (even before `-project`), as this may replace the need to provide any of the other arguments (if they are provided in the file).

common/plugins/eu.esdihumboldt.hale.common.headless/src/eu/esdihumboldt/hale/common/headless/transform/filter/InstanceFilterDefinition.java
```diff
@@ -159,10 +163,7 @@ private boolean applyFilters(Instance instance) {
 			return false;
```
stempler: I think there is still a case not handled: If there is only excluded types, no other filters, all instances will be rejected.

app/plugins/eu.esdihumboldt.hale.app.transform.test/src/eu/esdihumboldt/hale/app/transform/test/ExecuteTest.groovy
```diff
@@ -78,26 +84,29 @@ class ExecuteTest extends GroovyTestCase {
...
+			]) { //
+				File output, int code ->
+				// check exit code
+				assert code == 0
+			}
+			validateArgsFileHydroXml(targetFile)
+
+			tempArgsFile.delete();
+
+		}catch(Exception ex){
```
stempler: Just catching the exception is like ignoring the test. We should rather find out why it is not working an fix the test.

app/plugins/eu.esdihumboldt.hale.app.transform/src/eu/esdihumboldt/hale/app/transform/ExecApplication.groovy
```diff
@@ -124,6 +124,7 @@ class ExecApplication extends AbstractApplication<ExecContext> {
 		println """
 Usage:
 $baseCommand
+	 [-args-file <Path>]
```
stempler: Please correct the indendation here, and also below for the filters. There should only be spaces used, no tabs.

PR #91, opened 2016-07-28
Sort cells when writing an alignment to file

### LINKED ISSUE #63: Specific cell order in alignment file
Having a reproducible order for cells in an alignment when it is saved would allow for easier comparison in a diff.
It is suggested to do the ordering based on the cell target(s).

stempler:
closes #63 

Also improved handling of errors occurring while saving the alignment that would result in the alignment being overridden by an empty file.


stempler:
retest this please

Issue #87, opened 2016-07-27
Loading data fails in projects with JDBC data sources


stempler:
Occurs when loading the project. Observed with Oracle and PostgreSQL sources. It seems that the wrong JDBC driver is being used (SQLite). Probably due to an issue with loading the driver class in OSGi.

PR #124, opened 2016-08-05
Handling JDBC driver loading mismatch

arunverma87:
issue #87


io/plugins/eu.esdihumboldt.hale.io.jdbc/src/eu/esdihumboldt/hale/io/jdbc/JDBCConnection.java
```diff
@@ -47,6 +50,20 @@
...
+		try {
+			driver = DriverManager.getDriver(jdbcUri.toString());
+		} catch (Exception ex) {
+			// Expected driver is not been loaded, so need to load it manually
+			// using prefix attribute in extension
+			try {
+				driver = findDriver(jdbcUri);
+				DriverManager.registerDriver(driver);
+			} catch (ClassNotFoundException classNotFoundEx) {
+				// TODO:: handle it
```
stempler: Yes, at least the exception should be logged.

io/plugins/eu.esdihumboldt.hale.io.jdbc/src/eu/esdihumboldt/hale/io/jdbc/extension/DriverConfigurationExtension.java
```diff
@@ -66,4 +68,19 @@ protected DriverConfiguration create(String elementId, IConfigurationElement ele
...
+	 * @param jdbcUri uri of JDBC connection
+	 * @return {@link DriverConfiguration}
+	 */
+	public DriverConfiguration findDriver(URI jdbcUri) {
+		for (DriverConfiguration config : getElements()) {
+			if (config.matchURIPrefix(jdbcUri))
+				return config;
+		}
+
+		return null;
```
stempler: This method may return null. You should document that, and where you use it also handle that case.

Issue #116, opened 2016-08-04
Generic schema selection page for importing JDBC schemas


stempler:
Right now the default behavior is to import tables and views from all schemas in a database. This should be configurable for the user, and the availability of these configuration options should be customizable for different kinds of databases.

PR #126, opened 2016-08-09
Generic Schema selection while importing tables and views

arunverma87:
pull request related to #116


io/plugins/eu.esdihumboldt.hale.io.jdbc.ui/src/eu/esdihumboldt/hale/io/jdbc/ui/SchemasRetrievalPage.java
```diff
@@ -106,71 +118,146 @@ public boolean updateConfiguration(ImportProvider provider) {
...
 		try {
+			// load configuration
+			loadConfiguration();
+
+			// if page is enable for driver
+			if (!isEnable) {
+				disposeControl();
+				// message
+				Label label = new Label(page, SWT.WRAP);
+				label.setText("Schema retrieval is not enabled for selected driver.");
```
stempler: What the user is interested here is selecting schemas not retrieving schemas - I think this message would be rather confusing.
Maybe a message like "Restricting the loaded tables to specific database schemas is not enabled for this driver. All database tables will be loaded."

io/plugins/eu.esdihumboldt.hale.io.jdbc/schema/eu.esdihumboldt.hale.io.jdbc.config.exsd
```diff
@@ -176,6 +176,30 @@
                </appinfo>
             </annotation>
          </attribute>
+         <attribute name="isSchemaSelectionEnable" type="boolean" use="required">
```
stempler: I would prefer having a child element for `driver` that encompasses all the configuration related to schema selection. Then the attribute names can also be more concise.
In the extension it might look like this:
```
<driver ...>
  ...
  <schema-selection enabled="true" multiple="false" selector="xxx.Class" />
</driver>
```

io/plugins/eu.esdihumboldt.hale.io.jdbc.ui/plugin.xml
```diff
@@ -16,6 +16,13 @@
...
          </provider>
       </configPage>
+      <configPage
+            class="eu.esdihumboldt.hale.io.jdbc.ui.SchemasRetrievalPage"
+            order="1">
+         <provider
+               ref="eu.esdihumboldt.hale.io.jdbc.schema.reader">
+         </provider>
+      </configPage>
       <!--<configPage
```
stempler: Please remove the commented out version.

io/plugins/eu.esdihumboldt.hale.io.jdbc.postgresql/src/eu/esdihumboldt/hale/io/jdbc/postgresql/PostSchemaSelector.java
```diff
@@ -0,0 +1,74 @@
...
+		Statement statemnt = null;
+		String sql = "select schema_name from information_schema.schemata;";
+		ResultSet rs;
+		try {
+			statemnt = connection.createStatement();
+			rs = statemnt.executeQuery(sql);
+			while (rs.next()) {
+				schemas.add(rs.getString("schema_name"));
+			}
+			return schemas;
```
stempler: Wasn't there the intention to exclude `information_schemas` and `pg_*` schemas from the result? I think in this class would be the appropriate place to filter the returned list.

arunverma87: I have removed in SchemaSelector page using filtering.

io/plugins/eu.esdihumboldt.hale.io.jdbc.ui/src/eu/esdihumboldt/hale/io/jdbc/ui/SchemasRetrievalPage.java
```diff
@@ -106,71 +118,146 @@ public boolean updateConfiguration(ImportProvider provider) {
...
+			}
+		});
+
+		schemaTable.addFilter(new ViewerFilter() {
+
+			@Override
+			public boolean select(Viewer viewer, Object parentElement, Object element) {
+				String value = (String) element;
+
+				if (value.equalsIgnoreCase("information_schema") || value.startsWith("pg_"))
```
stempler: This is logic specific to a certain type of database. The UI should not have to care about that. The schema selector for a database should only return what is appropriate.

stempler:
Review done - please check the individual items and don't hesitate to ask me if there are any questions. Thanks.

io/plugins/eu.esdihumboldt.hale.io.jdbc/schema/eu.esdihumboldt.hale.io.jdbc.config.exsd
arunverma87: One can add only one choice in any Element and we already added one Choice for prefix and which is unbounded, So I thought I should not add in to that Choice otherwise Schema-selection also be unbounded.

io/plugins/eu.esdihumboldt.hale.io.jdbc.postgresql/plugin.xml
```diff
@@ -16,16 +16,18 @@
...
-            multipleSchemaSelection="true"
             name="PostgreSQL/PostGIS"
             testUri="jdbc:postgresql://localhost/db"
             uriBuilder="eu.esdihumboldt.hale.io.jdbc.postgresql.PostURIBuilder">
          <prefix
                value="jdbc:postgresql">
          </prefix>
+         <schema-selection
+               enabled="true"
+               multiple="false"
```
stempler: Multiple should be true for PostgreSQL (as before).

io/plugins/eu.esdihumboldt.hale.io.jdbc.ui/src/eu/esdihumboldt/hale/io/jdbc/ui/SchemasRetrievalPage.java
```diff
@@ -135,7 +133,7 @@ protected void onShowPage(boolean firstShow) {
 				disposeControl();
 				// message
 				Label label = new Label(page, SWT.WRAP);
-				label.setText("Schema retrieval is not enabled for selected driver.");
+				label.setText("Schema selection is not enabled for selected driver.");
```
stempler: I still think it would be better to give more information to the user. What does this actually mean to the user? He might get the idea that he should do something to enable it.

I revise my previous suggestion to this:
"Restricting the tables to load to tables from specific database schemas is not supported for this database driver. All database tables will be loaded."

stempler:
LGTM

Issue #130, opened 2016-08-11
CSV reader should allow to specify decimal divisor


thorsten-reitz:
Currently, the CSV reader defaults to the system setting for determining the decimal divisor in numeric fields. This fails if the source file used a different decimal divisor. It should be possible to specify the divisor (`.` or `,`) in the configuration wizard.

PR #158, opened 2016-09-08
Comma as a decimal seperator handling for CSV Schema reader

arunverma87:
Test on CLI, which works perfectly with decimal parameter.

related to #130


io/plugins/eu.esdihumboldt.hale.io.csv/plugin.xml
```diff
@@ -105,6 +105,17 @@
                   default="\">
             </valueDescriptor>
          </providerParameter>
+         <providerParameter
+               description="Divisor is used to identify float value"
```
stempler: *Divisor* sounds strange to me in that context when talking about numbers. Wouldn't `separator` be [more appropriate](http://dict.leo.org/ende/index_de.html#/search=Dezimaltrenner&searchLoc=0&resultOrder=basic&multiwordShowSingle=on&pos=0)?

stempler:
Can you please also add information on the newly added parameter to the changelog? Thanks

PR #180, opened 2016-10-06
Issue related to MsAccess connection in import schema 

arunverma87:
UcanAccess driver does not support encoded jdbc url. But to support space in database path we have to encode path prior to create connection and decode it prior to read schema or instance. If user import schema using "From file" option, it does not create any issues but if user load schema using "From Database (JDBC)" option, it throws an exception. So it would be better to remove msaccess driver from "From Database (JDBC)" option.


ui/plugins/eu.esdihumboldt.hale.ui.application/HALE.product
```diff
@@ -98,7 +98,6 @@ http://www.wetransform.to/services/support
       <feature id="eu.esdihumboldt.hale.ui.feature.dropins"/>
       <feature id="de.fhg.igd.hale.fme.feature"/>
       <feature id="eu.esdihumboldt.hale.platform.resources"/>
-      <feature id="eu.esdihumboldt.hale.ui.feature.updates"/>
```
stempler: Why this change?

io/plugins/eu.esdihumboldt.hale.io.jdbc/src/eu/esdihumboldt/hale/io/jdbc/extension/DriverConfigurationExtension.java
```diff
@@ -63,7 +63,9 @@ protected String getIdAttributeName() {
 	@Override
 	protected DriverConfiguration create(String elementId, IConfigurationElement element) {
 		if (element.getName().equals("driver")) {
-			return new DriverConfiguration(elementId, element);
+			DriverConfiguration driverConfiguration = new DriverConfiguration(elementId, element);
+			if (!driverConfiguration.isFileBased())
```
stempler: The DriverConfiguration fullfills more purposes than just the Driver listing in the UI. Wouldn't it make more sense to just hide file based drivers in the UI?

io/plugins/eu.esdihumboldt.hale.io.jdbc.msaccess/plugin.xml
```diff
@@ -11,6 +11,7 @@
       <driver
             class="net.ucanaccess.jdbc.UcanaccessDriver"
             id="eu.esdihumboldt.hale.io.jdbc.msaccess.driver"
+            isFileBased="true"
```
stempler: I would be nice to do this configuration also for SpatiaLite, as it is also file based.

Issue #174, opened 2016-09-23
See object counts per context


thorsten-reitz:
When I load source data and create condition contexts on the source schema, I want to see how many objects and properties there are for these contexts.

PR #199, opened 2016-10-12
Object counts per condition context

arunverma87:
related to issue #174


ui/plugins/eu.esdihumboldt.hale.ui/src/eu/esdihumboldt/hale/ui/service/population/internal/PopulationServiceImpl.java
```diff
@@ -232,22 +285,103 @@ public void resetPopulation(DataSet dataSet) {
...
 	 */
-	private void addToPopulation(Group group, EntityDefinition groupDef) {
-		for (QName propertyName : group.getPropertyNames()) {
-			EntityDefinition propertyDef = AlignmentUtil.getChild(groupDef, propertyName);
+	private void addToPopulation(Group group, EntityDefinition groupDef, List<ChildContext> path) {
+		Iterable<? extends EntityDefinition> children = entityDefinitionService
+				.getChildren(groupDef);
+		if (children != null && children.iterator().hasNext()) {
+			for (EntityDefinition def : children) {
+				if (groupDef instanceof TypeEntityDefinition)
```
stempler: Can this ever happen?

ui/plugins/eu.esdihumboldt.hale.ui/src/eu/esdihumboldt/hale/ui/service/population/internal/PopulationServiceImpl.java
```diff
@@ -232,22 +285,103 @@ public void resetPopulation(DataSet dataSet) {
...
+	private void evaluateContext(Group group, EntityDefinition groupDef, List<ChildContext> path) {
 
-			if (propertyDef != null) {
-				Object[] values = group.getProperty(propertyName);
+		if (path == null || path.isEmpty()) {
+			// group or instance at end of path
+			increase(groupDef, 1);
+		}
+		else {
+			ChildContext context = path.get(0);
```
stempler: Isn't here always the last path element relevant? Because that should be the path element that the child has in addition to its parent. The parent is on the instance side represented by the group.

ui/plugins/eu.esdihumboldt.hale.ui/src/eu/esdihumboldt/hale/ui/service/population/internal/PopulationServiceImpl.java
```diff
@@ -232,22 +285,103 @@ public void resetPopulation(DataSet dataSet) {
...
 
-				increase(propertyDef, values.length);
+				if (context.getChild().getName().equals(groupDef.getDefinition().getName())) {
+					increase(groupDef, values.length);
+				}
 
 				for (Object value : values) {
 					if (value instanceof Group) {
-						addToPopulation((Group) value, propertyDef);
+						addToPopulation((Group) value, groupDef, subPath);
```
stempler: Here you pass all Group values on, but shouldn't that be done only for those that were counted for this level?
For instance: If it did not match the condition at this level, it should not be counted on the next.

ui/plugins/eu.esdihumboldt.hale.ui/src/eu/esdihumboldt/hale/ui/service/population/internal/PopulationServiceImpl.java
stempler: Hm, I am starting to see that you changed the way that addToPopulation works by specifying what you want to limit the count to, via the provided EntityDefinition. It would be good if you would reflect that in the method documentation. "the group entity definition" does not tell anything - my first assumption was that it is directly related to the group.

ui/plugins/eu.esdihumboldt.hale.ui.service.helper/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,12 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: Hale UI Service Implementer
+Bundle-SymbolicName: eu.esdihumboldt.hale.ui.service.helper
+Bundle-Version: 3.1.0.qualifier
+Bundle-RequiredExecutionEnvironment: JavaSE-1.7
+Bundle-Vendor: wetransform GmbH
+Import-Package: eu.esdihumboldt.hale.common.align.model,
```
stempler: By convention these projects/bundles should rather be placed in `common` instead of `ui` and named accordingly.

PR #204, opened 2016-10-25
Added Mssql support in Hale

arunverma87:


io/plugins/eu.esdihumboldt.hale.io.jdbc.mssql/src/eu/esdihumboldt/hale/io/jdbc/mssql/MsSqlGeometries.java
```diff
@@ -0,0 +1,231 @@
...
+					String srsName = CRS.toSRS(geom.getCRSDefinition().getCRS());
+
+					// Todo:: Do we really need transformation for 'geometry'
+					// columnDataType instances? because this type represents
+					// data in a Euclidean (flat) coordinate system.
+
+					// getting Axis order of geometry CRS
+					CRS.AxisOrder axisOrder = CRS.getAxisOrder(geom.getCRSDefinition().getCRS());
+					// if axis order is not x/y ordering then will need to
+					// transform geometry to target crs with longiture first
```
stempler: That isn't exactly correct, because it may for instance be a projected reference system that does not use latitude/longitude deegrees at all.
Anyway, it doesn't hurt to do the transformation.
A better check might be to just compare the axis order with that of the target CRS - if it is the same, no transformation needs to be done.

io/plugins/eu.esdihumboldt.hale.io.jdbc.mssql/src/eu/esdihumboldt/hale/io/jdbc/mssql/MsSqlGeometries.java
```diff
@@ -0,0 +1,231 @@
...
+					} catch (NumberFormatException nfEx) {
+						// TODO::Using UI ask user to enter default SRId, if
+						// can not extract it.
+					}
+				}
+
+				Geometry targetGeometry;
+				if (targetCRS != null) {
+					MathTransform transform = CRS
+							.findMathTransform(geom.getCRSDefinition().getCRS(), targetCRS);
```
stempler: findMathTransform can be very costly if done many times. We should look into a global mechanism that we can use where there is an automatic caching done.
We would need that as well for the reprojection Groovy helper function that is to be implemented in context of another issue.

io/plugins/eu.esdihumboldt.hale.io.jdbc.mssql/src/eu/esdihumboldt/hale/io/jdbc/mssql/MsSqlGeometries.java
```diff
@@ -0,0 +1,231 @@
...
+			Geometry jtsGeom = null;
+			int srId = 0;
+
+			if (rs.next()) {
+				srId = rs.getInt(1);
+				String geomAsText = rs.getString(2);
+				// conversion to JTS via WKT
+				WKTReader wktReader = new WKTReader(new GeometryFactory());
+				try {
+					jtsGeom = wktReader.read(geomAsText);
```
stempler: Can you state if this covers all geometry types available in MS SQL. Are there any that we don't support?

io/plugins/eu.esdihumboldt.hale.io.jdbc.mssql/src/eu/esdihumboldt/hale/io/jdbc/mssql/MsSqlGeometries.java
```diff
@@ -0,0 +1,231 @@
...
+				} catch (Exception e) {
+					log.error("Could not load geometry from database", e);
+				}
+			}
+
+			CRSDefinition crsDef = null;
+
+			String authName = SRSUtil.getAuthorizedName(srId, connection);
+			if (authName != null && authName.equals("EPSG")) {
```
stempler: Would be good if we don't rely on case here.

ui/plugins/eu.esdihumboldt.hale.ui.application/HALE.product
```diff
@@ -100,6 +100,7 @@ http://www.wetransform.to/services/support
       <feature id="eu.esdihumboldt.hale.platform.resources"/>
       <feature id="eu.esdihumboldt.hale.ui.feature.updates"/>
       <feature id="eu.esdihumboldt.hale.io.feature.msaccess"/>
+      <feature id="eu.esdihumboldt.hale.io.feature.mssql" version="3.1.0.qualifier"/>
```
stempler: Better not to include a specific version number. Just causes problems.

stempler:
In addition I assume there needs to be an update to the target platform. Is there a corresponding pull request for hale-platform planned?

io/plugins/eu.esdihumboldt.hale.io.jdbc.mssql.test/src/eu/esdihumboldt/hale/io/jdbc/mssql/test/MsSqlInstanceReader.java
```diff
@@ -0,0 +1,71 @@
...
+import de.fhg.igd.slf4jplus.ALoggerFactory;
+import eu.esdihumboldt.hale.common.core.io.supplier.LocatableInputSupplier;
+import eu.esdihumboldt.hale.io.jdbc.JDBCInstanceReader;
+
+/**
+ * An extension of {@link JDBCInstanceReader} for MsSQL Server Database
+ * 
+ * @author Arun
+ */
+public class MsSqlInstanceReader extends JDBCInstanceReader {
```
arunverma87: This is in a test. I have to provide it to support special character in URI. I did it before I add general mechanism to support  special character in connection string.

io/plugins/eu.esdihumboldt.hale.io.jdbc.mssql.test/src/eu/esdihumboldt/hale/io/jdbc/mssql/test/MsSqlSchemaReader.java
```diff
@@ -0,0 +1,71 @@
...
+import de.fhg.igd.slf4jplus.ALoggerFactory;
+import eu.esdihumboldt.hale.common.core.io.supplier.LocatableInputSupplier;
+import eu.esdihumboldt.hale.io.jdbc.JDBCSchemaReader;
+
+/**
+ * An extension of {@link JDBCSchemaReader} for MsSQL Server Database
+ * 
+ * @author Arun
+ */
+public class MsSqlSchemaReader extends JDBCSchemaReader {
```
arunverma87: This is in a test. I have to provide it to support special character in URI. I did it before I add general mechanism to support  special character in connection string.

ui/plugins/eu.esdihumboldt.hale.ui.application/HALE.product
arunverma87: Ok

arunverma87:
@stempler , I updated my code on based on your code, still there are couple of comments(like FindMathTransform and  AxisOrder),  I want to discuss with you before doing any changes on them.

Regarding target platform, Microsoft does not provide JDBC driver on maven. So I first installed it in to local maven and then added it to target platform.  Please suggest me where should I put driver library in target platform project, So I can bundle them. Should I put in to modules directory?

arunverma87:
@stempler I merged new hale target platform pull request by you. Please merge this pull request.

stempler:
Can you also add the documentation on MS SQL specific behavior and limitations that we discussed?
I think it would be best to have MS SQL specific sections in the help for Database Import and Export for data in the help reference.

Also there is one open comment regarding the case of the authority name - do you know that it always will be uppercase?

io/plugins/eu.esdihumboldt.hale.io.jdbc.mssql/src/eu/esdihumboldt/hale/io/jdbc/mssql/MsSqlGeometries.java
arunverma87: I checked in MSSQL server, authority name is always in a uppercase

arunverma87:
I think, I should test it with real time data first. Then I can be more sure with concrete result for limitations.

stempler:
Well but you can already include what you know, namely the specific handling of certain geometry types when reading data, and how when writing data is dealt with the SRS.

arunverma87:
Please review latest commits @stempler, So I can complete this as soon as possible.

Issue #181, opened 2016-10-07
Interpolate Arcs during Import


thorsten-reitz:
Hale studio doesn't support Arcs other than passing them through. They can't be rendered in the map view, and when a geometry with arcs is processed (buffered, aggregated), the arc isn't taken into account correctly.  Therefore we need to interpolate any arcs during the reading process, to a linestring made of straight segments.

The interpolation method must take topology into account, i.e. if two arc nodes or segments touch each other, that relation needs to be maintained. There may be no new gaps or overlaps.

DoD:
- Add an option to interpolate arcs at least the following readers: JDBC, GML
- This option needs a single control: maximum positional error
- Process all arcs

Implementation suggestion: Use grid based iterative interpolation

PR #211, opened 2016-11-10
Arc interpolation to LineString

arunverma87:
Utility class for interpolation of arc
Added ui support (i.e. "Interpolation Setting Page") to specify "Max Position Error" 
Integrate interpolation setting page with GML reader

Pull request related to #181


util/plugins/eu.esdihumboldt.util.test/src/eu/esdihumboldt/util/geometry/InterpolationTest.java
```diff
@@ -0,0 +1,209 @@
...
+	 */
+	@Test
+	public void testInterpolation() {
+		System.out.println("-- Test-" + testIndex + " begin --");
+
+		Interpolation<LineString> interpolation = new ArcInterpolation(this.arcCoordinates, e);
+		Geometry interpolatedArc = interpolation.interpolateRawGeometry();
+
+		assertNotNull(interpolatedArc);
+		Assert.assertEquals(interpolatedArc.getClass(), generatedGeometryType);
```
stempler: It would be good if the test would test some of the characteristics of the interpolated arcs.
For instance that the positional error of the points is indeed equal to or less than the setting and maybe if the points lie on the grid.

io/plugins/eu.esdihumboldt.hale.io.gml.geometry/src/eu/esdihumboldt/hale/io/gml/geometry/handler/ArcHandler.java
```diff
@@ -0,0 +1,104 @@
...
+	 */
+	@Override
+	public Object createGeometry(Instance instance, int srsDimension, IOProvider reader)
+			throws GeometryNotSupportedException {
+		@SuppressWarnings("unchecked")
+		DefaultGeometryProperty<LineString> lineStringGeomProperty = (DefaultGeometryProperty<LineString>) super.createGeometry(
+				instance, srsDimension, reader);
+		// TODO:: add support to ask maximum positional error from user..
+		double maxPositionalError = reader
+				.getParameter(InterpolationConstant.INTERPOL_MAX_POSITION_ERROR).as(Double.class);
```
stempler: You declared the parameter as optional, but this will fail if the parameter is not provided. Where does your default value come into play?

util/plugins/eu.esdihumboldt.util/src/eu/esdihumboldt/util/geometry/interpolation/ArcInterpolation.java
```diff
@@ -0,0 +1,253 @@
...
+		super(coordinates, maxPositionalError);
+	}
+
+	/**
+	 * @see eu.esdihumboldt.util.geometry.interpolation.Interpolation#validateRawCoordinates()
+	 */
+	@Override
+	protected boolean validateRawCoordinates() {
+		if (rawGeometryCoordinates.length != 3) {
+			log.error("Invalid arc geometry. Arc must be represented by 3 points in GML.");
```
stempler: Just a note - you decided to use the 3 point representation for the interpolation, that's fine, but the class is not GML specific and the representation neither, so I would remove the GML reference here.

util/plugins/eu.esdihumboldt.util/src/eu/esdihumboldt/util/geometry/interpolation/ArcInterpolation.java
```diff
@@ -0,0 +1,253 @@
...
+
+		double centerX = ((aSlope * bSlope * (BC_Mid.y - AB_Mid.y)) + (aSlope * BC_Mid.x)
+				- (bSlope * AB_Mid.x)) / (aSlope - bSlope);
+
+		double centerY = AB_Mid.y - ((centerX - AB_Mid.x) / aSlope);
+
+		return new Coordinate(centerX, centerY);
+	}
+
+	private LineString interpolateToLineString(Coordinate[] arcCoordinates, Coordinate center,
```
stempler: Would this also work directly with an ArcByCenterPoint representation? Then it might sense to expose that method or a similar method to easier support that kind of representation.

I see that this method at least relies on three points, but maybe it can be made to work also with two points (or a third point can be calculated).

util/plugins/eu.esdihumboldt.util/src/eu/esdihumboldt/util/geometry/interpolation/Interpolation.java
```diff
@@ -0,0 +1,166 @@
...
+	 */
+	protected abstract T interpolatedGeometry();
+
+	/**
+	 * relocate geometry coordinate to nearest grid point.
+	 * 
+	 * @param coordinate geometry coordinate
+	 * @return relocated grid coordinate
+	 */
+	protected Coordinate pointToGrid(Coordinate coordinate) {
```
stempler: Because it is hard to create test cases to test the functionality of interpolation exactly, I think it would be good to test parts of the algorithm that can be easily unit tested. This method is a candidate for that, it is quite easy to come up with tests here and cover all the corner cases.

util/plugins/eu.esdihumboldt.util/src/eu/esdihumboldt/util/geometry/interpolation/Interpolation.java
```diff
@@ -0,0 +1,166 @@
...
+public abstract class Interpolation<T extends Geometry> {
+
+	/**
+	 * Grid step size factor
+	 */
+	protected final int GRID_FACTOR = 2;
+	/**
+	 * Next coordinate distance factor.
+	 */
+	protected final double NEXT_COORDINATE_DISTANCE_FACTOR = 4;
```
stempler: Can you say something about the background of this magic number?

util/plugins/eu.esdihumboldt.util/src/eu/esdihumboldt/util/geometry/interpolation/ArcInterpolation.java
```diff
@@ -0,0 +1,253 @@
...
+
+		double distance = Math.sqrt(Math.pow((current.x - nextArcCoordinate.x), 2)
+				+ Math.pow((current.y - nextArcCoordinate.y), 2));
+		if (distance <= e)
+			return true;
+
+		return false;
+	}
+
+	private List<Coordinate> getNeighbouringCoordinates(Coordinate current,
```
stempler: Why not just take neighbours on the grid? Or am I missing something here?

stempler:
@thorsten-reitz Would be helpful if you can have a look at the algorithm as well.
Also, there is still the open topic with non-Arc points that might have to be moved, which needs to be discussed at some point.

thorsten-reitz:
I did a quick review of the algorithm yesterday. In principle, it looked OK to me. I think it makes sense to first interpolate the arc in angular increments, and then to move the created points to the grid. There were a few questions though that I didn't immediately find answers to by glancing at the code:

1. When moving points to the grid, could 2 or more points of a single interpolated arc end up in the same grid point? If yes, do you check that the excess points are removed? Otherwise the segment would be degenerate.
2. How do you match the grid spacing width to the interpolation segment length? Basically, what do you do to ensure that there is no bad aliasing when moving the points to the grid?

I would also make sense to use one internal representation of an arc (e.g. the center + radius one you use now) and to convert the other GML representations into that in a preprocessing step.

util/plugins/eu.esdihumboldt.util/src/eu/esdihumboldt/util/geometry/interpolation/ArcInterpolation.java
arunverma87: This method gets all the neighbour coordinates of Arc not grid. First, algorithm search for the neighbour coordinates of arc then it moves most suited neighbour to grid point (nearest).

arunverma87:
Algorithm start with the search of next neighbour coordinate with (4*e) distance if none of the neighbour fits in the circular arc , algorithm half the distance factor (4/2 *e)  and again search for the best suited neighbour. Here grid size is also started with (2*e), So, as per my point of view I don't think that can be a case. Although, It is good to check that 2 points are not ending up on same grid point.

util/plugins/eu.esdihumboldt.util/src/eu/esdihumboldt/util/geometry/interpolation/Interpolation.java
arunverma87: this is the moving distance factor from one arc coordinate to find neighbour arc coordinate. Algorithm start with the distance (``NEXT_COORDINATE_DISTANCE_FACTOR * e ``) to search for the  next most deserved neighbour.

util/plugins/eu.esdihumboldt.util/src/eu/esdihumboldt/util/geometry/interpolation/ArcInterpolation.java
arunverma87: I did not consider ArcByCenterPoint representation. I started with simple Arc representation.

stempler: Ok, I added an improvement ticket regarding that, see #212

util/plugins/eu.esdihumboldt.util/src/eu/esdihumboldt/util/geometry/interpolation/ArcInterpolation.java
stempler: I don't quite get it. To me it just looks like this method is generating points that are a specific distance away on different axes in different directions and combinations. Where is here any relation to the arc?

arunverma87: From these coordinates, algorithm finds best suited coordinate for Arc and then it will be move to grid point.

stempler:
@thorsten-reitz OK from my point of view, some more testing would be good I think.
Also, there is still the open topic with non-Arc points that might have to be moved, which needs to be discussed at some point. It might also make sense to allow configuring that start- and end-points are not moved.

Issue #175, opened 2016-09-28
Show only mandatory properties in Schema Explorer


thorsten-reitz:
Add an Option to source and target schema explorer to hide optional properties

thorsten-reitz:
@stempler  Minor quibble, when I toggle to show only mandatory properties, "location" is still included for any GML feature type, though it's 0...1. Does that make sense

stempler:
@thorsten-reitz  @arunverma87 I think groups are not included in the check, so that should be added.

stempler:
I think the icon could be improved. Right now it does not really hint at that it is related to optional/mandatory properties.

As the icon is active when optional properties are shown, it could symbolize optional properties.

thorsten-reitz:
I agree. What about picking up the red star that we use to indicate mandatory properties?

stempler:
Yes, that would work. But it would be different to the other filters, where the icons symbolize what is currently shown that is hidden when you deactivate it.

thorsten-reitz:
Works now.

PR #215, opened 2016-11-16
Added optional properties filter

arunverma87:
Pull request related to #175


ui/plugins/eu.esdihumboldt.hale.ui.common/src/eu/esdihumboldt/hale/ui/common/definition/viewer/OptionalPropertiesFilter.java
```diff
@@ -0,0 +1,60 @@
...
+		}
+		if (element instanceof EntityDefinition) {
+			EntityDefinition entityDef = (EntityDefinition) element;
+
+			Definition<?> def = entityDef.getDefinition();
+
+			if (def instanceof PropertyDefinition) {
+				Cardinality cardinality = ((PropertyDefinition) def)
+						.getConstraint(Cardinality.class);
+				return !(cardinality.getMinOccurs() == 0 && ((PropertyDefinition) def)
```
stempler: When minOccurs is zero for an element it is always optional, independent of the NillableFlag.
If an element must occur more than once it may still be optional, if it is nillable and has only optional children.

ui/plugins/eu.esdihumboldt.hale.ui/src/eu/esdihumboldt/hale/ui/service/entity/OptionalPropertiesFilter.java
```diff
@@ -50,11 +62,39 @@ public boolean select(Viewer viewer, Object parentElement, Object element) {
...
+					return !areChildrenOptional(entityDef);
+				}
 			}
 		}
 		return true;
 	}
 
+	private boolean areChildrenOptional(EntityDefinition entityDef) {
+
+		Collection<? extends EntityDefinition> children = eds.getChildren(entityDef);
```
stempler: Using the `EntityDefinitionService` here is a bit overkill, as it will also retrieve children with contexts, which don't have to be checked (as they are just duplicates of other children with the same info). But it shouldn't be to much additional effort as usually there are not many contexts - so you can leave it like it is if you like, but it would be good if you add a comment there stating that definitions are checked that don't need to and that this might be improved.

ui/plugins/eu.esdihumboldt.hale.ui.common/src/eu/esdihumboldt/hale/ui/common/definition/viewer/OptionalPropertiesFilter.java
```diff
@@ -97,4 +97,61 @@ private boolean areChildrenOptional(EntityDefinition entityDef) {
 		return true;
 	}
 
+	private Collection<? extends EntityDefinition> getChildren(EntityDefinition entityDef) {
```
stempler: Sorry to comment again - I think it would be really helpful to have this method available in `AlignmentUtil`, as I think this is not the only place that this is done / needed.

Issue #166, opened 2016-09-13
Support for formatted number output


thorsten-reitz:
Currently, hale uses java standard number serialization, but in many situations a defined number format, with a fixed number of decimals before and after the point, potentially with leading zeroes is required. It should be possible to define this for formats that don't have a constraint expressing that (many databases do, but XSD doesn't). The preferred implementation for this is as a constraint on the property, with a writer configuration option.

DoD:
- Provide control for this in XML only.

thorsten-reitz:
Check how to enforce serialization to formatted numbers first -> Spike

stempler:
We analyzed this and found that writing the coordinates of geometry objects is handled in a central place. So adding support for custom number formatting there would be quite easy. My understanding of the original use case was that this is mainly related to the representation of geometries.

@thorsten-reitz Is that correct and would an implementation that targets the encoded geometries be sufficient?

thorsten-reitz:
@stempler yes, that is correct and would be OK for now. I don't think a global setting makes sense for individual fields otherwise. Is there an XML type like BigDecimal that allows control over the serialisation of numbers?

PR #216, opened 2016-11-17
added support of formatted number output for geometry coordinates

arunverma87:
Related to #166


io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriter.java
```diff
@@ -134,6 +134,12 @@
 	public static final String PARAM_OMIT_NIL_REASON = "xml.notNil.omitNilReason";
 
 	/**
+	 * The parameter name for the flag specifying if the output of geometry
+	 * coordinates should be formatted.
+	 */
+	public static final String PARAM_GEOMETRY_FORMAT = "geometry.write.format";
```
stempler: As `format` could be anything I would vote for chosing a name that makes it easier to identify what the settings is about. Maybe `decimalFormat` instead of `format`?

util/plugins/eu.esdihumboldt.util/src/eu/esdihumboldt/util/geometry/NumberFormatter.java
```diff
@@ -0,0 +1,58 @@
...
+	private static final ALogger log = ALoggerFactory.getLogger(NumberFormatter.class);
+
+	/**
+	 * Format number to specified format
+	 * 
+	 * @param value double value
+	 * @param format a pattern
+	 * @return String presentation of a formatted number
+	 */
+	public static String formatTo(double value, String format) {
```
stempler: As this is called for every single double value I think it would make sense to instead pass the `DecimalFormat` to the writers instead of the string, avoiding to create and configure the format many many times.

io/plugins/eu.esdihumboldt.hale.io.gml.ui/src/eu/esdihumboldt/hale/io/gml/ui/GmlWriterSettingsPage.java
```diff
@@ -99,9 +112,44 @@ protected void createContent(Composite page) {
...
+		Group writeFormat = new Group(page, SWT.NONE);
+		writeFormat.setLayout(GridLayoutFactory.swtDefaults().numColumns(2).create());
+		writeFormat.setText("Formatted number output for geometry");
+		groupData.applyTo(writeFormat);
+
+		Label lbl = new Label(writeFormat, SWT.NONE);
+		lbl.setLayoutData(GridDataFactory.swtDefaults().align(SWT.END, SWT.CENTER).create());
+		lbl.setText("Format: ");
+
+		formattedText = new Text(writeFormat, SWT.SINGLE | SWT.BORDER);
```
stempler: Instead of having the empty text field implicitly setting no format I would prefer if there was a checkbox that activates having a format. The format field can then have a sensible default from the start.

io/plugins/eu.esdihumboldt.hale.io.gml.ui/src/eu/esdihumboldt/hale/io/gml/ui/GmlWriterSettingsPage.java
```diff
@@ -99,9 +112,44 @@ protected void createContent(Composite page) {
...
+		formattedText.setLayoutData(GridDataFactory.swtDefaults().align(SWT.FILL, SWT.CENTER)
+				.grab(true, false).create());
+		// filler
+		Label lblfiller = new Label(writeFormat, SWT.NONE);
+		lblfiller.setLayoutData(GridDataFactory.swtDefaults().align(SWT.END, SWT.CENTER).create());
+
+		Label formatDesc = new Label(writeFormat, SWT.NONE);
+		formatDesc.setLayoutData(GridDataFactory.swtDefaults().align(SWT.FILL, SWT.CENTER)
+				.grab(true, false).create());
+		formatDesc.setText("(for example 00000.000)");
```
stempler: I think having leading zeros is a case that would not be used often (if at all), so I would prefer an example / default that does not include them.

PR #220, opened 2016-11-18
Optional Properties filter

arunverma87:
group is added in the filter check. Related to #175

### DIFF
--- a/ui/plugins/eu.esdihumboldt.hale.ui.common/src/eu/esdihumboldt/hale/ui/common/definition/viewer/OptionalPropertiesFilter.java
+++ b/ui/plugins/eu.esdihumboldt.hale.ui.common/src/eu/esdihumboldt/hale/ui/common/definition/viewer/OptionalPropertiesFilter.java
@@ -56,6 +56,9 @@ public class OptionalPropertiesFilter extends ViewerFilter {
 						.getConstraint(Cardinality.class);
 				if (cardinality.getMinOccurs() == 0)
 					return false;
+				else
+					return !areChildrenOptional(entityDef);
+
 			}
 			else if (def instanceof PropertyDefinition) {
 				Cardinality cardinality = ((PropertyDefinition) def)
@@ -63,9 +66,8 @@ public class OptionalPropertiesFilter extends ViewerFilter {
 				if (cardinality.getMinOccurs() == 0)
 					return false;
 
-				if (((PropertyDefinition) def).getConstraint(NillableFlag.class).isEnabled()) {
+				if (((PropertyDefinition) def).getConstraint(NillableFlag.class).isEnabled())
 					return !areChildrenOptional(entityDef);
-				}
 			}
 		}
 		return true;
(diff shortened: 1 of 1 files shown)


ui/plugins/eu.esdihumboldt.hale.ui.common/src/eu/esdihumboldt/hale/ui/common/definition/viewer/OptionalPropertiesFilter.java
```diff
@@ -56,16 +56,18 @@ public boolean select(Viewer viewer, Object parentElement, Object element) {
...
+
 			}
 			else if (def instanceof PropertyDefinition) {
 				Cardinality cardinality = ((PropertyDefinition) def)
 						.getConstraint(Cardinality.class);
 				if (cardinality.getMinOccurs() == 0)
 					return false;
 
-				if (((PropertyDefinition) def).getConstraint(NillableFlag.class).isEnabled()) {
+				if (((PropertyDefinition) def).getConstraint(NillableFlag.class).isEnabled())
```
stempler: No else case? If it is not nillable, then it is for sure not optional.

ui/plugins/eu.esdihumboldt.hale.ui.common/src/eu/esdihumboldt/hale/ui/common/definition/viewer/OptionalPropertiesFilter.java
```diff
@@ -89,16 +91,20 @@ private boolean areChildrenOptional(EntityDefinition entityDef) {
...
 				if (cardinality.getMinOccurs() != 0)
 					return false;
+				else if (!areChildrenOptional(entityDef))
+					return false;
 			}
 			else if (def instanceof PropertyDefinition) {
 				Cardinality cardinality = ((PropertyDefinition) def)
 						.getConstraint(Cardinality.class);
 				if (cardinality.getMinOccurs() != 0)
 					return false;
```
stempler: That is wrong. If minOccurs is not zero, you still have to check if it is Nillable, if it is zero you don't have to check.

ui/plugins/eu.esdihumboldt.hale.ui.common/src/eu/esdihumboldt/hale/ui/common/definition/viewer/OptionalPropertiesFilter.java
```diff
@@ -89,16 +91,20 @@ private boolean areChildrenOptional(EntityDefinition entityDef) {
...
-			}
 
-			if (!areChildrenOptional(child))
-				return false;
+				if (((PropertyDefinition) def).getConstraint(NillableFlag.class).isEnabled()) {
```
stempler: No else here as well? If it is not Nillable (and minOccurs is not zero, which is the only condition that Nillable needs to be checked anyway), then it is not optional.

ui/plugins/eu.esdihumboldt.hale.ui.common/src/eu/esdihumboldt/hale/ui/common/definition/viewer/OptionalPropertiesFilter.java
```diff
@@ -77,15 +86,22 @@ private boolean areChildrenOptional(EntityDefinition entityDef) {
...
 					return false;
 			}
+			else if (def instanceof PropertyDefinition) {
+				Cardinality cardinality = ((PropertyDefinition) def)
+						.getConstraint(Cardinality.class);
 
-			if (!areChildrenOptional(child))
-				return false;
+				if (cardinality.getMinOccurs() != 0 && ((PropertyDefinition) def)
+						.getConstraint(NillableFlag.class).isEnabled()) {
```
stempler: But if minOccurs != 0 and the NillableFlag is NOT enabled you should return false here, without checking the children.

PR #221, opened 2016-11-22
MsSQL fix related wkt

arunverma87:


io/plugins/eu.esdihumboldt.hale.io.jdbc.mssql/src/eu/esdihumboldt/hale/io/jdbc/mssql/MsSqlGeometries.java
```diff
@@ -221,19 +221,20 @@ public void run() {
...
+					String authorizedId = null;
+					if (index > 0) {
+						authorityName = srsName.substring(0, index);
+						authorizedId = srsName.substring(index + 1).trim();
+					}
+					// we don't need wkt.
+					SRSUtil.addSRSinCache(srId, authorityName, authorizedId, null);
 				}
-				String wkt = crsDef.getCRS().toWKT();
-				SRSUtil.addSRSinCache(srId, authorityName, authorizedId, wkt);
```
stempler: So it was not used at all before? Otherwise you could just omit it if toWKT() causes an error.

arunverma87: Yes. But some how the same error was not logged in to my error log (Otherwise i could have solved it) So I didn't bother about debugging it then. but today, when I debug it then I came to know this error.

stempler: Ok. I'm not sure, but I interpret that as that the WKT was not used before.

Issue #228, opened 2016-11-23
Arc interpolation takes very long or is not terminating


stempler:
I have issues with the arc interpolation taking seemingly forever. I was not able to complete loading a project because loading the data did not finish in a sensible amount of time.

refs #181

Internal example project `aaa-au-flurstuecke`. Please create a test case from the respective geometry causing this problem.

stempler:
The problem geometry seems to be `LINESTRING (353248.457 5531386.407, 353249.438 5531386.407, 353250.399 5531386.217)`.

PR #230, opened 2016-11-24
bug fixing of interpolation of arc

arunverma87:
related to #228


io/plugins/eu.esdihumboldt.hale.io.gml.geometry/src/eu/esdihumboldt/hale/io/gml/geometry/handler/ArcHandler.java
```diff
@@ -78,20 +81,23 @@ public Object createGeometry(Instance instance, int srsDimension, IOProvider rea
 				.getParameter(InterpolationConstant.INTERPOL_MAX_POSITION_ERROR).as(Double.class);
 
 		if (maxPositionalError == null || maxPositionalError.doubleValue() <= 0) {
-			log.warn(
-					"Value of Max positional error parameter, for interpolation operation, is not valid. Default value has been taken.");
+			if (reportedWarning.compareAndSet(false, true)) {
```
stempler: This is not ideal, as it will only give a warning once in a running instance of hale, and not once per read operation. But it is definitely an improvement over the log spamming.

Issue #219, opened 2016-11-18
Arc interpolation: consistent points between different kinds of geometries


stempler:
Improvement for #181 

The current Arc interpolation implementation may lead to topological inconsistencies in respect to other kinds of geometries.
To deal with this there should be two modes:
1. Keep all original points, e.g. Arc start- and end-points (**default**)
2. Move all points (of all geometries) to the interpolation grid

stempler:
The current implementation of the interpolation will result in problems when assembling LinearRings, with errors like this: "Points of LinearRing do not form a closed linestring".

thorsten-reitz:
Seemed to work for the data sets I have available.

PR #232, opened 2016-11-28
Handling consisting points between different kind of geometries in interpolation

arunverma87:
related to #219 

Mode 2 : Moving all geometries to universal grid.

### DIFF
--- a/io/plugins/eu.esdihumboldt.hale.io.gml.geometry/src/eu/esdihumboldt/hale/io/gml/geometry/handler/GenericGeometryHandler.java
+++ b/io/plugins/eu.esdihumboldt.hale.io.gml.geometry/src/eu/esdihumboldt/hale/io/gml/geometry/handler/GenericGeometryHandler.java
@@ -325,7 +325,7 @@ public class GenericGeometryHandler extends InterpolationSupportedGeometryHandle
 		getInterpolationRequiredParameter(reader);
 		if (!isKeepOriginal()) {
 			Coordinate[] newCoordinates = moveToUniversalGrid(lineString.getCoordinates());
-			LineString newLine = getGeometryFactory().createLinearRing(newCoordinates);
+			LineString newLine = getGeometryFactory().createLineString(newCoordinates);
 			return newLine;
 		}
 		return lineString;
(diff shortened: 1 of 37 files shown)


io/plugins/eu.esdihumboldt.hale.io.gml.geometry/src/eu/esdihumboldt/hale/io/gml/geometry/handler/GenericGeometryHandler.java
```diff
@@ -236,23 +240,23 @@ else if (!commonCrs.equals(crs)) {
 				// create a MultiPolygon
 				Polygon[] polygons = new Polygon[geomList.size()];
 				for (int i = 0; i < geomList.size(); i++) {
-					polygons[i] = (Polygon) geomList.get(i);
+					polygons[i] = movePolygonToUniversalGrid((Polygon) geomList.get(i), reader);
```
stempler: The geometries that are combined here (Polygons, Points, etc.) have been created somewhere else (other handlers) where they should have been already moved. We're not actually creating geometries from coordinates that we are reading here. So I think any adaptions to the geometries themselves can be omitted.

util/plugins/eu.esdihumboldt.util/src/eu/esdihumboldt/util/geometry/interpolation/Interpolation.java
```diff
@@ -33,30 +33,31 @@
...
 	 */
-	protected final double NEXT_COORDINATE_DISTANCE_FACTOR = 4;
+	protected static final double NEXT_COORDINATE_DISTANCE_FACTOR = 4;
+
 	/**
-	 * Distance of next coordinate from current coordinate
+	 * Rounding scale for grid cell double value
 	 */
-	protected final double NEXT_COORDINATE_DISTANCE;
+	private static final int ROUNDING_SCALE = 6;
```
stempler: Would be much preferable to use the respective constant here, or even the enum value instead of this magic number.

arunverma87:
@stempler would you please review it again?

PR #233, opened 2016-11-30
simple feature helpers and custom function improvements

stempler:


stempler:
> Do I understand it right that the changes also include general handling of spilts of features in case merging of multi* doesn't work?

Not in that sense. But the helper function can be used in a Groovy (type) script to make that decision.

PR #239, opened 2016-12-05
Arc interpolation fixing and added support for circle type

arunverma87:
PR related to #228

### DIFF
--- a/io/plugins/eu.esdihumboldt.hale.io.gml.geometry/src/eu/esdihumboldt/hale/io/gml/geometry/handler/PolygonHandler.java
+++ b/io/plugins/eu.esdihumboldt.hale.io.gml.geometry/src/eu/esdihumboldt/hale/io/gml/geometry/handler/PolygonHandler.java
@@ -151,7 +151,9 @@ public class PolygonHandler extends InterpolationSupportedGeometryHandler {
 			ringValues = PropertyResolver.getValues(instance, "exterior.Ring", false);
 			values = combineCollections(values, ringValues);
 			List<LinearRing> outerRing = new ArrayList<>(1);
+
 			if (values != null && !values.isEmpty()) {
+				LinearRing outer = null;
 				Iterator<Object> iterator = values.iterator();
 				while (iterator.hasNext()) {
 					Object value = iterator.next();
@@ -160,10 +162,11 @@ public class PolygonHandler extends InterpolationSupportedGeometryHandler {
 						// GeometryProperty<LinearRing> instance
 						GeometryProperty<LinearRing> ring = (GeometryProperty<LinearRing>) ((Instance) value)
 								.getValue();
-						outerRing.add(ring.getGeometry());
+						outer = ring.getGeometry();
 						crs = checkCommonCrs(crs, ring.getCRSDefinition());
 					}
 				}
+				outerRing.add(outer);
 				outerRing = moveLinerRingsToUniversalGrid(outerRing, reader);
 				polygon = getGeometryFactory().createPolygon(outerRing.get(0), holes);
 			}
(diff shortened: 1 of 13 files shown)


io/plugins/eu.esdihumboldt.hale.io.gml.geometry/src/eu/esdihumboldt/hale/io/gml/geometry/handler/PolygonHandler.java
```diff
@@ -160,10 +162,11 @@ public Object createGeometry(Instance instance, int srsDimension, IOProvider rea
 						// GeometryProperty<LinearRing> instance
 						GeometryProperty<LinearRing> ring = (GeometryProperty<LinearRing>) ((Instance) value)
 								.getValue();
-						outerRing.add(ring.getGeometry());
+						outer = ring.getGeometry();
```
stempler: What is the reason for this change?

util/plugins/eu.esdihumboldt.util.test/src/eu/esdihumboldt/util/geometry/interpolation/CircleInterpolationTest.java
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+//0.01,3.2 3.33,3.33 0.01,-3.2
+	/**
+	 * test algorithm
+	 */
+	@Test
+	public void testInterpolation() {
+		System.out.println("-- Test-" + testIndex + " begin --");
+		if (skipTest) {
```
stempler: What is the purpose of this condition?

The danger I see here that looking at the test results you will never know that a test was actually skipped, because it will be listed as passed test.

Why should tests be skipped?
If there is a use case for that it might be wiser to use JUnit `assume*` methods (see [here](http://stackoverflow.com/questions/1689242/conditionally-ignoring-tests-in-junit-4) for an example).

io/plugins/eu.esdihumboldt.hale.io.gml.geometry/src/eu/esdihumboldt/hale/io/gml/geometry/handler/PolygonHandler.java
arunverma87: Before I was adding each value to list but I think we need a last value, while iterating each value.

stempler: Did you encounter a specific problem with this? Because generally I think there should only be one outer ring, and in this case the behavior of both approaches should actually be the same. So did you encounter a case where there were multiple outer rings found here?

util/plugins/eu.esdihumboldt.util.test/src/eu/esdihumboldt/util/geometry/interpolation/CircleInterpolationTest.java
arunverma87: I created parameterized tests with around 16 test cases, and this one is only for my convenience, if I want to run only single test while debugging.

io/plugins/eu.esdihumboldt.hale.io.gml.geometry/src/eu/esdihumboldt/hale/io/gml/geometry/handler/PolygonHandler.java
arunverma87: No I couldn't encounter any issue with that, but i changed it again, just to ensure no problem should be generated because of interpolation any more.

stempler: If you changed it again it did not make it into the PR

Issue #195, opened 2016-10-07
Validate code list values


thorsten-reitz:
Check whether the coded value assigned to a target element, e.g. `href` is valid according to either:
1. A codelist assigned in hale to that element
2. A codelist defined via a `codespace` attribute [TODO clarify use case]

thorsten-reitz:
Works as expected:

![grafik](https://cloud.githubusercontent.com/assets/3470356/21055127/83bca39a-be30-11e6-845b-861d74bc1433.png)

PR #241, opened 2016-12-07
code list validator

stempler:


stempler:
Solves #195 in respect to using information from code lists assigned in a hale project to validate values of the respective properties.

PR #242, opened 2016-12-09
Enhancement in jdbc schema selection

arunverma87:


thorsten-reitz:
What exactly does this do? Provide an option to select all schemas of a database?

arunverma87:
Yes. for example, in SQL server, all schemas are selected by default and if we want single schema then we have to deselect others one by one. I think it is annoying for user.

Issue #165, opened 2016-09-13
Add SKOS Codelist reader


thorsten-reitz:
Allow to import codelists using the SKOS format from Files or URLs, and allow to assign these code lists to schema elements.

thorsten-reitz:
The codelist reader works in principle, though I have one improvement request: Given that a codelist such as http://www.locationframework.eu/codelist/EuroGeoNamesLocationTypeValue has multiple language `prefLabels`, I would prefer to either use the system setting like for the INSPRIE code lists, or to have a choice, so that I get consistent labels in one language for the concepts.

PR #244, opened 2016-12-12
Added SKOS reader

arunverma87:
PR related to #165


io/plugins/eu.esdihumboldt.hale.io.codelist.skos.test/src/data/test1.rdf
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<rdf:RDF xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#" xmlns:skos="http://www.w3.org/2004/02/skos/core#" xmlns:dc="http://purl.org/dc/elements/1.1/">
+	<skos:Concept rdf:about="http://vocab.to.wetransform/term/W080/1/CONT0006">
```
stempler: Is that correct, concept with an upper C? Did you check the respective schema? [Here](http://vocab.ices.dk/services/rdf/collection/PARAM) for instance a lower case element is used.

io/plugins/eu.esdihumboldt.hale.io.codelist.skos.test/src/eu/esdihumboldt/hale/io/codelist/skos/test/SkosCodeListTest.java
```diff
@@ -0,0 +1,157 @@
...
+	 * 
+	 * test url 2 consisting concepts (case: fallback).
+	 * 
+	 * @throws Exception throws exception if something wrong occurs
+	 */
+	@Test
+	public void testSKOSFromURL2() throws Exception {
+
+		CodeList codeList = readCodeList(
+				URI.create("http://vocab.ices.dk/services/rdf/collection/PARAM"));
```
stempler: I don't think really relying on the online resource for that test makes sense. The online resource may change and thus the number of concepts or their definition.

io/plugins/eu.esdihumboldt.hale.io.codelist.skos/plugin.xml
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<?eclipse version="3.4"?>
+<plugin>
+   <extension
+         point="org.eclipse.core.contenttype.contentTypes">
+      <content-type
```
stempler: The XML content type could be used as parent content type.

io/plugins/eu.esdihumboldt.hale.io.codelist.skos/src/eu/esdihumboldt/hale/io/codelist/skos/reader/SkosCodeListReader.java
```diff
@@ -0,0 +1,55 @@
...
+	protected IOReport execute(ProgressIndicator progress, IOReporter reporter)
+			throws IOProviderConfigurationException, IOException {
+		progress.begin("Loading SKOS code list", ProgressIndicator.UNKNOWN);
+		try {
+			URI loc = getSource().getLocation();
+			InputStream in = getSource().getInput();
+			codelist = new SkosCodeList(in, loc);
+			progress.setCurrentTask("Code list loaded.");
+		} catch (Exception ex) {
+			throw new RuntimeException(ex);
```
stempler: You should handle the exception here, report the error and report the whole process as failed.

io/plugins/eu.esdihumboldt.hale.io.codelist.skos/src/eu/esdihumboldt/hale/io/codelist/skos/reader/SkosCodeList.java
```diff
@@ -0,0 +1,449 @@
...
+	public CodeEntry getEntryByIdentifier(String identifier) {
+		return entriesByIdentifier.get(identifier);
+	}
+
+	@Override
+	public URI getLocation() {
+		return location;
+	}
+
+	private boolean getConceptSchemes() {
```
stempler: All these get(ter) methods w/ boolean or void return type confuse me. Usually this naming convention is used for getters, but these seem to be methods that mostly are there for their side effects.

io/plugins/eu.esdihumboldt.hale.io.codelist.skos.test/src/eu/esdihumboldt/hale/io/codelist/skos/test/SkosCodeListTest.java
arunverma87: I want to test the fallback scenario, which is happening in this link only ( skos api could not load it). So i added this online resource.

io/plugins/eu.esdihumboldt.hale.io.codelist.skos.test/src/data/test1.rdf
stempler: Ok, interesting. Is this then actually loaded by the SKOS API or by the XML mechanism? I think I saw in the XML mechanism that it uses lower case "concept".

arunverma87: It is loaded by skos API itself.

stempler: Ok, so it is not unlikely that the usage of `skos:concept` in other places is incorrect, and we only support it for compatibility reasons.

PR #245, opened 2016-12-12
first version of Merge cell migrator

stempler:


stempler:
Was not tested in detail yet.

Issue #234, opened 2016-12-01
Integrate schematron validator


stempler:
Implementation of #222, #223 via schematron validator.

I see the following tasks as part of this issue:

- Extend target platform with required dependencies
- Extend post encoding validators with the possibility to configure them (here we need at least the Schematron location) + required adaptions to UI and CLI
- Implementation of the Schematron validator itself

Not mandatory, but probably makes sense to do as part of the changes:

- support running multiple post encoding validators + required adaptions to UI and CLI

PR #257, opened 2016-12-16
Integrated schematron validator

florianesser:
refs #234


io/plugins/eu.esdihumboldt.hale.io.schematron/plugin.xml
```diff
@@ -0,0 +1,41 @@
...
+               optional="false">
+            <parameterBinding
+                  class="java.lang.String">
+            </parameterBinding>
+         </providerParameter>
+      </provider>
+   </extension>
+   <extension
+         point="org.eclipse.core.contenttype.contentTypes">
+      <content-type
```
stempler: I think the content type definition should be removed. It is already defined in `eu.esdihumboldt.hale.io.xml`.

io/plugins/eu.esdihumboldt.hale.io.xml/META-INF/MANIFEST.MF
```diff
@@ -13,8 +13,8 @@ Import-Package: com.google.common.collect;version="17.0.0",
...
  org.apache.http.client.fluent;version="4.3.6",
  org.slf4j;version="1.5.11"
-Bundle-ClassPath: lib/xercesImpl-2.9.1.jar,
- .
+Bundle-ClassPath: .
 Export-Package: eu.esdihumboldt.hale.io.xml.validator
 Require-Bundle: eu.esdihumboldt.hale.common.core;bundle-version="2.2.0",
- org.eclipse.core.contenttype;bundle-version="3.4.100"
+ org.eclipse.core.contenttype;bundle-version="3.4.100",
+ xerces.xercesImpl;bundle-version="2.9.1"
```
stempler: Interesting that this still was there. I think the XercesValidator is not used anyway (instead the XMLApiValidator is used), so this should not have any effect on the XML Schema validation.

PR #265, opened 2017-01-03
Support for multiple post-export validators

florianesser:


common/plugins/eu.esdihumboldt.hale.common.headless/src/eu/esdihumboldt/hale/common/headless/transform/ValidationJob.java
```diff
@@ -71,53 +73,63 @@ public ValidationJob(InstanceValidator validator, ReportHandler reportHandler,
...
+
+			// add report to report service
+			reportHandler.publishReport(report);
 
-		// add report to report service
-		reportHandler.publishReport(report);
+			// show message to user
+			if (report.isSuccess()) {
+				// info message
+				log.userInfo(report.getSummary());
```
stempler: If run in the UI, when mutiple validators are used, this will lead to multiple dialogs being displayed to the user, each after one validator succeeded. It might be better to log normal messages here (e.g. `info` instead of `userInfo`) and provide a general summary requesting the user's attention after all validators completed.

common/plugins/eu.esdihumboldt.hale.common.headless/src/eu/esdihumboldt/hale/common/headless/transform/ValidationJob.java
```diff
@@ -71,53 +73,63 @@ public ValidationJob(InstanceValidator validator, ReportHandler reportHandler,
...
+			}
+			else {
+				// error message
+				log.userError(report.getSummary());
```
stempler: Same as above.

ui/plugins/eu.esdihumboldt.hale.ui/src/eu/esdihumboldt/hale/ui/io/instance/InstanceExportWizard.java
```diff
@@ -117,72 +117,84 @@ public InstanceExportWizard() {
...
+									.getService(ReportService.class);
+							repService.addReport(report);
+							// show message to user
+							if (report.isSuccess()) {
+								// info message
+								log.userInfo(report.getSummary());
+							}
+							else {
+								// error message
+								log.userError(report.getSummary()
```
stempler: Same as in the comment related to the ValidationJob - IMHO a user message would be more appropriate after all validators completed (instead of for each).

PR #269, opened 2017-01-06
project validator

stempler:


io/plugins/eu.esdihumboldt.hale.io.validation/plugin.xml
```diff
@@ -0,0 +1,60 @@
...
+<?eclipse version="3.4"?>
+<plugin>
+   <extension
+         point="eu.esdihumboldt.hale.io.action">
+      <action
+            categoryName="Validator configurations"
+            id="eu.esdihumboldt.hale.io.validation.read"
+            name="Load validator configuration"
+            resourceName="validator configuration"
+            type="eu.esdihumboldt.hale.io.validation.ValidationConfigurationReaderImpl">
```
stempler: This should be the interface - you state here what kind of I/O providers are allowed for the action.

io/plugins/eu.esdihumboldt.hale.io.validation/plugin.xml
```diff
@@ -0,0 +1,60 @@
...
+   <extension
+         point="eu.esdihumboldt.util.resource">
+      <resourceType
+            id="validatorConfiguration"
+            name="Validator Configuration">
+      </resourceType>
+   </extension>
+   <extension
+         point="eu.esdihumboldt.hale.io.provider">
+      <provider
```
stempler: Registering the reader for schematron I think would be better done in the schematron bundle. Like that schematron rules can only be imported if the corresponding validator is there.

io/plugins/eu.esdihumboldt.hale.io.validation/plugin.xml
```diff
@@ -0,0 +1,60 @@
...
+            name="Validator configuration">
+         <contentType
+               ref="eu.esdihumboldt.hale.io.schematron.schema">
+         </contentType>
+      </provider>
+      <provider
+            class="eu.esdihumboldt.hale.io.validation.ProjectValidator"
+            id="eu.esdihumboldt.hale.io.validation.projectvalidator"
+            name="Project validator">
+         <contentType
```
stempler: The content type references here raise an interesting point. Right now it would be OK like that, but the validator should actually state that it can perform validation on any content type supported by a validator it delegates to.

ui/plugins/eu.esdihumboldt.hale.ui.application/plugin.xml
```diff
@@ -126,10 +126,6 @@
                name="aboutImage"
                value="images/hale_about.png">
          </property>
-         <property
```
stempler: I'm not sure why Eclipse kicks that out when launching the product via the Product editor. Maybe it is no longer supported with Eclipse Mars, or it does not have to be stated because that's the default value anyway.

io/plugins/eu.esdihumboldt.hale.io.validation/plugin.xml
florianesser: How would it do that? Is there a catch-all content type that matches all others?

stempler: That might be a way. I think at least for text files there is a content type defined by Eclipse, could be that the XML content type extends it. But I would leave it as it is for now, and leave this for later when the problem actually comes up.

stempler:
@florianesser Looks ready to be merged for me. Any objections? Please approve if there aren't any (because I can't as I created the pull request)..

Issue #198, opened 2016-10-12
Provide UI helper to insert project variables


thorsten-reitz:
In all functions that allow usage of project variables, it should be possible to directly insert such variables in a similar way as how variables are inserted in the formatted string function.

thorsten-reitz:
Is there already a build to test that? 

General suggestion to make my life easier ;): when adding the "to be tested", add a comment pointing to the build that includes the feature.

florianesser:
Not yet, I will add a link here as soon as it's done.

thorsten-reitz:
Looks good to me!

PR #271, opened 2017-01-11
Added content assistance for project variables

florianesser:
refs #198


ui/plugins/eu.esdihumboldt.hale.ui.common.definition.editors/src/eu/esdihumboldt/hale/ui/common/definition/editors/DefaultPropertyEditor.java
```diff
@@ -257,6 +260,22 @@ public void widgetSelected(SelectionEvent e) {
 			});
 		}
 
+		// add project variable content assistance
+		if (entity != null) {
```
stempler: I think the condition is not necessary - as stated in the class documentation it may be null if it is unavailable, but should not be needed for the content assist.

PR #277, opened 2017-01-16
add crsProvider to GeometryAdvisor, remove UI dependency

doemming:
add crsProvider to GeometryAdvisor
remove SelectCRSDialog and UI dependency from mssql plugin


stempler:
LGTM

PR #282, opened 2017-01-23
ensure parameter list is not empty (sources too)

### LINKED ISSUE #276: Newly created custom function produces error
It is often not possible to use a custom function that was newly created just before.
There is a NullPointerException in `DefaultCustomPropertyFunction$1.getDefinedParameters():122`.

It seems that editing the function after creation seems to work around the issue.

stempler:
fixes #276


stempler:
retest this please

Issue #188, opened 2016-10-07
Paginate/Partition Requests to WFS as a data source


thorsten-reitz:
When performing "Transform External Data" or "Import Source Data", hale should partition requests, e.g. by using the WFS 2.0 pagination capabilities or by using WFS 1.1.0 vendor-specific parameters.

Ideally, hale first requests `resultType=hits`, and then uses `&startIndex=0&maxFeatures=500` for individual requests. This feature might require the user to select the type of WFS (Geoserver or deegree, ...)

florianesser:
Feature build: https://builds.wetransform.to/job/hale/job/hale~publish(master)/185/

stempler:
New build will be https://builds.wetransform.to/job/hale/job/hale~publish(master)/189/

florianesser:
New feature build will be https://builds.wetransform.to/job/hale/job/hale%7Epublish(master)/191/

thorsten-reitz:
Tested build 191.
Looks good to me.

PR #283, opened 2017-01-23
Do not rely on numberOfFeatures/numberMatched attributes

florianesser:
refs #188

- Use RESULTTYPE=hits query to determine total number of features but do not fail if WFS server does not include a numberOfFeatures/numberMatched attribute in its response
- While storing instances, use InstanceCollection.size() to display total number of features to the user if hasSize() is true
- In case of a communication/processing failure, report exception message


io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/reader/internal/wfs/WfsBackedGmlInstanceCollection.java
```diff
@@ -228,7 +252,7 @@ public WfsBackedGmlInstanceIterator iterator() {
 	 */
 	@Override
 	public boolean isEmpty() {
-		return empty;
+		return size > 0;
```
stempler: This seems wrong

PR #284, opened 2017-01-25
Added switch to activate/deactivate WFS pagination

florianesser:
refs #188


stempler:
I resolved the conflicts in the provider extensions online via GitHub, so I hope that is all right. The conflicts arose because I also did changes there for the interpolation stuff (and also changed the parameter order).

Issue #225, opened 2016-11-23
Merge on large numbers of instances may make HALE unresponsive, lead to OutOfHeapSpace errors or crash application


thorsten-reitz:
Occasionally, a Merge on a large number of instances leads to unstable behaviour, i.e. the application or the JVM crashing. A reporduction case is available internally based on EUROSTAT LUCAS_DE data.

thorsten-reitz:
Tested in Build:
https://builds.wetransform.to/job/hale/job/hale%7Epublish(master)/190/

LGTM, can't produce the crash anymore.

PR #285, opened 2017-01-25
incremental merging of instances

stempler:
refs #225

Improves the memory consumption for large merges, especially if much of the data of the individual instances does not end up in the merged source instance.
Performance is likely lower than before, because the check if a value is unique now has to be done for each instance against all values that have been added before.

Changes the behavior of "auto-detect" slightly, with auto detect now for all properties only unique values are retained. As with the other aspects if the merge this is still restricted to first level properties.


stempler:
> Any particular reason why the change to the Mapping View is also in this PR?

Ah yes, started that yesterday and forgot about it. Didn't see that. That's actually not complete yet I think.

PR #290, opened 2017-01-30
fix restoring bindings for primitive types from hale schema definition

### LINKED ISSUE #272: Binding constraint not correctly restored for primitive types
A binding constraint with a byte type (`byte`) is not correctly restored from a hale schema definition (the class retrieved is `null`). This is probably because it is a primitive type.

It is not clear where the binding comes from exactly - schema was a MS SQL schema. We should check if primitive type binding should be allowed at all (they cannot be null!). Either fix that primitive types are used here or fix resolving the corresponding class. Also check the case of other types that might be critical, like byte arrays (`[B`).

see https://github.com/halestudio/hale/blob/master/common/plugins/eu.esdihumboldt.hale.common.schema/src/eu/esdihumboldt/hale/common/schema/model/constraint/type/factory/BindingFactory.java#L50

stempler:
fixes #272


common/plugins/eu.esdihumboldt.hale.common.schema/src/eu/esdihumboldt/hale/common/schema/model/constraint/factory/DefaultClassResolver.java
```diff
@@ -0,0 +1,53 @@
...
+	private static final ALogger log = ALoggerFactory.getLogger(DefaultClassResolver.class);
+
+	@Override
+	public Class<?> loadClass(String className) {
+		Class<?> result = null;
+
+		try {
+			// loading via ClassUtils also supports primitive type classes
+			result = ClassUtils.getClass(className);
+			// result = Class.forName(className);
```
florianesser: Was the old line left on purpose?

stempler: As `Class.forName` is the obvious candidate here I chose to leave it there, so that it is clear that this  was replaced and that using it is no adequate solution. Might be cleaner though to mention it in the comment and remove the line, true.

Issue #185, opened 2016-10-07
Provide an augmentation function to assign values from a collector


thorsten-reitz:
This function allows assigning values to any target element from any collector defined previously, either in a Groovy Script or in a "Define Collector" augmentation function (see #209). 

Workflow:

1. Click on any target property element
2. After confirmation of the Cell entities, set which defined collector to use 

Use Cases:

* Collect and create references in GML

florianesser:
Feature build: https://builds.wetransform.to/job/hale/job/hale~publish(master)/200/

thorsten-reitz:
LGTM

PR #298, opened 2017-01-31
Function to assign values from a Collector

florianesser:
refs #185 

- Implemented function that assigns references from collected values


common/plugins/eu.esdihumboldt.hale.common.align/plugin.xml
```diff
@@ -445,6 +445,31 @@
             <parameterBinding></parameterBinding>
          </functionParameter>
       </propertyFunction>
+      <propertyFunction
```
stempler: Please move the function definition to the bundle you implemented the function in - only the core functions are defined here.

common/plugins/eu.esdihumboldt.hale.common.align/src/eu/esdihumboldt/hale/common/align/model/functions/AssignReferenceFromCollectorFunction.java
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+package eu.esdihumboldt.hale.common.align.model.functions;
+
+/**
+ * Constants for assign references from collector function
+ * 
+ * @author Florian Esser
+ */
+public interface AssignReferenceFromCollectorFunction {
```
stempler: Same here, please move as well.

common/plugins/eu.esdihumboldt.hale.common.align/plugin.xml
```diff
@@ -445,6 +445,31 @@
...
+            category="eu.esdihumboldt.hale.align.general"
+            icon="icons/augmentation.gif"
+            identifier="eu.esdihumboldt.hale.common.align.collector.assign"
+            name="Assign collected references">
+         <targetProperties>
+            <property
+                  eager="false"
+                  maxOccurrence="-1"
```
stempler: I think this should be `1`, there can only one schema element be supplied as target.
The `minOccurrence` as well, as w/o a target the function cannot be used.

cst/plugins/eu.esdihumboldt.cst.functions.collector/src/eu/esdihumboldt/cst/functions/collector/AssignReferenceFromCollector.java
```diff
@@ -0,0 +1,136 @@
...
+					"Fatal: No collector has been created yet. Check function priority.");
+		}
+
+		ParameterValue collectorName = getParameterChecked(PARAMETER_COLLECTOR);
+		if (collectorName == null || collectorName.isEmpty()) {
+			throw new TransformationException("Fatal: No collector name was specified.");
+		}
+
+		Collector collector = mainCollector.getAt(collectorName.getValue().toString());
+		if (collector == null) {
```
stempler: Actually this will be never null as the collector transparently creates a child collector the first time it is accessed. The check doesn't hurt though.

cst/plugins/eu.esdihumboldt.cst.functions.collector/src/eu/esdihumboldt/cst/functions/collector/AssignReferenceFromCollector.java
```diff
@@ -0,0 +1,136 @@
...
+					collectorName.getValue().toString()));
+		}
+
+		MultiValue result;
+
+		List<Object> collectedReferences = helper.extractCollectedValues(collector);
+		final TypeDefinition resultPropertyType = resultProperty.getDefinition().getPropertyType();
+		if (resultPropertyType.getConstraint(HasValueFlag.class).isEnabled()) {
+			// Target property can take values, assign directly
+			result = new MultiValue(collectedReferences);
```
stempler: In addition, in case the target element is a reference, we could do the reference processing on the IDs (e.g. when the target is the `href` attribute directly).

common/plugins/eu.esdihumboldt.hale.common.schema/src/eu/esdihumboldt/hale/common/schema/model/constraint/property/Reference.java
```diff
@@ -122,4 +122,15 @@ public Object extractId(Object refValue) {
...
 	}
 
+	/**
+	 * Converts the identifier of a referenced object to the reference value.
+	 * The default implementation just returns the given reference.
+	 * 
+	 * @param reference the reference
+	 * @return the reference value
+	 */
+	public Object toId(Object reference) {
```
stempler: I think the method name is a bit misleading. It should convert an ID to a reference. (Opposed to the existing extractId method that converts a reference to an ID)

PR #303, opened 2017-02-02
[WIP] Release preparations for 3.1.1

florianesser:


stempler:
@florianesser @thorsten-reitz So far we did not have What's new pages for bugfix releases, but these were added on top of the page for the minor release. I would vote for keeping that approach. What's your opinion?

PR #306, opened 2017-02-02
Rename AssignReferencesFromCollector

florianesser:
Rename to AssignFromCollector to better reflect what the function
actually does, as the function can assign collected values to any kind
of target property and is not limited to assigning references.

### DIFF
--- a/cst/plugins/eu.esdihumboldt.cst.functions.collector/src/eu/esdihumboldt/cst/functions/collector/explanations/AssignFromCollectorExplanation.java
+++ b/cst/plugins/eu.esdihumboldt.cst.functions.collector/src/eu/esdihumboldt/cst/functions/collector/explanations/AssignFromCollectorExplanation.java
@@ -20,7 +20,6 @@ import java.util.Map;
 
 import eu.esdihumboldt.hale.common.align.model.Cell;
 import eu.esdihumboldt.hale.common.align.model.CellUtil;
-import eu.esdihumboldt.hale.common.align.model.ChildContext;
 import eu.esdihumboldt.hale.common.align.model.Entity;
 import eu.esdihumboldt.hale.common.align.model.impl.mdexpl.MarkdownCellExplanation;
 import eu.esdihumboldt.hale.common.core.service.ServiceProvider;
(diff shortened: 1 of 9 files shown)


cst/plugins/eu.esdihumboldt.cst.functions.collector/src/eu/esdihumboldt/cst/functions/collector/explanations/AssignFromCollectorExplanation.java
```diff
@@ -0,0 +1,79 @@
...
+
+		FunctionDefinition<? extends ParameterDefinition> function = loadFunction(
+				cell.getTransformationIdentifier(), provider);
+
+		// Default values for all bindings
+		binding.put("_constraintsEvaluated", false);
+		binding.put("_hasValue", false);
+		binding.put("_isReference", false);
+
+		if (!function.getTarget().isEmpty() && !cell.getTarget().isEmpty()) {
```
stempler: I don't think you actually need the function information here. You know that your function may only have one target, so no need to differentiate. You can use `CellUtil.getFirstEntity(cell.getTarget())` to get the target entity. That also avoids the problem that `cell.getTarget()` can (in theory) be null.

stempler:
retest this please

cst/plugins/eu.esdihumboldt.cst.functions.collector/src/eu/esdihumboldt/cst/functions/collector/explanations/AssignFromCollectorExplanation.java
```diff
@@ -41,38 +40,32 @@
...
-				ChildContext childContext = entity.getDefinition().getPropertyPath().iterator()
-						.next();
-				PropertyDefinition resultProperty = childContext.getChild().asProperty();
-				TypeDefinition resultPropertyType = resultProperty.getPropertyType();
-
-				boolean isReference = resultProperty.getConstraint(Reference.class).isReference();
-				binding.put("_isReference", isReference);
+		Entity entity = CellUtil.getFirstEntity(cell.getTarget());
+		if (entity != null) {
+			ChildContext childContext = entity.getDefinition().getPropertyPath().iterator().next();
```
stempler: That would only be the first element in the property path, and not necessarily the actual property that is mapped (could be a parent property). The best way is to do `entity.getDefinition().getDefinition()` and use that - that is always the last element of the path. You can just check if it is a `PropertyDefinition` and cast it accordingly.

Issue #313, opened 2017-02-08
Arbitrary SQL queries as data source


stempler:
Based on the existing JDBC database access there should be also the possibility to use a custom SQL query. What needs to be done is:

- support importing an SQL query as schema/type
- support importing data from the database based on that type
- add needed specific UI support

thorsten-reitz:
What enhancements are still open on this issue?

stempler:
Main question is if we want to keep the behavior that we have there right now regarding loading the data (importing the data from the database only once, implying that the same queries are used as specified for the schema and requiring a reload when a new SQL schema type is added).

That also implies how we might deal with possible improvements regarding the possibility to
- adapt the query for loading data
- change the query after initial loading of the data
- change the schema query after initial loading of the schema type

thorsten-reitz:
I've created enhancements and will now close this issue for this milestone.

PR #314, opened 2017-02-08
first implementation of SQL based schema/instance reader

stempler:
refs #313 

Right now, loading instances works via the normal JDBC instance reader. For schema types based on SQL queries it uses exactly that SQL query.


cst/plugins/eu.esdihumboldt.cst.functions.collector/plugin.xml
```diff
@@ -15,7 +15,6 @@
       <propertyFunction
             category="eu.esdihumboldt.hale.align.general"
             cellExplanation="eu.esdihumboldt.cst.functions.collector.explanations.AssignFromCollectorExplanation"
-            icon="icons/augmentation.gif"
```
florianesser: Is there another icon available that would fit? Or should we look for one/create one specifically for this function?

stempler: I removed this because the icon does not exist in the bundle and this produced a warning. If the function is an augmentation (no source) it will have that icon by default.
If we come up with a better fitting icon we should add it, but I think that has not the highest priority.

stempler:
retest this please

stempler:
retest this please

stempler:
retest this please

stempler:
An important point that still needs to be handled is to allow importing multiple queries from the same database. hale blocks that by default because it assumes it is the same resource.

stempler:
retest this please

stempler:
retest this please

stempler:
retest this please

Issue #300, opened 2017-02-01
Spatial index for data sources


stempler:
As part of the data source processing integrate building a spatial index. The index should store `InstanceReference`s by bounding box. The basic capabilities are:

- retrieve `InstanceReference`s by bounding box
- retrieve `InstanceReference`s by bounding box and type

Access to the index should at least be available in Groovy transformation functions.

stempler:
In hale studio, the index should only be created once (or when the source data is reloaded).

Investigate if we can use that opportunity to introduce a more general concept of source data processing/indexing as prerequisite for the transformation.

thorsten-reitz:
I tested the [build](https://builds.wetransform.to/job/hale/job/hale%7Epublish(master)/205/), seems to work OK. Should we close this ticket and put any open work into enhancement issues?

PR #315, opened 2017-02-10
Add spatial index service

florianesser:
Added a spatial index service based on the use case described in #300.


stempler:
Regarding the build error - probably the `ext/geom` folder is not picked up for that. You could either add it [here](https://github.com/halestudio/hale/blob/master/build/build.gradle#L94) or move it to `ext/styledmap`.

florianesser:
Ah, thanks! I'll add the folder to the build.gradle then.

florianesser:
retest this please

stempler:
retest this please

PR #317, opened 2017-02-13
Add extension for instance processing

florianesser:


florianesser:
retest this please

common/plugins/eu.esdihumboldt.hale.common.instance/src/eu/esdihumboldt/hale/common/instance/processing/InstanceProcessor.java
```diff
@@ -0,0 +1,50 @@
...
+	 * @param services the service provider
+	 */
+	void setServiceProvider(ServiceProvider services);
+
+	/**
+	 * Processing before an instance is stored.
+	 * 
+	 * @param instance the instance
+	 */
+	void beforeStore(Instance instance);
```
stempler: What do you see as use cases for having both before and after methods? That the instances are stored is strongly related to the context of the StoreInstancesJob, but instances could be processed also in a different context (e.g. when no OrientDB is used for the transformation).

common/plugins/eu.esdihumboldt.hale.common.instance.orient/src/eu/esdihumboldt/hale/common/instance/orient/storage/StoreInstancesJob.java
```diff
@@ -120,6 +128,10 @@ public IStatus run(IProgressMonitor monitor) {
 			// use intent
 			db.declareIntent(new OIntentMassiveInsert());
 
+			final InstanceProcessingExtension ext = new InstanceProcessingExtension(
+					serviceProvider);
+			final List<InstanceProcessor> processors = ext.getInstanceProcessors();
```
stempler: An improvement for later could be that we are more flexible regarding which processors are actually run (e.g. here only those that are required for the transformation). For now, and as this is also running in the UI, I think it is best to run all as you did here.

common/plugins/eu.esdihumboldt.hale.common.instance/src/eu/esdihumboldt/hale/common/instance/processing/InstanceProcessor.java
```diff
@@ -0,0 +1,50 @@
...
+import eu.esdihumboldt.hale.common.core.service.ServiceProvider;
+import eu.esdihumboldt.hale.common.instance.model.Instance;
+import eu.esdihumboldt.hale.common.instance.model.InstanceReference;
+
+/**
+ * {@link Instance} processor
+ * 
+ * @author Florian Esser
+ */
+public interface InstanceProcessor {
```
stempler: Some instance processors might need the information when the processing (of the current batch of instances) is finished. The processor might for instance want to free up resources that it allocated. Do you think that could be easily added? We could use the `Closable` interface to extend from.

stempler:
@florianesser It's a bit annoying that the builds are aborted right now. I wonder why they are taking so long and if it is because they get stuck every time or not. It happens sometimes that the build just gets stuck and it would just be hanging around for days if we had not introduced the automatic abortion time limit. You can change the Job configuration to increase the time limit and see if that helps, but usually successful builds are around 15 min while the limit is 30 min I think.

florianesser:
@stempler The last line in `hale_tests.out` reads

`13.02.2017 11:30:41 INFO  e.e.h.c.t.d.c.HaleDockerClient.startContainer(140) - Preparing container for image kartoza/postgis...`

Could that be the point where it's stuck?

florianesser:
retest this please

common/plugins/eu.esdihumboldt.hale.common.instance/src/eu/esdihumboldt/hale/common/instance/processing/InstanceProcessor.java
florianesser: My idea was to provide hooks for 
- instance manipulation before storage in OrientDB (currently also handled by `StoreInstancesJob.processInstances` which I thought could maybe be replaced)
- making use of the `InstanceReference` after storage in OrientDB
and needing only one implementation to cover both.

I agree that this is tightly coupled to the StoreInstancesJob context. If we want to make this more generic and extendable, we could – instead of having before/after methods – add execution phases like e.g. `ON_LOAD` (for index population) or `AFTER_TYPE_TRANSFORMATION` (for filling collectors).

PR #320, opened 2017-02-21
Implement spatial index

florianesser:
Add spatial index implementation (cref. #300) and functionality to automatically index instances stored to OrientDB. The index is exposed to Groovy functions via `geom` helper functions and the new `_spatialIndex` binding.


florianesser:
retest this please

common/plugins/eu.esdihumboldt.hale.common.headless/src/eu/esdihumboldt/hale/common/headless/transform/Transformation.java
```diff
@@ -457,7 +462,27 @@ public void done(IJobChangeEvent event) {
...
+					serviceProvider);
+			final List<InstanceProcessor> processors = ext.getInstanceProcessors();
+
+			ResourceIterator<Instance> it = sourceToUse.iterator();
+			try {
+				while (it.hasNext()) {
+					Instance instance = it.next();
+
+					ResolvableInstanceReference resolvableRef = new ResolvableInstanceReference(
+							new PseudoInstanceReference(instance), sourceToUse);
```
stempler: Having the memory fill with all the instances is a problem. Maybe we should instead introduce a flag for the execution if instance processing is needed (or even which kind is needed) and then use the OrientDB storage.

cst/plugins/eu.esdihumboldt.cst.functions.geometric/src/eu/esdihumboldt/cst/functions/geometric/GeometryHelperFunctions.groovy
```diff
@@ -150,6 +153,69 @@ class GeometryHelperFunctions {
...
+	static boolean _boundaryCovers(Map args) {
+		return args.geometry.getGeometry().getBoundary().covers(args.line.getGeometry());
+	}
+
+	/**
+	 * Specification for the _spatialIndexQuery function
+	 */
+	public static final HelperFunctionSpecification _spatialIndexQuery_spec = new HelperFunctionSpecification(
+	"Query a spatial index. The bounding box of the given geometry is computed and used to query the spatial index.", "Collection of instances that match the query.",
+	new HelperFunctionArgument("spatialIndex",
```
stempler: Can't we make the service available to the helper function? I think it is a good idea anyway to make the `ServiceProvider` available there.

cst/plugins/eu.esdihumboldt.cst.functions.geometric/src/eu/esdihumboldt/cst/functions/geometric/GeometryHelperFunctions.groovy
```diff
@@ -150,6 +153,69 @@ class GeometryHelperFunctions {
...
+	 * Query a spatial index
+	 *
+	 * @param args the function arguments
+	 * @return a Collection of Instances that match the spatial query
+	 */
+	@Nullable
+	static Collection<Instance> _spatialIndexQuery(Map args) {
+		def result = []
+
+		Geometry geometry = args.geometry.getGeometry()
```
stempler: I would vote for a consistent behavior w/ the other geometry helper functions regarding what can be passed in for a geometry. For instance that also Instances can be passed in. It should be possible to reuse the functionality for that (as this should already be done in a couple of other places).

ui/plugins/eu.esdihumboldt.hale.ui/src/eu/esdihumboldt/hale/ui/service/project/internal/ProjectServiceImpl.java
```diff
@@ -1018,6 +1021,21 @@ public void run(IProgressMonitor monitor)
 				InstanceService is = PlatformUI.getWorkbench().getService(InstanceService.class);
 				is.dropInstances();
 
+				// inform instance processors of the drop
+
+				// XXX This could be done in a more general way by listening to
+				// datasetChanged events from InstanceService
```
stempler: Yes, I think so to. We only need to find a place where the corresponding `InstanceServiceListener` may "live".

cst/plugins/eu.esdihumboldt.cst.functions.geometric/src/eu/esdihumboldt/cst/functions/geometric/GeometryHelperFunctions.groovy
florianesser: Yes, I'd definitely prefer that, too, but I didn't see a way how, since the helper function is static and the `SpatialIndexService` has project scope (so I can't use `HalePlatform.getServiceProvider()`). Is there a way to get the service in a static way so that it works both in the UI and the CLI?

common/plugins/eu.esdihumboldt.hale.common.headless/src/eu/esdihumboldt/hale/common/headless/transform/Transformation.java
florianesser: Wouldn't this require rather in-depth knowledge of the performed transformations on the side of the user who uses the CLI?

Maybe it would be sufficient for the moment to simply not do instance processing in cases when OrientDB is not used, since there currently is exactly one `InstanceProcessor` (`SpatialIndexInstanceProcessor`) and its state can only be accessed from Groovy functions, which (if I understand it correctly) will lead to OrientDB usage anyway.

PR #322, opened 2017-02-24
slightly more robust boundaryCovers helper

stempler:


cst/plugins/eu.esdihumboldt.cst.functions.geometric/src/eu/esdihumboldt/cst/functions/geometric/GeometryHelperFunctions.groovy
```diff
@@ -167,9 +167,22 @@ class GeometryHelperFunctions {
...
+
+		if (!geom) {
+			throw new IllegalArgumentException('No geometry found for geometry argument')
+		}
+		if (!line) {
+			throw new IllegalArgumentException('No geometry found for line argument')
+		}
+
+		/*
+		 * FIXME check if geometries have the same SRS
```
florianesser: Should we do something about that right now, like compare the SRIDs of both geometries? And what do we do if they don't match?

stempler: If they don't match the best thing would be to convert one geometry to the reference system to the other, if that makes sense.
Minimal solution would be throwing an error in that case, with the use case tackled with that right now that would probably be enough.

PR #329, opened 2017-03-06
add fallback for size calculation

doemming:
calculate table size by counting query result if count query does not
work


stempler:
Did you hit that case? Not sure if it makes sense to do that with a potentially large query. JDBC will probably then fetch all objects just for getting the size.

doemming:
yes, this happened when I used a complex stored procedure in a query. The query works, but the size calculation did not, so I was looking for a fallback that works in any case (as long as the original query works).

florianesser:
If the reason why you need the result set size is to prevent `isEmpty() == true` you could instead simply check if there is at least one result in the result set and if not set `count = 0`. That way you won't have to iterate the whole (and possiblye huge) result set and can still prevent the problem caused by a wrong `isEmpty()` result.

io/plugins/eu.esdihumboldt.hale.io.jdbc/src/eu/esdihumboldt/hale/io/jdbc/JDBCTableCollection.java
```diff
@@ -304,8 +304,19 @@ public int size() {
 
 			return count;
 		} catch (SQLException e) {
-			log.warn("Could not determine query size", e);
-			return UNKNOWN_SIZE;
+			log.warn("Could not determine query size by count query");
+			try (Connection connection = createConnection()) {
+				Statement st = connection.createStatement();
+				ResultSet res = st.executeQuery(sqlQuery);
+				int count = 0;
```
florianesser: For example like so:
`				if (!res.next()) {
					return 0;
				}
				else {
					return UNKNOWN_SIZE;
				}
`

io/plugins/eu.esdihumboldt.hale.io.jdbc/src/eu/esdihumboldt/hale/io/jdbc/JDBCTableCollection.java
```diff
@@ -304,8 +304,19 @@ public int size() {
 
 			return count;
 		} catch (SQLException e) {
-			log.warn("Could not determine query size", e);
-			return UNKNOWN_SIZE;
+			log.warn("Could not determine query size by count query");
+			try (Connection connection = createConnection()) {
+				Statement st = connection.createStatement();
```
florianesser: Additionally, you could add `st.setMaxRows(1)` to add a server side limit to the query (if the JDBC driver supports that).

stempler:
I think what @florianesser proposed would be a good solution that serves both ends.

doemming:
sounds good, I will test if this works.

io/plugins/eu.esdihumboldt.hale.io.jdbc/src/eu/esdihumboldt/hale/io/jdbc/JDBCTableCollection.java
```diff
@@ -304,8 +304,20 @@ public int size() {
...
+				ResultSet res = st.executeQuery(sqlQuery);
+				if(res.next()) {
```
florianesser: Please review the formatting (GitHub doesn't do that automatically if you edit via the browser)

io/plugins/eu.esdihumboldt.hale.io.jdbc/src/eu/esdihumboldt/hale/io/jdbc/JDBCTableCollection.java
stempler: @doemming What about this? Can that be added or might it cause problems?

doemming: how do I know if the current jdbc driver supports this?

florianesser: If the driver does not support it, calling `setMaxRows` will simply have no effect, but if the driver does support it, it will limit the amount of work performed on the server.

doemming: thx @florianesser .  I just committet setMaxRows(1) now.

Issue #291, opened 2017-01-30
Log in to hale connect via hale studio


thorsten-reitz:
As a hale user, I want to be able to log in to my hale connect account via hale studio to upload/download transformation projects and access custom functions.

Workflow:
1. On the welcome page, enter your haleconnect account info (email and password). Choose whether credentials should be remembered.

Alternate Workflow:
1. In window -> preferences, go to the Account tab
2. Enter your credentials and press "Login". The system gives feedback ("Login successful/failed") 

Definition of Done:
1. Login uses haleconnect.com User Service to obtain JWT
2. JWT is then stored and used to access resources or to upload resources

Enhancements:
1. Also enable registration with haleconnect.com via the welcome page.

florianesser:
[Feature build](https://builds.wetransform.to/job/hale/job/hale~publish(master)/213/)

florianesser:
[Feature build](https://builds.wetransform.to/job/hale/job/hale%7Epublish(master)/214/)

thorsten-reitz:
I tested build 214 and would suggest the following enhancements:

- On the welcome screen, show "Logged in as <username>"
- In the preferences, my given usernae and password from the welcome page are not stored, and my login status is not shown correctly (It still says I can login)
- A logout doesn't seem to be possible?

stempler:
I assume the SWT version of the welcome page part is not there yet? Currently it looks to me like this on my system:

![screenshot from 2017-03-21 10-50-57](https://cloud.githubusercontent.com/assets/375475/24141672/76205c1a-0e24-11e7-9f97-5a28b2648d0a.png)

I also think it would be good to have the current status (am I logged in?) reflected in the settings.

florianesser:
[Feature build](https://builds.wetransform.to/job/hale/job/hale%7Epublish(master)/217/) with enhancements, without SWT version of Welcome Page

PR #334, opened 2017-03-16
Add hale connect login support

florianesser:
Basic hale connect login support is provided via a login dialog and a login
section on the Welcome page. Support for storing user name, password
and the base URL of hale connect's user service is also added.

The new HaleConnectService is used as a facade to the hale connect API
calls and to store the state of the hale connect interaction (like the JWT).

cref. #291


stempler:
Platform update is available in
https://gitlab.wetransform.to/hale/hale-build-support/raw/0a486813a6a507f7bde50cd44aa619e07427fbb6/updatesites/platform

See https://github.com/halestudio/hale-platform/pull/10

florianesser:
The master password prompt appears as soon as the stored hale connect password is first accessed by the application. If the Welcome Page is shown at startup, the prompt will appear immediately since the input fields are filled it with the stored values.

PR #342, opened 2017-03-24
always use cached schema as fall-back if loading source fails

stempler:
Subclasses may disable that behavior.


stempler:
retest this please

PR #343, opened 2017-03-24
use shorter IDs for core value types

stempler:
@eleazar-araujo Did this mainly to have shorter value type names in the Json representation of the schema. For instance `properties` instead of `eu.esdihumboldt.hale.common.core.properties`.


eleazar-araujo:
Understood. Thanks

PR #348, opened 2017-03-27
allow null values in InstanceAccessor

stempler:
Added two ways of dealing with nulls in InstanceAccessor:

- treat null values as non-existing values (default behavior)
- treat null values as they are and list them

There is no way though to discern no value and a null value even with the later setting, if the `first()` or `value()` access methods are used.


util/plugins/eu.esdihumboldt.util.groovy/src/eu/esdihumboldt/util/groovy/paths/PathWithNulls.java
```diff
@@ -0,0 +1,81 @@
+/*
+ * Copyright (c) 2013 Simon Templer
```
florianesser: Outdated header

PR #350, opened 2017-03-29
add MATERIALIZED VIEW table type for reading schema

stempler:
- [ ] test if this actually works
- [ ] check if this poses problems for databases w/o that table type (otherwise only add it for PostgreSQL)

see also https://sourceforge.net/p/schemacrawler/discussion/495990/thread/09186044/


thorsten-reitz:
Will not be included in 3.3.0 since a test case is missing.

PR #352, opened 2017-04-03
support for extra dependencies defined in manifest

stempler:
...to be able to have correct dependencies for Maven artifacts of bundles
that bundle libraries internally.

Added extra dependency for schematron bundle.

As a result it should be possible to remove the `org.opengis.cite:schema-utils:1.8` dependency in https://github.com/halestudio/hale-cli/pull/24.


stempler:
Also extended the [wiki page](https://github.com/halestudio/hale/wiki/Extend-the-target-platform) with documentation on using this.

PR #355, opened 2017-04-13
Add support for loading projects from hale connect

florianesser:


io/plugins/eu.esdihumboldt.hale.io.haleconnect/src/eu/esdihumboldt/hale/io/haleconnect/HaleConnectInputSupplier.java
```diff
@@ -0,0 +1,89 @@
...
+		} catch (ApiException e) {
+			throw new RuntimeException(e.getMessage(), e);
+		}
+
+		return new BufferedInputStream(new FileInputStream(response.getData()));
+	}
+
+	@Override
+	public URI getLocation() {
+		return URI.create(MessageFormat.format("{0}:project:{1}:{2}:{3}", SCHEME_HALECONNECT,
```
florianesser: The format of the location URI here is open for discussion. My rationale behind using an URN-like format is that I didn't want to bind the project tightly to any REST service URL which might change down the road.

io/plugins/eu.esdihumboldt.hale.io.haleconnect/src/eu/esdihumboldt/hale/io/haleconnect/internal/HaleConnectServiceImpl.java
```diff
@@ -156,11 +174,197 @@ public void removeListener(HaleConnectServiceListener listener) {
...
+			return null;
+		}
+
+		List<String> orgIds = this.getSession().getOrganisationIds();
+		if (orgIds.isEmpty()) {
+			return null;
+		}
+
+		// XXX Cannot handle multiple organisations!
+		return orgIds.iterator().next();
```
florianesser: My impression was that multiple organisation roles are not supported by hale connect at the moment. If that changes, the user would need a way to choose an "active" organisation in hale connect.

florianesser:
As soon as the required dependencies are available in `hale-platform` (halestudio/hale-platform#11), the target platform needs to be updated accordingly.

io/plugins/eu.esdihumboldt.hale.io.haleconnect/src/eu/esdihumboldt/hale/io/haleconnect/HaleConnectInputSupplier.java
stempler: I'm not sure if having a special kind of input supplier will work in every case, i.e. when the project is not loaded via the source, but via recent project files or the command line.

In principle the URI format looks good to me. If the custom input supplier does not work out, a custom `URLStreamHandler` might be an option. I think there is an extension point or service in OSGi that can be used, though we would need to create a solution that is usable outside OSGi as well. And it would need to be extendable, due to the nature of `URL.setURLStreamHandlerFactory(...)`.

stempler:
My first tests look good, though I was not able to test loading a project w/o the using the haleconnect source.

Loading one of the projects timed out, this may need to be tweaked, though in that case it wasn't possible to download the project archive via the web interface either.

io/plugins/eu.esdihumboldt.hale.io.haleconnect/src/eu/esdihumboldt/hale/io/haleconnect/HaleConnectInputSupplier.java
stempler: An alternative to a custom `URLStreamHandler` could also be using a custom reader, that knows how to handle the URL scheme. This would be similar for instance to the JDBC reader.
This approach might be more flexible in that it for instance allows for specific error handling.

PR #356, opened 2017-04-18
Prevent static compile errors

florianesser:
Groovy compiler complains about accessing methods of  the `geometry`property of
methods returning `GeometryProperty<?>`.

An alternative solution where we would be able to keep the `@CompileStatic`s would 
be to let all relevant methods return `GeometryProperty<? extends Geometry>`
instead.


stempler:
Where does this surface as a problem?
Are these only errors displayed in Eclipse or are there actual compile or runtime errors?

florianesser:
Jenkins doesn't complain, so I assume it's an Eclipse-specific thing, especially since this surfaced only recently after an update of my Eclipse Groovy plugin. I thought it might be a good idea to clean this up anyway, since Eclipse reports this as errors and the type access is actually a bit unclean.

stempler:
`GeometryProperty<?>` and `GeometryProperty<? extends Geometry>` should essentially be the same, due to the way that `GeometryProperty` is defined, or do I miss something?

PR #358, opened 2017-04-21
improve handling of loaded reports

stempler:
- Being unable to determine a report session ID from a file name is not an error
- Don't use formatted date for report session with ID/timestamp 0


ui/plugins/eu.esdihumboldt.hale.ui.views.report/src/eu/esdihumboldt/hale/ui/views/report/ReportListLabelProvider.java
```diff
@@ -93,7 +93,11 @@ public String getText(Object element) {
 			return ((Report<?>) element).getTaskName();
 		}
 		else if (element instanceof ReportSession) {
-			return df.format(new Date(((ReportSession) element).getId()));
+			long id = ((ReportSession) element).getId();
+			if (id == 0) {
```
florianesser: We could use a constant for the unknown/undetermined id here instead of a magic number

PR #377, opened 2017-05-09
support loading hale schema definitions w/ multiple schemas

stempler:


stempler:
retest this please

stempler:
retest this please

common/plugins/eu.esdihumboldt.hale.common.schema.persist/src/eu/esdihumboldt/hale/common/schema/persist/hsd/HaleSchemaUtil.java
```diff
@@ -33,4 +41,40 @@ public static InputStream getHaleSchemaXSD() {
...
+				schema.getTypes().forEach(type -> {
+					if (reporter != null) {
+						// check if type is already there
+						QName name = type.getName();
+						if (result.getType(name) != null) {
+							reporter.warn(
+									MessageFormat.format("Multiple definitions of type {0}", name));
+						}
+					}
+					result.addType(type);
```
florianesser: It appears that the last schema in the list to define a type wins here. Is the order of the schemas predictable? If not, couldn't this lead to undeterministic results?

stempler: I think the most likely situations here are:

1. There is no conflict (otherwise there was one in hale too, after loading the schemas, which we try to avoid for instance by allowing to load only one schema)
2. There is a conflict but the definitions are the same
3. There is a conflict and definitions differ

At that point we can't do much though I think. What would you think about raising the message from a warning to an error?

florianesser: I think that's definitely fine for the use case at hand. 

A possible problem that might arise in the future: if `combineSchema` is used in another context where `reporter` is `null`, this will silently use the last definition in case of multiple definitions. 

I think we can simply document that behaviour for the time being. If we want to guard against that case, we could throw an exception in the `reporter == null` case instead.

Issue #331, opened 2017-03-10
Partition GML/WFS FeatureCollection to multiple files when writing


thorsten-reitz:
Very large GML files are often problematic in all kinds of client/downstream applications, such a QGIS and hale connect.

Thus, add an option to the GML writers to write a FeatureCollection to 1..n files, all in one folder, adding a postfix counter to the filenames. 

Definition of Done:

* Allow the user to set a limit in terms of file size (when exceeded, do not write additional features) or in terms of feature counts per file. 
* Modify all in-document references in `href` attributes so that they point to the right file, ideally without requiring a second pass through the generated files (which could definitely be done when implementing the "features per file" option above).

stempler:
From the description I gather that the existing functionality for splitting into multiple files that is used for instance in the WFS export is not supposed to be used here, right?

Regarding the handling of feature references my question is what we want to support regaeding the further processing of the resulting files?
For instance does hale itself not have any support for these cross-references and also uploading these different files via WFS-T will not work, as references to other files will not be resolvable.

thorsten-reitz:
In principle, maintaining references and consistent state when writing to a WFS-T in many individual transactions would require an analysis of the graph of references between instances, I assume? Based on the type of graph, one of the following would have to happen:

Tree structures or directed acyclic graphs:

1. write root node first
2. incrementally write first-level children pointing to root node
3. incrementally write n-th level children pointing to n-1-th level children
4. Update root element with references to first-level children

Cyclic graphs:

1. identify node withthe highest cardinality and write that
2. Add nodes (in batches) that are once-removed from that center node, with references to the center node
3. Iterate over graph until all nodes are added
4. Update all nodes with additional references

thorsten-reitz:
There is a need to differentiate two use cases:

1. create a set of files for usage as files (e.g. in desktop GIS)
2. create a set of files to be used to fill a deegree service via WFS-T

Additional requirements:

* Clarify: deegree needs special GML ID formats including a FeatureType prefix and generates these on the fly (?)

thorsten-reitz:
minimum implementation: offer current partitioning mechanism (which keeps all objects that have references together in one fragment) and test how that works in various clients.

florianesser:
[Feature build](https://builds.wetransform.to/job/hale/job/hale%7Epublish(master)/267/). This build does not support partitioning when using `Transform external data` or the CLI.

thorsten-reitz:
[...]

PR #400, opened 2017-06-13
Add support for partitioning GML to multiple files

florianesser:
Add basic support for partitioning GML to multiple output files. This covers the basic requirements laid out in #331 which keeps all inter-related objects together in one part, regardless of the given number of maximum features per file.

Support for validating all generated output files is also added by replacing the original `LocatableOutputSupplier` object on-the-fly with a `MultiLocationOutputSupplier` as a hint for the validation mechanism that multiple files need to be validated.

### DIFF
--- a/ui/plugins/eu.esdihumboldt.hale.ui/src/eu/esdihumboldt/hale/ui/io/instance/InstanceExportWizard.java
+++ b/ui/plugins/eu.esdihumboldt.hale.ui/src/eu/esdihumboldt/hale/ui/io/instance/InstanceExportWizard.java
@@ -18,6 +18,8 @@ package eu.esdihumboldt.hale.ui.io.instance;
 
 import java.io.File;
 import java.io.InputStream;
+import java.io.OutputStream;
+import java.net.URI;
 import java.text.MessageFormat;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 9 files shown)


ui/plugins/eu.esdihumboldt.hale.ui/src/eu/esdihumboldt/hale/ui/io/instance/InstanceExportWizard.java
```diff
@@ -34,6 +36,8 @@
 import eu.esdihumboldt.hale.common.core.io.supplier.FileIOSupplier;
 import eu.esdihumboldt.hale.common.core.io.supplier.Locatable;
 import eu.esdihumboldt.hale.common.core.io.supplier.LocatableInputSupplier;
+import eu.esdihumboldt.hale.common.core.io.supplier.LocatableOutputSupplier;
+import eu.esdihumboldt.hale.common.core.io.supplier.MultiLocationOutputSupplier;
```
stempler: As the wizard is the only place where the validation is adapted, I assume the validation as part of the headless transformation and possibly the validation via "Transform external data" will not work?
Do you think there is an easy way to apply that there too?

stempler: All right, so for now let's keep in mind that this does not work and leave that for an improvement.

PR #407, opened 2017-06-28
remove warnings

stempler:
Down to three warnings now (in my workspace).


io/plugins/eu.esdihumboldt.hale.io.haleconnect.ui/src/eu/esdihumboldt/hale/io/haleconnect/ui/projects/ShareProjectWizard.java
```diff
@@ -47,14 +45,14 @@
 
 	private static final ALogger log = ALoggerFactory.getLogger(ShareProjectWizard.class);
 
-	private final HaleConnectService haleConnect;
+//	private final HaleConnectService haleConnect;
```
florianesser: Oh, and I think you can remove that completely, unless you kept it on purpose?

stempler: Just thought that it is there because it might be needed later, and adapting the constructor now just to revert it some point might not make sense. But if you think it's obsolete then surely it can be removed.

PR #411, opened 2017-07-17
Treat save as export in case of non-file target

florianesser:
Guard against cases where the ProjectWriter does not have a target location or
the target location is not a file.


florianesser:
Retest this please

PR #412, opened 2017-07-18
Don't fail loading a project in headless mode when code lists cannot be imported

florianesser:


florianesser:
retest this please

PR #443, opened 2017-09-05
limit messages in reports

stempler:
In some cases reports with several hundred thousand of even millions of messages are generated. Due to the report model all of these are kept in memory. Usually the information in the messages involves a lot of redundancy.

To reduce the need for memory, improve the response times in the UI and to make the report analysis easier, this PR introduces the notion of message limits in reports. Messages that exceed this limit are counted but not included explicitly in the report.


common/plugins/eu.esdihumboldt.hale.common.core/src/eu/esdihumboldt/hale/common/core/report/impl/AllInMemoryReporter.java
```diff
@@ -0,0 +1,176 @@
+/*
+ * Copyright (c) 2012 Data Harmonisation Panel
```
florianesser: Severe case of outdated header :)

stempler: Yes, but that's ok because that's actually the original `DefaultReporter` class, just renamed.

florianesser:
This is currently an absolute message cap, right? Does that mean that there is a probability that important messages may get lost? If so, would it make sense to add some kind of debug switch to disable the cap?

stempler:
Wouldn't hurt to add, though usually, when you have 1000 errors it's enough to look at those first. But it can be definitely helpful.

What would you propose? A switch or custom caps?

florianesser:
Yes, for errors I agree. But at least the info messages are probably something that the user can't or won't do anything about, right? I think custom caps with the option to set it to 0 would work.

Maybe in the long run we can group messages by type/kind and cap by group, so that rare message will always appear in the log.

PR #447, opened 2017-09-19
fix jdbc collection size and add errors when storing instances to report

stempler:


io/plugins/eu.esdihumboldt.hale.io.jdbc/src/eu/esdihumboldt/hale/io/jdbc/JDBCTableCollection.java
```diff
@@ -316,8 +322,10 @@ public int size() {
 					return 0;
 				}
 			} catch (SQLException e2) {
-				log.warn("Could not determine query size by counting query result", e2);
-				return UNKNOWN_SIZE;
+				log.error("Could not determine query size by counting query result", e2);
+				// we can't query the database, so we won't be able to yield
+				// instances
+				return 0;
```
florianesser: Aren't there cases where `query` is a valid statement that can be executed but `SELECT COUNT(*) FROM (" + query + ") tmp" can't?

stempler: Well yes, but this is when even the original query cannot be executed.
But I think what should be added in addition is trying the original query even if there is no count query. Otherwise in that case we may also end up with a case where isEmpty yields a wrong result.

PR #449, opened 2017-09-21
Resolve original Instance before doing isInserted check

florianesser:


ui/plugins/eu.esdihumboldt.hale.ui/src/eu/esdihumboldt/hale/ui/service/instance/internal/orient/OrientInstanceService.java
```diff
@@ -193,9 +194,17 @@ public InstanceCollection getInstances(DataSet dataset) {
 
 			@Override
 			public boolean match(Instance instance) {
-				if (instance instanceof OInstance) {
-					return ((OInstance) instance).isInserted();
+				// If instance is an InstanceDecorator, it can't be checked
+				// whether the instance was actually inserted.
+				Instance originalInstance = instance;
+				while (originalInstance instanceof InstanceDecorator) {
+					originalInstance = ((InstanceDecorator) originalInstance).getOriginalInstance();
```
stempler: Could there be also cases where a decorator decorates a decorator? If yes, is that already covered?

florianesser: Yes, that can (and does) happen. That's why `getOriginalInstance` is called in a `while` loop until the result is no longer an `InstanceDecorator`.

PR #462, opened 2017-10-18
Instance index for Merge and Join

florianesser:


florianesser:
retest this please

florianesser:
retest this please

common/plugins/eu.esdihumboldt.hale.common.instance.index/src/eu/esdihumboldt/hale/common/instance/index/InstanceIndexServiceImpl.java
```diff
@@ -0,0 +1,268 @@
...
+import eu.esdihumboldt.hale.common.instance.model.impl.DefaultInstanceCollection;
+
+/**
+ * Maintains instance indexes
+ * 
+ * @author Florian Esser
+ */
+public class InstanceIndexServiceImpl implements InstanceIndexService {
+
+	private final Map<QName, MultimapInstanceIndex> indexes = new HashMap<>();
```
stempler: Is there a specific reason why the concrete `InstanceIndex` implementation is used here and not the interface?

ui/plugins/eu.esdihumboldt.hale.ui.index/src/eu/esdihumboldt/hale/ui/index/internal/InstanceIndexUpdateServiceImpl.java
```diff
@@ -0,0 +1,137 @@
...
+	}
+
+	@Override
+	public void customFunctionsChanged() {
+		// TODO Action required?
+	}
+
+	@Override
+	public void alignmentChanged() {
+		// TODO Action required?
```
stempler: Clear index and reindex in that case?

common/plugins/eu.esdihumboldt.hale.common.instance.index/src/eu/esdihumboldt/hale/common/instance/index/InstanceIndexServiceImpl.java
florianesser: No specific reason, I refactored it so that now the interface `HaleInstanceIndex` is used. Always using `InstanceIndex` directly is a bit cumbersome because of the generics. Definitely could be refactored further in the future in case we want to use another index implementation.

florianesser:
> 1. Does it make sense to keep the old functions around or is it save to completely replace them.

My intention was to use the old handlers (`JoinHandler` and `PropertiesMergeHandler`) as fallbacks in cases where the index can't be used, e.g. if the instances of the source `InstanceCollection` don't implement `Identifiable`.

> 2. With the use of the index, how much effort would you see in supporting (a) using derived values (e.g. removing or adding a prefix for a certain property) for specific join conditions or (b) supporting more complex conditions than equality?

I think both could be done with reasonable effort, i.e. I don't think that these would require major changes in the index implementation. Currently, the `IndexJoinIterator` uses `InstanceIndex.getInstancesByValue()` to find instances with a specific value (equality check). This method could be extended to support more complex requests like the ones you mentioned.

Issue #478, opened 2017-12-08
Shapefiles: Prefill charset dialog with value from accompanying .cpg file


florianesser:
When importing a shapefile as a schema or instance source, check for an accompanying `.cpg` file and, in case it exists, use the character set specified in it to prefill the charset selection dialog.

florianesser:
[Feature build](https://builds.wetransform.to/job/hale/job/hale%7Epublish(master)/313/)

thorsten-reitz:
LGTM.

PR #479, opened 2017-12-08
Use value in .cpg to prefill charset dialog when importing shapefile

florianesser:
see #478

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ See the [change log guidelines](http://keepachangelog.com/) for information on h
 
 ## [Unreleased]
 
+### Added
+
+- When importing a shapefile resource, prefill character set dialog with encoding read from accompanying `.cpg` file
+
 ### Fixed
 
 - Fixed hale connect login on Welcome Page to work for user names and passwords w/ special characters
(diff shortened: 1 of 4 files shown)


thorsten-reitz:
The CPG is an optional, but official file for a shapefile, which is a simple ASCII file with the name of the encoding. QGIS for example also writes the CPG file. Much of our data actually contains these.

stempler:
Just wondering where to find such information. The [specification I found](https://www.esri.com/library/whitepapers/pdfs/shapefile.pdf) is probably to old and not complete.

thorsten-reitz:
This article mentioned the file as part of a shapfile: http://desktop.arcgis.com/en/arcmap/10.3/manage-data/shapefiles/shapefile-file-extensions.htm

stempler:
Seems to be more info [here](https://www.loc.gov/preservation/digital/formats/fdd/fdd000280.shtml#specs) as well. Would be interesting to know though if the value is always compatible to the encodings available in Java.
Also, it might make sense to make it a default when reading Shapefiles, not only if loaded in the UI. Maybe Geotools does that already ans we disable the behavior by specifying an explicit encoding.

florianesser:
> Also, it might make sense to make it a default when reading Shapefiles, not only if loaded in the UI. Maybe Geotools does that already and we disable the behavior by specifying an explicit encoding.

Good point, I'll look into that.

Issue #281, opened 2017-01-23
Improve behaviour on missing Bursa-Wolf parameters


thorsten-reitz:
During export of a GML FeatureCollection with a configured reprojection, Geotools throws an exception with the text "Missing Bursa-Wolf Parameters".  More Information is available here: http://docs.geotools.org/stable/userguide/faq.html

We've had this error reported by users at least 10 times by now. I therefore propose to catch this error and ask for complete CRS information, and to store it so that the PRJ info is ignored henceforth.

florianesser:
[Feature build](https://builds.wetransform.to/job/hale/job/hale%7Epublish(master)/316/)

thorsten-reitz:
I hit a problem with one of my Ukraine Shapefiles while testing that. Will provide the data.

florianesser:
@thorsten-reitz [The latest RC build](https://builds.wetransform.to/job/hale/job/hale%7Ebuild(staging)/8/) also contains a fix that should prevent the axis flip.

thorsten-reitz:
LGTM.

PR #480, opened 2017-12-11
Improve CRS lookup for shapefiles

florianesser:
This fixes a bug in `EPSGResolveCache` that prevented the CRS database lookup to work properly. This will usually prevent the `missing Bursa-Wolf parameters` exception (#281) in cases where the `.prj` file contains an incomplete WKT definition but Geotools can map it to an EPSG code nonetheless.

In cases where Geotools can't map the WKT definition to an EPSG code, the CRS dialog prompt will now be forced, so that the user can select an EPSG code, or, in cases where none exists, provide a WKT definition.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,12 @@ See the [change log guidelines](http://keepachangelog.com/) for information on h
 ### Added
 
 - When importing a shapefile resource, prefill character set dialog with encoding read from accompanying `.cpg` file
+- Enhanced CRS detection when parsing GML files
 
 ### Fixed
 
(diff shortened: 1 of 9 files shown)


florianesser:
retest this please

florianesser:
Handled the headless case so that the definition from the `.prj` will be used as is if it can't be resolved via Geotools. Furthermore, CRS detection in GML files is improved so that an encountered `srsName` attribute value is now used also for nested geometries. The CRS lookup dialog is now prefilled with the encountered value to facilitate the selection of the correct CRS.

PR #483, opened 2017-12-18
Test axis order of Geotools CRS for different prefixes

florianesser:


common/plugins/eu.esdihumboldt.hale.common.instance.test/src/eu/esdihumboldt/hale/common/instance/geometry/CRSDefinitionUtilTest.java
```diff
@@ -0,0 +1,68 @@
...
+					prefix.getPrefix() + wgs84Code));
+			CoordinateReferenceSystem wgs84latlon = CRS.decode(prefix.getPrefix() + wgs84Code);
+			assertEquals(AxisOrder.NORTH_EAST, CRS.getAxisOrder(wgs84latlon));
+
+			System.out.println(MessageFormat.format("Testing lon/lat \"{0}\"",
+					prefix.getPrefix() + wgs84Code));
+			CoordinateReferenceSystem wgs84lonlat = CRS.decode(prefix.getPrefix() + wgs84Code,
+					true);
+
+			// Longitude first currently works only with EPSG_CODE prefix
```
stempler: If that's the case [here](https://github.com/halestudio/hale/blob/0d2342f7c4d963e0f66027861e883f44ee10440e/common/plugins/eu.esdihumboldt.hale.common.instance/src/eu/esdihumboldt/hale/common/instance/geometry/impl/CodeDefinition.java#L110) we should make sure that this prefix is used.

Issue #455, opened 2017-10-05
Support CQL functions in filter expressions for condition contexts


florianesser:
Allow something like `strToLowerCase(VALUE) like '%m%'` as a filter expression.

florianesser:
Can be tested with [new RC](https://builds.wetransform.to/job/hale/job/hale%7Epublish(master)/326/)

thorsten-reitz:
LGTM

thorsten-reitz:
Was funded by Hansestadt Hamburg.

PR #491, opened 2018-01-16
Allow ECQL expressions in CQLFilterField

florianesser:
Allow extended CQL expressions in filters (#455)

### DIFF
--- a/doc/plugins/eu.esdihumboldt.hale.doc.user/html/reference/cql_filter.html
+++ b/doc/plugins/eu.esdihumboldt.hale.doc.user/html/reference/cql_filter.html
@@ -6,9 +6,12 @@
 </head>
 <body>
 	<h1>CQL Filter</h1>
-	<p>The Common Query Language (CQL) is used to define expressions
+	<p>
+		The Common Query Language (CQL) is used to define expressions
 		and filters in several parts of <i>hale studio</i>. It can be used to define
-		conditions on schema elements or to filter instances.</p>
+		conditions on schema elements or to filter instances. <i>hale studio</i> also 
+		supports the extended CQL (ECQL) syntax as defined by Geotools. 
+	</p>
 
 	<h2>Property references</h2>
 	<p>In a filter, you can refer to a property value by the property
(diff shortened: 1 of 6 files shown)


doc/plugins/eu.esdihumboldt.hale.doc.user/html/reference/cql_filter.html
```diff
@@ -58,6 +61,156 @@ <h2>Literals</h2>
...
+	<h2>Filter functions</h2>
+	<p>
+		For more complex filters you can use one or a combination of multiple functions. A selection of available functions is provided below. 
+		For the full list of functions, see GeoServer's <a href="http://docs.geoserver.org/latest/en/user/filter/function_reference.html">function reference</a>, 
+		but note that not all of the functions listed there are necessarily supported in a filter.
+	</p>
+	<p>
+		When defining a filter, replace all arguments with either a property name or a lexical expression 
+		(e.g. <tt>'John'</tt> or <tt>47</tt>). Also note, that all expressions must evaluate to true or false, so you may
+		have to add "<tt>= true</tt>" or "<tt>= false</tt>" after the function call, e.g. <tt>strMatches(name, 'Tom.*') = true<tt>.
```
stempler: How does `strMatches(name, 'Tom.*') = true` differ from `strMatches(name, 'Tom.*')`? Is that some kind of special behavior in the filters?

florianesser: The difference is that Geotools compiles `strMatches(name, 'Tom.*') = true` to `IsEqualToImpl` (which implements `org.opengis.filter.Filter` => ok) but `strMatches(name, 'Tom.*')` to `FilterFunction_strMatches` (which is a `FunctionExpression` and not a `Filter`).

I extended `FilterGeoECqlImpl` so that it now transparently tries to turn an `Expression` into a
`Filter` by wrapping the `Expression` in a `PropertyIsEqualTo` (same as adding `= true` to the expression).

PR #494, opened 2018-01-19
Set JVM maximum memory to 2 GiB for 64 bit builds

florianesser:


thorsten-reitz:
Should this still be merged in for the 3.3.2 release? If not, why?

florianesser:
Yes, I'll merge it for 3.3.2. The memory increase affects the 64-bit builds, so I think the Mac and Linux builds will get 2 GiB, too.

PR #527, opened 2018-02-27
add preset for AAA schema

stempler:
@JuergenWeichand We would like to add a preset for the AAA schema to hale (now that we know an address that is accessible on the web). Currently it would look like below - do you have suggestions for improving the name / description or even a 16x16 image that we can use?

hale-aaa


stempler:
At least right now the URL is not available:
```
$ ping repository.gdi-de.org
ping: unknown host repository.gdi-de.org
```
So it might make sense to add the schema to the cached resources in hale. If there are no changes expected to the schemas at the same URL then that would be viable.

PR #534, opened 2018-02-27
Fix fallback mechanism in index merge handler

florianesser:
Use fallback if instance is *not* `Identifiable`...


stempler:
The implementation of `is` also looks fishy to me:
https://github.com/halestudio/hale/blob/ad88b63f66b96cb42907ab5cd0ddec32950b3868/common/plugins/eu.esdihumboldt.hale.common.instance/src/eu/esdihumboldt/hale/common/instance/model/Identifiable.java#L46

If an object is not instance of Identifiable it is cast to it?

florianesser:
retest this please

PR #535, opened 2018-03-01
Cell Replace: Preserve annotations and pass existing function parameters to wizard

florianesser:
When replacing a `Cell`, copy all annotations of the old `Cell` to the new one. Additionally, pass all existing function parameters of the old `Cell` to the function wizard. All parameters from the old function that are also supported by the new function will the be pre-filled in the wizard.


florianesser:
@stempler It wasn't, so I added that the documentation and a couple of other properties get copied.

PR #539, opened 2018-03-06
Fix task-related notifications

florianesser:
Don't notify `TaskServiceListeners` when no tasks were actually added/removed.


ui/plugins/eu.esdihumboldt.hale.ui.views.mapping/src/eu/esdihumboldt/hale/ui/views/mapping/AlignmentView.java
```diff
@@ -713,7 +713,7 @@ private boolean isDisabledForCurrentType(Cell cell) {
 	private void refreshGraph() {
 
 		final Display display = PlatformUI.getWorkbench().getDisplay();
-		display.syncExec(new Runnable() {
+		display.asyncExec(new Runnable() {
```
florianesser: @stempler With this change the graph updates faster when there are multiple cells with tasks in the alignment. Do you think that multiple asynchronous updates that are started in quick succession could cause problems here?

stempler: Good question. I would not expect more problems than when `sync` is used.

PR #545, opened 2018-03-13
[Do not merge] Use Groovy 2.4

stempler:
Solved the compilation problems with the Alignment Merger tests for me. Let's see what the test run says on possible other implications. I think it is quite a risky change to merge into the current snapshot and I would prefer to do it on a new snapshot.

One other thing: The Groovy Eclipse compiler seems to support invoke dynamic now, so we could try if we can use that.


stempler:
Closed in favor of https://github.com/halestudio/hale/pull/548

PR #548, opened 2018-03-14
[Do not merge] Use invoke dynamic support

stempler:
extends on https://github.com/halestudio/hale/pull/545


stempler:
Would be interesting to see if we can gain some performance with this, both in hale itself and the transformation functions.

@florianesser Do you think it makes sense to merge this to the staging branch to be able to do some further testing before we decide on a merge to master?

florianesser:
@stempler I pushed the `groovy-2.4-indy` branch to `staging` and created a [build](https://builds.wetransform.to/job/hale/job/hale%7Ebuild(staging)/15/).

stempler:
My tests so far with that version showed no problems (in relation to the Groovy changes). I can't say anything about if the performance has been improved or not.
My suggestion would be to merge it after the 3.4 release so any downstream projects that use the 3.4 snapshot don't include these changes unintentionally.

Issue #559, opened 2018-03-20
Task icon on cell disappears when it is replaced


stempler:
Even though tasks on a cell are retained when a cell is replaced the icon in the Alignment view is not displayed after the replace. It is displayed again for instance when the project is saved and reloaded.

stempler:
Maybe a similar problem as in the Mapping view before https://github.com/halestudio/hale/commit/516a7adfd18cfbf4a6c7fe6e41c1bcd6cf89230c?

At least testing with the indy-Version I see that updates to the Alignment view often are not done when they should.

stempler:
> At least testing with the indy-Version I see that updates to the Alignment view often are not done when they should.

That behavior could also be related to https://github.com/halestudio/hale/commit/e3d4f1c1886b188e76245be9214c3365ebcbf7d6

florianesser:
[Build w/ fix](https://builds.wetransform.to/job/hale/job/hale%7Epublish(master)/362/)

PR #561, opened 2018-03-20
Retain task icon in CellFigure if cell is replaced

florianesser:
This addresses the problem described in #559 by implementing `equals()` and `hashCode()` for `DefaultCell`.

Equality is checked by comparing the `id` property. This assumes that no two cells in an alignment will have the same `id` under any circumstances. If this cannot be guaranteed, this change will possibly break `DefaultCell` equality checks.


stempler:
> If this cannot be guaranteed, this change will possibly break DefaultCell equality checks.

Well, the original cell and the cell it was replaced with would be two candidates :wink: 

Generally hard to guess if there are any consequences from the change, e.g. places where Cells are compared, or put in a Set.

florianesser:
I agree, I'll look for another approach.

Issue #544, opened 2018-03-13
disable scientific notation on number values for GML export


doemming:
Although scientific notation is valid according XML specification, often project specific needs require non scientific notation of numbers.
Possible solutions could be 
- adding an option to disable scientific notation on GML export.
- support this feature via groovy based functions, e.g. provide value class that extends from Integer (e.g. FormattedInteger) and overwrite toString() with custom behaviour. This way might allow to use different serialisations on specific attributes. The value call could be used explicit via groovy script to set a target value.

stempler:
@doemming @JohannaOtt Would you please test this for your use case with the [current build](https://builds.wetransform.to/job/hale/job/hale~publish(master)/368/) (once it is done)?

stempler:
Tested by @JohannaOtt

Issue #398, opened 2017-06-08
Number formatting in geometries must be done aware of neighbors


stempler:
The [number formatting when encoding GML geometries](https://github.com/halestudio/hale/blob/master/io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/writer/internal/geometry/writers/AbstractGeometryWriter.java#L238) is done individually for each ordinate.

But it is important that this is done in context of all coordinates, to be able to detect duplicates that are created (which should be eliminated). The formatting may lead to invalid geometries in both cases, this should be handled as well.

An example: A LinearRing in a Polygon contains a set of coordinates that only differ at the 5th position after the decimal point. If the ordinates are formatted to only show three digits after the decimal point there will be duplicate points in the geometry.

PR #566, opened 2018-03-22
Add option to format non-integer decimals in XML/GML output

florianesser:
This implements the feature requested in #544. 

The requested possibility to output decimals numbers in non-scientific notation can be achieved by supplying an adequate pattern to the new ~`gml.decimalFormat`~ `xml.decimalFormat` parameter, e.g. `0.############`.

The output will be rounded if the number digits in the fractional part of the pattern is lower than in the value to be formatted.


io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriter.java
```diff
@@ -149,11 +151,17 @@
...
+	 * The name of the parameter specifying how the output of geometry
 	 * coordinates should be formatted.
 	 */
 	public static final String PARAM_GEOMETRY_FORMAT = "geometry.write.decimalFormat";
 
+	/**
+	 * The name of the parameter specifying how the output of Double values
+	 * should be formatted.
+	 */
+	public static final String PARAM_DECIMAL_FORMAT = "gml.decimalFormat";
```
stempler: I would prefer `xml.` as prefix as this is not specific to GML but applies to any type of XML output.

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriter.java
```diff
@@ -1530,6 +1562,14 @@ private void writeElementValue(Object value, PropertyDefinition propDef)
 							propType.getConstraint(ElementType.class).getDefinition()));
 				}
 			}
+			else if (value instanceof Double || value instanceof Float
+					|| value instanceof BigDecimal) {
+				// Apply formatting only to decimal values, not integers
+				String representation = DecimalFormatUtil.applyFormatter((Number) value,
```
stempler: I think it would be better to continue using the XML specific conversion if no format is configured (instead of just using `String.valueOf(...)`).

io/plugins/eu.esdihumboldt.hale.io.gml.ui/src/eu/esdihumboldt/hale/io/gml/ui/GmlWriterSettingsPage.java
```diff
@@ -131,54 +149,164 @@ protected void createContent(Composite page) {
...
 		GridDataFactory.fillDefaults().grab(true, false).applyTo(writeFormat);
-		writeFormat.setText("Formatted Number");
+		writeFormat.setText("Output formatting");
 
-		checkNumberFormat = new Button(writeFormat, SWT.CHECK);
-		checkNumberFormat.setText("Use a formatted number output for geometry coordinates");
-		GridDataFactory.swtDefaults().span(2, 1).applyTo(checkNumberFormat);
-		checkNumberFormat.addSelectionListener(new SelectionAdapter() {
+		enableCoordinateFormat = new Button(writeFormat, SWT.CHECK);
+		enableCoordinateFormat.setText("Use a formatted number output for geometry coordinates");
```
stempler: While we are at it, should we add a warning that this may break the geometries by introducing duplicate points? The formatting will not "fix" a geometry nor warn about geometries that become invalid (e.g. a polygon that ha less than three distinct points).

florianesser: As a side note, the problem w/ duplicate points is also addressed in #398

Issue #452, opened 2017-09-27
Basic transformation summary and custom fail conditions


stempler:
The goal here is to be enable a user to specify under what conditions a transformation is successful, and allow him to easily evaluate if it was successful under these conditions or not.

The idea is to add support for a transformation summary, that collects information from the reports and possibly other sources to provide that in a way that can be easily processed. The suggestion would be to use a structure that is compatible to JSON, to enable easy use outside the scope of hale studio.

In the first implementation the summary should contain information that we can easily access, e.g. the information from reports, which actions were run, if there were errors, warnings, etc.
Later it could be extended with additional information that can be used to analyse the transformation.

For specifying the success/failure conditions the idea is that the user can provide a file that states these conditions and relates to the transformation summary.
When the transformation is executed via the command line interface, the exit code should represent the result according to the success conditions.

In the process of this issue it seems also sensible to extend the report API with the notion of a task identifier added to a report, to better be able to identify specific reports.

This is a (not necessarily complete) list of tasks:

- [x] Add task identifier to reports
- [x] Add transformation summary API (maybe best to keep it generic, as there is also potential to use that for instance for CLI commands other than the transformation)
- [x] Add report information to transformation summary
- [x] Add information about total number of instances created in the transformation to the summary
- [x] Define how success conditions should be provided
- [x] Add support for evaluating success conditions

stempler:
Here are early drafts of possible summary and success condition files structure (in this case YAML). They are intended to serve as inspiration, not as blueprint. Also, they only include information from reports and still assume that there are only report names available, not task identifiers.

Summary:
```yaml
Instance transformation:
  - completed: true
    errors: 0
    warnings: 0
Instance validation:
  - completed: true
    errors: 0
    warnings: 0
GML file export:
  - completed: true
    errors: 0
    warnings: 0
```

Conditions:
```yaml
Instance transformation:
  required: true
  failOnError: true
  failOnWarning: false
Instance validation:
  required: false
  failOnError: true
  failOnWarning: true
GML file export:
  required: true
  failOnError: true
  failOnWarning: false
```

thorsten-reitz:
The conditions make sense to me. As to the transformation summary, I assume it is meant to be the summary of a single transformation run (with one source file/one set of soruce files)?

For the multi-transformation summary, the user might want to be able to get more information (in drill-down fashion):

- Which dataset contained which issues?
- Which target schema element is affected by which issues (e.g. validation warnings)

It could look like this:

```
Summary
[...]

PR #569, opened 2018-03-28
Statistics and custom success conditions for transformation command

stempler:
This provides support for custom success conditons as required for https://github.com/halestudio/hale/issues/452 as a Groovy script.


app/plugins/eu.esdihumboldt.hale.app.transform/src/eu/esdihumboldt/hale/app/transform/ExecTransformation.java
```diff
@@ -457,19 +438,89 @@ private void transform() throws InterruptedException, ExecutionException {
...
+		CompilerConfiguration compilerConfiguration = new CompilerConfiguration();
+		compilerConfiguration.setScriptBaseClass(DelegatingScript.class.getName());
+
+		// Configure the GroovyShell and pass the compiler configuration.
+		GroovyShell shell = new GroovyShell(getClass().getClassLoader(), new Binding(),
+				compilerConfiguration);
+		DelegatingScript script;
+		try (InputStream in = new DefaultInputSupplier(context.getSuccessEvaluationScript())
+				.getInput();
+				InputStreamReader reader = new InputStreamReader(in, StandardCharsets.UTF_8)) {
```
florianesser: Should we make the charset of the evaluation script configurable? Since it's a user input it can't be guaranteed to be in UTF-8.

common/plugins/eu.esdihumboldt.hale.common.core/src/eu/esdihumboldt/hale/common/core/report/util/StatsMerge.groovy
```diff
@@ -0,0 +1,165 @@
...
+		}
+		else {
+			combineSingleValue(key, a, b)
+		}
+	}
+
+	private Object combineSingleValue(String key, Object a, Object b) {
+		if (a instanceof Number && b instanceof Number) {
+			// add numbers
+			((Number) a).longValue() + ((Number) b).longValue()
```
florianesser: `Double`s will be truncated, is that ok?

doc/plugins/eu.esdihumboldt.hale.doc.user/html/tasks/transform_cli.html
```diff
@@ -349,6 +351,26 @@ <h4>Other options</h4>
...
+			They contain aggregated information on the different tasks that were run as part of the transformation.
+		</dd>
+		<dt><code>-successEvaluation &lt;file-or-URI-to-script&gt;</code></dt>
+		<dd>
+			This option allows you to customize the behavior of the transformation command in respect to if a run is interpreted as successful or not.
+			The script you can specify here is a Groovy script that is evaluated against the transformation statistics.
+			You can use the <code>-statisticsOut</code> parameter to get an idea of the contents of the statistics.<br>
+			A success evaluation script could for example look like this:<br>
+			<pre><code>// the transformation must have been completed
+assert aggregated['eu.esdihumboldt.hale.transform'].report.completed == true
```
florianesser: In case we don't want to make the charset for the eval script configurable, we should add here that the script is expected to be in UTF-8.

app/plugins/eu.esdihumboldt.hale.app.transform/src/eu/esdihumboldt/hale/app/transform/ExecTransformation.java
```diff
@@ -457,19 +438,89 @@ private void transform() throws InterruptedException, ExecutionException {
...
+		StatsCollector statistics = reportHandler.getStatistics();
+
+		if (context.getStatisticsFile() != null) {
+			// write to statistics file
+			try (Writer writer = Files.newBufferedWriter(context.getStatisticsFile().toPath(),
+					StandardCharsets.UTF_8)) {
+				writer.write(JsonOutput.prettyPrint(statistics.saveToJson(true)));
+			} catch (IOException e) {
+				ExecUtil.error("Error writing statistics file: " + e.getMessage());
+				// XXX should this fail the command?
```
florianesser: I think it's ok not fail here b/c the evaluation script will fail anyway as soon as it tries an `assert` on one of the statistics. And if the stats file isn't used (e.g. in the edge cases of an empty evaluation script or one that doesn't use any of the statistics) there's no reason to fail the command.

stempler: There could be a call that just outputs the statistics, w/o using an evaluation script. But I would still leave it as it is right now.

common/plugins/eu.esdihumboldt.hale.common.core/src/eu/esdihumboldt/hale/common/core/report/util/StatsMerge.groovy
stempler: If we add statistics that make use of floating point numbers we need to add support for that.
Some forms of aggregation like calculating a median would not work here and would require instead to collect the sum and count and calculate the median on the aggregated facts.

app/plugins/eu.esdihumboldt.hale.app.transform/src/eu/esdihumboldt/hale/app/transform/ExecTransformation.java
stempler: I think it is OK if it is not configurable. UTF-8 is often compatible with the default charsets and usage of special characters is unlikely for the logic (though they may be in comments).

doc/plugins/eu.esdihumboldt.hale.doc.user/html/tasks/transform_cli.html
stempler: Added a note to the documentation.

Issue #524, opened 2018-02-26
Support AdV CRS codes


stempler:
Support (some) AdV CRS codes by translating them to EPSG codes.
A nice addition would also being able to read default CRS information from an "NBA Bestandsdatenauszug".

florianesser:
@stempler I did a quick spike [here](https://github.com/florianesser/hale/tree/feature/adv-crs-codes) implementing a `CRSAuthorityFactory` for the AdV codes. Would that be a viable route to follow in your opinion?

florianesser:
> Would we then also be able to output the srsName w/ the AdV codes?

Not trivially. The AdV codes are mapped internally to EPSG codes and the `CoordinateReferenceSystem` objects returned by the new factory only contain the EPSG code. Therefore, after the initial lookup, they behave as if the EPSG code had been read instead of the AdV code.

If we want to also support writing these codes, we would also have to make changes [here](https://github.com/halestudio/hale/blob/ce78a4da5c90b18b3ceb70ebfe0cb5033a3d3caf/common/plugins/eu.esdihumboldt.hale.common.instance/src/eu/esdihumboldt/hale/common/instance/io/impl/AbstractGeoInstanceWriter.java#L286) for example. I would suggest to handle this in an enhancement ticket where we could consider to generalize the CRS support towards supporting multiple authorities (like AdV and possibly ISO in the future).

stempler:
> Not trivially. The AdV codes are mapped internally to EPSG codes and the CoordinateReferenceSystem objects returned by the new factory only contain the EPSG code. Therefore, after the initial lookup, they behave as if the EPSG code had been read instead of the AdV code.

If the `CodeDefinition` uses the AdV code when writing the output it could well use that code. Let's see how it behaves when testing.

florianesser:
[Feature build](https://builds.wetransform.to/job/hale/job/hale%7Epublish(master)/373/)

florianesser:
> If the CodeDefinition uses the AdV code when writing the output it could well use that code. Let's see how it behaves when testing.

Works indeed, here's a sample output when I configured the GML writer to transform to `urn:adv:crs:etrs89_lat-lon`:
```
<gml:Polygon gml:id="_c408bd00-eecc-46d2-b48c-4b195f36a0a4" srsName="urn:adv:crs:etrs89_lat-lon" srsDimension="2">
```

stempler:
I successfully tested
[...]

PR #574, opened 2018-03-29
Implement CRS authority factory for AdV-style CRS definitions

florianesser:
This adds support for a subset of AdV-style CRS definitions which are mapped to their corresponding EPSG  codes (#524). Two formats for the AdV codes are supported: `ADV:<code>` and `urn:adv:crs:<code>`. 

The following mappings are available:

- `ADV:ETRS89_UTM32` => `EPSG:25832`
- `ADV:ETRS89_UTM33` => `EPSG:25833`
- `ADV:ETRS89_Lat-Lon` => `EPSG:4258`
- `ADV:DE_DHDN_3GK2` => `EPSG:31466`
- `ADV:DE_DHDN_3GK3` => `EPSG:31467`
- `ADV:DE_DHHN92_NH` => `EPSG:5783`


ui/plugins/eu.esdihumboldt.hale.ui.common/src/eu/esdihumboldt/hale/ui/common/crs/AdvCrsAuthorityFactory.java
```diff
@@ -0,0 +1,363 @@
...
+ * or (at your option) any later version.
+ * 
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this distribution. If not, see <http://www.gnu.org/licenses/>.
+ * 
+ * Contributors:
+ *     wetransform GmbH <http://www.wetransform.to>
+ */
+
+package eu.esdihumboldt.hale.ui.common.crs;
```
stempler: I don't think a UI bundle is the right place for these factories as we would like to use them in headless mode as well.

Probably requires a different approach to registering the factories. Maybe a fragment to the Geotools bundle that registers the factories via `META-INF/services`?

florianesser: Ah, didn't think of that, I'll look into it.

florianesser:
Good catch, thanks! Added the new bundle to `eu.esdihumboldt.hale.common.feature.core`.

Issue #607, opened 2018-04-27
Partitioning by feature type can fail when transforming external/project data


florianesser:
When using the "Transform project data" or "Transform external data" wizard or doing a CLI transformation, the process may fail without producing any output when the "Create separate output file for every feature type" option was selected.

thorsten-reitz:
LGTM, tested with build 396.

PR #608, opened 2018-04-27
No passthrough if instances are partitioned by type

florianesser:
See #607 and #609

### DIFF
--- a/io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriter.java
+++ b/io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriter.java
@@ -507,12 +507,12 @@ public class StreamGmlWriter extends AbstractGeoInstanceWriter
 
 	@Override
 	public boolean isPassthrough() {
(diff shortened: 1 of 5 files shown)


io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriter.java
```diff
@@ -511,6 +511,9 @@ public boolean isPassthrough() {
 			InstanceCollectionPartitioner partitioner = getPartitioner(this, SimpleLog.NO_LOG);
 			return !partitioner.usesReferences();
 		}
+		else if (isPartitionByFeatureTypeConfigured()) {
```
stempler: So `isThresholdConfigured()` and `isPartitionByFeatureTypeConfigured()` cannot be true at the same time?
Otherwise I would move the `isPartitionByFeatureTypeConfigured()` condition to be the first check.

florianesser: They can't be at the moment, but I'll move it anyways in case that changes in the future.

Issue #611, opened 2018-05-04
"Isolated" check box cannot be selected in App-Schema alignment export


florianesser:
In the App-Schema export wizard on the `Configure workspaces` page, the checkbox in the `Isolated` column cannot be activated.

florianesser:
The GTK3 issues look like to have been fixed w/ the Photon platform upgrade. Since this won't be part of the bugfix release, I'm moving this ticket back to the 3.5.0 milestone.

thorsten-reitz:
LGTM

Issue #53, opened 2016-07-12
Application is very slow on Linux with GTK ~3.18


stempler:
This is due to a bug in the underlying SWT implementation. The same issue can be observed with Eclipse Mars. In Eclipse Neon the problem seems to be fixed. An update to Eclipse Neon however is not a trivial task for hale, as some UI mechanisms retained from Eclipse 3 seem to be no longer supported.

Workaround is switching to GTK 2 for running hale. This can be done by adapting the `HALE.ini` file.
You need to add the following lines before the `-vmargs` line:

```
--launcher.GTK_Version
2
```

milovanderlinden:
This should also be done if you get I/O wizard errors. Thankfully I came across this tip!

thorsten-reitz:
@milovanderlinden @johanvdw Thanks for the feedback.

@stempler probably we have to check what a migration to a current Eclipse version would mean. I don't see this happening before hale studio 4.0.0 though.

johanvdw:
Thorsten,
I saw this error on  HALE version 3.3.1.

thorsten-reitz:
@stempler @florianesser shouldn't this be fixed with the changes we have made?

florianesser:
@thorsten-reitz Yes, this should no longer be an issue with the upgrade to the Photon platform. @stempler Can you confirm?

stempler:
It is no longer a problem on my machine and I think using the Photon based version also worked for a couple of other people.

thorsten-reitz:
OK, I'll close this then.

PR #612, opened 2018-05-04
Fix "Isolated" checkbox in App-Schema alignment export wizard

florianesser:
See #611

@ridethepenguin @nmco Could you also review this, please?


nmco:
Thank you @florianesser, the fix looks good to me, I will manually test this Monday with GeoServer.

nmco:
I did a build of this branch and manually tested it:

image

It seems that it is not possible to select \ click the check-box for the isolated workspaces:

image

Please note that I tested this on Linux. Could this be someway related with the OS ?

nmco:
So removing the three commits of this PR, make the check-box selectable ... at least in my environment:

image

thorsten-reitz:
So this is apparently related to the OS - without the fix, it works on Linux, with fix it works on windows. Probably this needs to be refactored so that different code is used in the different builds?

Please note that @florianesser is on vacation and that we discussed on Friday that Geosolutions would take over responsibility for app-schema /hale studio plugin related patches.

ridethepenguin:
@nmco I have manually tested a HALE build from this branch on a native 64-bit Linux box running Fedora 27:

image

...and a 32-bit virtual machine running Windows 7:

image

Works for me in both environments: the checkbox is centered and it is editable. So... not sure what is going on? :-)
I've noticed the revision number I get is not the same as yours: are we building the same code?

image

On a side note: on Linux, I had to force SWT to use GTK2 instead of GTK3, or else I would not be able to open virtually any dialog because of [this bug](https://bugs.eclipse.org/bugs/show_bug.cgi?id=486069); I don't think it's related, but I wanted to mention it for the sake of completeness.

nmco:
@ridethepenguin regarding the revision numbers, I tested this several days ago with a merge of master (like you did) which means my HAED was different for obvious reasons.

Long story short, the issue is still present for me, but forcing HALE to use GTK2 instead of GTk3 fixed the issue for me (export SWT_GTK3=0).

So summarizing:
1. master, works fine for GTK3 but fails on Windows
2. @florianesser fix, works fine for Windows but not for GTK3

I will have to think about the best solution for this.

stempler:
@nmco Could be a similar issue then as in https://github.com/halestudio/hale/issues/53 that the underlying Eclipse/SWT version is just too old for current GTK versions.

nmco:
Thank you for the reference @stempler, so I would say let's go ahead and merge this fix. The workaround GTK3 users are using for the other related issues will work for this one too.

Issue #467, opened 2017-11-13
XML Schema element "all" not supported


stempler:
Schema types that use the XML Schema element ["all"](https://www.w3.org/TR/xmlschema11-1/#element-all) are not supported by the XML schema reader.
The all element specifies that the child elements can appear in any order and that each child element can occur zero or one time.
In the hale schema model, there is generally a specific order to properties assumed for XML. So when writing the order would be fixed, when reading hale already can handle properties that appear out of order (though there is a warning message that should be suppressed in this case).

Simple example see https://discuss.wetransform.to/#/thread/3kgn/Loading-XML-schema-with-xs:all-element-fails-with-

stempler:
Can probably be done by treating it similar to sequences (maybe with other handling for the cardinality of child elements?). Special handling would be required when reading XML, to suppress the warning that an element appears out of order.

thorsten-reitz:
LGTM.

PR #624, opened 2018-06-11
[WIP] Support XSD types with <xs:all>

florianesser:
Add support for `<xs:all>` in addition to `<xs:choice>` and `<xs:sequence>` in the XML schema reader (see #467)

To do:

- [x]  Assert that each element of the `<all>` group can be added only once


io/plugins/eu.esdihumboldt.hale.io.xsd/src/eu/esdihumboldt/hale/io/xsd/reader/XmlSchemaReader.java
```diff
@@ -962,6 +965,49 @@ else if (object instanceof XmlSchemaParticle) {
 			}
 			// </sequence>
 		}
+		else if (particle instanceof XmlSchemaAll) {
+			// <all>
+			XmlSchemaAll all = (XmlSchemaAll) particle;
+
+			// create group only if necessary (all groups that appear exactly
+			// once will result in no group if not forced)
```
stempler: Actually the group should be eliminated automatically. Not sure where that functionality is hooked up though.

florianesser:
Closing this in favour of a different implementation that supports an annotation on type level.

Issue #614, opened 2018-05-11
Partition Output by spatial extent


thorsten-reitz:
As a user, I want to partition the exported transformed data by its spatial extent.

Suggested approaches:

* Dataset bounding box quadtree scheme
* WGS84 Tile level quadtree scheme
* Projected Coordinate Reference System Quadtree scheme

At least one of these schemes should be implemented. The user should be able to pick the subdivision  level. Filenames of generated files should contain the treekey in the selected quadtree grid.

As each object may intersect multiple tile BBoxes, but shall be include din only one, I suggest to use the centroid of the object to determine to which tile it will be written.

florianesser:
[Feature build](https://builds.wetransform.to/job/hale/job/hale%7Epublish(master)/406/)

thorsten-reitz:
@florianesser that link doesn't work - is any of the newer builds also fine?

florianesser:
@thorsten-reitz Yes, you can use any of the newer builds! We recently reduced the number of builds that Jenkins keeps, so it might happen more often in the future that the specific feature build has already been deleted. It's not a problem to use a later build, though.

thorsten-reitz:
LGTM

PR #634, opened 2018-06-21
Add support for partitioning GML output by spatial extent

florianesser:
Adds support for spatial partitioning as described in #614. 

Both a dataset bounding box scheme and a WGS84-based global boundary schema are supported: In the first case, the boundary of the quadtree used for partitioning is the bounding box of the exported data. In the second case, the centroid of the each geometry is transformed to WGS 84 and the boundary of the quadtree is the world.


io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriter.java
```diff
@@ -302,53 +354,233 @@ protected IOReport execute(ProgressIndicator progress, IOReporter reporter)
...
+				}
+
+				qtProgress.advance(1);
+				if (++i % 100 == 0) {
+					qtProgress.setCurrentTask(
+							MessageFormat.format("Adding geometries ({0} instances processed)", i));
 				}
 			}
+			qtProgress.end();
+			qtProgress.begin("Building quadtree", ProgressIndicator.UNKNOWN);
```
stempler: There does not seem to be a semantical difference in the task name before ("Build quadtree") and this one.

PR #655, opened 2018-07-10
remove invalid AdV CRS code mappings

stempler:


florianesser:
I'd like to check the other ones for plausibility as well. I'd suggest to remove all mappings where we find discrepancies to the definitions in the GeoInfoDok.

PR #663, opened 2018-07-24
Support XSD types with <xs:all>

florianesser:
Add support for <xs:all> in addition to <xs:choice> and <xs:sequence> in the XML schema reader (see #467). This PR replaces #624.

Changed the implementation to use a type-based constraint rather than a group-based constraint b/c in most practical cases the `<xs:all>` group will be removed and the constraint won't be available.

### DIFF
--- a/common/plugins/eu.esdihumboldt.hale.common.schema/src/eu/esdihumboldt/hale/common/schema/model/constraint/type/IgnoreOrderFlag.java
+++ b/common/plugins/eu.esdihumboldt.hale.common.schema/src/eu/esdihumboldt/hale/common/schema/model/constraint/type/IgnoreOrderFlag.java
@@ -1,0 +1,79 @@
+/*
+ * Copyright (c) 2018 wetransform GmbH
+ * 
+ * All rights reserved. This program and the accompanying materials are made
+ * available under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation, either version 3 of the License,
+ * or (at your option) any later version.
+ * 
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this distribution. If not, see <http://www.gnu.org/licenses/>.
+ * 
+ * Contributors:
+ *     wetransform GmbH <http://www.wetransform.to>
+ */
+
+package eu.esdihumboldt.hale.common.schema.model.constraint.type;
+
+import eu.esdihumboldt.hale.common.schema.model.Constraint;
+import eu.esdihumboldt.hale.common.schema.model.TypeConstraint;
+import eu.esdihumboldt.hale.common.schema.model.constraint.AbstractFlagConstraint;
+import net.jcip.annotations.Immutable;
+
+/**
+ * Specifies if the order of the immediate children of the type should be
+ * ignored, e.g. in case the <code>&lt;complexContent&gt;</code> is composed of
+ * <code>&lt;all&gt;</code> instead of <code>&lt;sequence&gt;</code>.
+ * 
+ * @author Florian Esser
+ */
+@Immutable
+@Constraint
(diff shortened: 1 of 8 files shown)


common/plugins/eu.esdihumboldt.hale.common.schema/src/eu/esdihumboldt/hale/common/schema/model/constraint/type/IgnoreOrderFlag.java
```diff
@@ -0,0 +1,80 @@
...
+		super(enabled);
+	}
+
+	/**
+	 * @see eu.esdihumboldt.hale.common.schema.model.TypeConstraint#isInheritable()
+	 */
+	@Override
+	public boolean isInheritable() {
+		// XXX Not inheritable b/c the primary use case at the time of
+		// implementation is to support <xs:all> groups in type definitions and
```
stempler: Interesting - do you have any reference why it would not be possible to extend such a type?

io/plugins/eu.esdihumboldt.hale.io.xsd/src/eu/esdihumboldt/hale/io/xsd/reader/XmlSchemaReader.java
```diff
@@ -963,6 +965,59 @@ else if (object instanceof XmlSchemaParticle) {
...
 		}
+		else if (particle instanceof XmlSchemaAll) {
+			// <all>
+			XmlSchemaAll all = (XmlSchemaAll) particle;
+
+			// Declaring group must be the type, <all> groups can't be
+			// children of other groups.
+			if (!(declaringGroup instanceof XmlTypeDefinition)) {
+				reporter.error(
+						"<xs:all> groups that are not immediate children of a type definition are not supported.");
```
stempler: Is this not supported here or is that not valid in XML Schema? The message and the comment seem to hint at different things.

common/plugins/eu.esdihumboldt.hale.common.schema/src/eu/esdihumboldt/hale/common/schema/model/constraint/type/IgnoreOrderFlag.java
florianesser: Seems my comment there was a bit premature. What I had in mind was that it is not possible to extend an "all"-type and add more elements to it (although even that seems to be [allowed to a certain degree in XSD 1.1](https://stackoverflow.com/questions/22918403/how-to-use-xsall-when-complextype-needs-to-be-extended-in-xml-schema)). What I didn't consider was extending the type and adding only attributes, which is certainly possible. So this needs reconsideration.

io/plugins/eu.esdihumboldt.hale.io.xsd/src/eu/esdihumboldt/hale/io/xsd/reader/XmlSchemaReader.java
florianesser: It's not valid in XML Schema as `<all>` can't be put inside of `<sequence>`s or `<choice>`s ([source](https://www.w3.org/TR/xmlschema-1/#declare-contentModel)). I'll update the error message to better reflect that.

thorsten-reitz:
@florianesser @stempler is this still something that is worth picking up again?

stempler:
> is this still something that is worth picking up again?

Well, if we have a use case where support for this is required.

thorsten-reitz:
Would it make sense to merge this as-is to see if it works with the JVF_DTM schema?

stempler (approved):

Issue #575, opened 2018-03-29
inline transformation seems to be very slow and cannot be used in CLI


JohannaOtt:
Using the inline transformation slows down the transformation process significantly. When the function cell is enabled, the transformation takes ~1.5 h for 7.000 objects.
When it is disabled, the transformation takes ~7 min for 7.000 objects.

The problem occurs in a AdV BU 3D-Alignment - if testing with it is necessary I can provide it.

JohannaOtt:
when trying to execute a transformation project containing the Inline Transformation-function via the CLI, an error occurs. As the project is supposed to be delivered to the customer via hale connect, the Inline Transformation-function should be available via CLI as well:
20.06.2018 09:20:19 ERROR e.e.c.ConceptualSchemaTransformer.transform(169) - Skipping property transformation: Executing property transformation failed. (642 times)
eu.esdihumboldt.hale.common.align.transformation.function.TransformationException: Transformation service not available for inline transformation
        at eu.esdihumboldt.cst.functions.core.inline.InlineTransformation.evaluate(InlineTransformation.java:135) ~[na:na]
        at eu.esdihumboldt.hale.common.align.transformation.function.impl.AbstractSingleTargetPropertyTransformation.evaluate(AbstractSingleTargetPropertyTransformation.java:57) ~[na:na]
        at eu.esdihumboldt.hale.common.align.transformation.function.impl.AbstractPropertyTransformation.execute(AbstractPropertyTransformation.java:102) ~[na:na]
        at eu.esdihumboldt.cst.internal.FunctionExecutor.executeTransformation(FunctionExecutor.java:224) ~[na:na]
        at eu.esdihumboldt.cst.internal.FunctionExecutor.processValid(FunctionExecutor.java:133) ~[na:na]
        at eu.esdihumboldt.hale.common.align.model.transformation.tree.visitor.CellNodeValidator.visit(CellNodeValidator.java:110) ~[na:na]
        at eu.esdihumboldt.hale.common.align.model.transformation.tree.impl.CellNodeImpl.accept(CellNodeImpl.java:103) ~[na:na]
        at eu.esdihumboldt.hale.common.align.model.transformation.tree.impl.TargetNodeImpl.accept(TargetNodeImpl.java:177) ~[na:na]
        at eu.esdihumboldt.hale.common.align.model.transformation.tree.impl.TargetNodeImpl.accept(TargetNodeImpl.java:173) ~[na:na]
[...]

PR #713, opened 2018-12-07
Inline transformations: Improve caching of Groovy scripts and fix exception when used in CLI

florianesser:
The method of storing compiled scripts in the `ExecutionContext` of either the `TransformationFunction` or the `Cell` does not work in the case of inline transformations. If a property is mapped by an inline
transformation, a separate transformer instance is instantiated for every property value. As the `ConceptualSchemaTransformer` itself is stateless, the cached scripts do not get passed on and are recompiled for every property value, causing a significant slowdown (as reported in #575).

This new simple thread-local cache avoids this problem as long as the CSTs for the inline property transformations are all executed within the same thread.

A possible drawback is the increased memory usage as, over the course of a transformation run, every Groovy script will be cached. If this turns out to be a problem, it can be mitigated by using a more sophisticated cache implementantion with e.g. an upper memory limit.

@stempler Do you see any problems with this approach, especially not storing the cached script in the execution context but (thread-)globally?

### DIFF
--- a/common/plugins/eu.esdihumboldt.hale.common.headless/src/eu/esdihumboldt/hale/common/headless/impl/ProjectTransformationEnvironment.java
+++ b/common/plugins/eu.esdihumboldt.hale.common.headless/src/eu/esdihumboldt/hale/common/headless/impl/ProjectTransformationEnvironment.java
@@ -34,7 +34,9 @@ import eu.esdihumboldt.hale.common.align.service.TransformationFunctionService;
 import eu.esdihumboldt.hale.common.align.service.impl.AlignmentFunctionService;
 import eu.esdihumboldt.hale.common.align.service.impl.AlignmentTransformationFunctionService;
 import eu.esdihumboldt.hale.common.align.transformation.service.TransformationSchemas;
+import eu.esdihumboldt.hale.common.align.transformation.service.TransformationService;
 import eu.esdihumboldt.hale.common.codelist.service.CodeListRegistry;
+import eu.esdihumboldt.hale.common.core.HalePlatform;
 import eu.esdihumboldt.hale.common.core.io.HaleIO;
 import eu.esdihumboldt.hale.common.core.io.IOAdvisor;
 import eu.esdihumboldt.hale.common.core.io.IOProviderConfigurationException;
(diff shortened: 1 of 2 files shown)


stempler:
@florianesser I think it might be potential problem.

My first question would be if you know if this method for getting the script is also used when running scripts while editing? That would probably quickly fill up the cache.
Also otherwise, when often editing scripts in the UI and testing the transformation, or when using the same hale studio instance to load many projects, that could lead to problems.

On the CLI or in other applications it is less of a problem, as how it is used now there is usually a separate process for a transformation. It could be though, if there are applications that run many transformations from different projects.

florianesser:
> My first question would be if you know if this method for getting the script is also used when running scripts while editing? That would probably quickly fill up the cache.

As far as I see it, the compiling in the editor does not use the `GroovyUtil.getScript()` method (which does the caching) but is done directly. 

For a `GroovyCreate` function e.g. the compiling is done here: https://github.com/halestudio/hale/blob/6c44e3f2b8207d8847050e48f9adccc10f26cfb5/ui/plugins/eu.esdihumboldt.hale.ui.functions.groovy/src/eu/esdihumboldt/hale/ui/functions/groovy/GroovyCreatePage.java#L118

> Also otherwise, when often editing scripts in the UI and testing the transformation, or when using the same hale studio instance to load many projects, that could lead to problems.

This would indeed be a problem if the transformations are executed in the GUI thread or in a thread that is reused (e.g. as part of a thread pool). If the thread used for executing the transformation is not reused, the thread-local script cache is discarded as soon as the thread terminates. I'll investigate how it's currently done.

florianesser:
@JohannaOtt, it'd be great if you could test the AdV Building 3D alignment (see #575) with [this build](https://builds.wetransform.to/job/hale/job/hale%7Ebuild(staging)/lastSuccessfulBuild/artifact/build/target/hale-studio-3.5.0.SNAPSHOT-win32.win32.x86_64.zip). 

A good test scenario would be opening multiple hale studio instances, doing multiple edits of the alignment's Groovy scripts (simple changes should be enough) and doing multiple transformation runs. Please monitor RAM usage and hale/system stability.

JohannaOtt:
grafik
grafik
grafik

I tested with three instances of the mentioned hale studio 3.5.0 SNAPSHOT, each using Xmx2g and each transforming 150 Building instances (with slightly different Groovy Scripts). The transformations take about 45 seconds.
Doing the same transformation with one instance of hale studio 3.4.1 (using Xmx4g), it takes about 5 minutes.
Do you want me to do any more testing e.g. with more instances of hale studio or more Building instances?

florianesser:
@stempler I think this is ready for merging, or do you think it requires more testing?

Issue #707, opened 2018-11-13
hsd does not load


JohannaOtt:
Version 3.5.0.SNAPSHOT
Build 444

When trying to load a hsd file created with hale studio 3.4.1 it fails with the following error. Please note that this also affects creating schemas from hsd files on haleconnect

groovy.lang.GroovyRuntimeException: Could not find matching constructor for: eu.esdihumboldt.hale.common.core.io.report.impl.IOMessageImpl(org.codehaus.groovy.runtime.GStringImpl)
	at groovy.lang.MetaClassImpl.invokeConstructor(MetaClassImpl.java:1742)
	at groovy.lang.MetaClassImpl.invokeConstructor(MetaClassImpl.java:1549)
	at org.codehaus.groovy.vmplugin.v7.IndyInterface.selectMethod(IndyInterface.java:236)
	at eu.esdihumboldt.hale.common.schema.persist.hsd.XmlToSchema$_populateDefinition_closure4.doCall(XmlToSchema.groovy:184)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.codehaus.groovy.reflection.CachedMethod.invoke(CachedMethod.java:98)
	at groovy.lang.MetaMethod.doMethodInvoke(MetaMethod.java:325)
	at org.codehaus.groovy.runtime.metaclass.ClosureMetaClass.invokeMethod(ClosureMetaClass.java:264)
	at groovy.lang.MetaClassImpl.invokeMethod(MetaClassImpl.java:1034)
	at groovy.lang.Closure.call(Closure.java:418)
	at groovy.lang.Closure.call(Closure.java:434)
	at org.codehaus.groovy.runtime.DefaultGroovyMethods.each(DefaultGroovyMethods.java:2125)
	at org.codehaus.groovy.runtime.DefaultGroovyMethods.each(DefaultGroovyMethods.java:2110)
	at org.codehaus.groovy.runtime.DefaultGroovyMethods.each(DefaultGroovyMethods.java:2151)
	at org.codehaus.groovy.runtime.dgm$163.doMethodInvoke(Unknown Source)
[...]

PR #717, opened 2018-12-18
fix(hsd): don't log error if constraint type can't be recreated

florianesser:
Sometimes a constraint in a hale schema definition cannot be recreated, e.g. because the plugin that defines the constraint is unavailable. To allow loading the HSD anyway, the message level is reduced to a warning.

https://github.com/halestudio/hale/issues/707

### DIFF
--- a/common/plugins/eu.esdihumboldt.hale.common.schema.persist/src/eu/esdihumboldt/hale/common/schema/persist/hsd/json/JsonToSchema.groovy
+++ b/common/plugins/eu.esdihumboldt.hale.common.schema.persist/src/eu/esdihumboldt/hale/common/schema/persist/hsd/json/JsonToSchema.groovy
@@ -76,6 +76,15 @@ public class JsonToSchema {
 		}
 	}
 
+	protected void warn(String message, Throwable t = null) {
+		if (reporter) {
+			reporter.warn(new IOMessageImpl(message, t))
+		}
+		else {
+			log.warn(message, t)
+		}
+	}
+
 	/**
 	 * Read schemas from the given reader.
 	 * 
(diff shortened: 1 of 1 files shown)


common/plugins/eu.esdihumboldt.hale.common.schema.persist/src/eu/esdihumboldt/hale/common/schema/persist/hsd/json/JsonToSchema.groovy
```diff
@@ -197,11 +206,11 @@ public class JsonToSchema {
 						Object constraint = desc.getFactory().restore(config, definition, typeProvider, classResolver)
 						definition.setConstraint(constraint)
 					} catch (Exception e) {
-						reporter.error(new IOMessageImpl("Failed to restore constraint of type $id", e))
+						error("Failed to restore constraint of type $id", e)
 					}
 				}
 				else {
-					reporter.error(new IOMessageImpl("Could not find factory for constraint with type $id"))
+					warn("Could not find factory for constraint with type $id")
```
stempler: Why would an error message prevent loading the HSD? Wouldn't that just show up as error in the report?

florianesser: Sorry, I'll try to clarify:

The problem in #707 (HSD not loading) was caused by using a non-existent constructor of `IOMessageImpl`. This was fixed by using the local `error()` method instead.

In a second step I changed the message level to warning to allow transformations to succeed even if the constraints could not be recreated. In the case of FGDB constraints this would prevent successful transformation of projects that use HSDs created from an FGDB.

PR #725, opened 2019-01-09
Release preparations for 3.5.0

florianesser:


doc/plugins/eu.esdihumboldt.hale.doc.user/html/new/3_5_0.xhtml
```diff
@@ -0,0 +1,95 @@
...
+				<td class="section" colspan="2" id="3_5_0_Features" rowspan="1"><h2>New features</h2></td>
+			</tr>
+
+			<tr id="deegree-config-export">
+				<td class="title">Deegree workspace configuration export</td>
+				<td class="content">
+					<p>
+						A bit like the Geoserver App-Schema and Xtraserver configuration, it is now also 
+						possible to generate deegree workspace configurations from hale studio alignments.
+						This feature is only available through the command line interface.
```
stempler: No, it is actually only available via the UI right now.

doc/plugins/eu.esdihumboldt.hale.doc.user/html/new/3_5_0.xhtml
```diff
@@ -0,0 +1,95 @@
...
+			<tr>
+				<td class="section" colspan="2" id="3_5_0_Features" rowspan="1"><h2>New features</h2></td>
+			</tr>
+
+			<tr id="deegree-config-export">
+				<td class="title">Deegree workspace configuration export</td>
+				<td class="content">
+					<p>
+						A bit like the Geoserver App-Schema and Xtraserver configuration, it is now also 
+						possible to generate deegree workspace configurations from hale studio alignments.
```
stempler: Generated is not a complete workspace, but only the the FeatureStore configuration. I think it makes sense to change that here so people do not expect that styles, layers, themes, service configuration etc. are also generated.

doc/plugins/eu.esdihumboldt.hale.doc.user/html/new/3_5_0.xhtml
florianesser: @thorsten-reitz Needs to be changed in the blog post as well

florianesser: I'd suggest to put `This feature is currently not available in the command line interface.` instead.

stempler: Sounds good :+1: 
I think if the dependency would be added to hale-cli it could probably be used via the `schema rewrite` command, but I did not test that.

florianesser: I pushed new commits with updates. I also adapted the corresponding changelog entry.

Issue #752, opened 2019-08-15
"Unknown owner type" error when opening projects from hale connect


JohannaOtt:
Version 4.0.0.SNAPSHOT
Build 557890456

I want to open a transformation project from hale connect using the wetransform-support account. When doing so, this pop up window with erroe messages appear as soon as I hit the "Select" button:
![grafik](https://user-images.githubusercontent.com/34163838/63077568-ffab4900-bf38-11e9-9c2b-7de137a9d780.png)
These are the error messages related to it 
![grafik](https://user-images.githubusercontent.com/34163838/63077692-76484680-bf39-11e9-9253-c54f22def2c4.png)
This is the stacktrace of one of them:
java.lang.IllegalStateException: Unknown owner type
	at eu.esdihumboldt.hale.io.haleconnect.HaleConnectProjectInfo.getOwner(HaleConnectProjectInfo.groovy:105)
	at eu.esdihumboldt.hale.io.haleconnect.ui.projects.ChooseHaleConnectProjectWizardPage$6.getText(ChooseHaleConnectProjectWizardPage.groovy:328)
	at org.eclipse.jface.viewers.ColumnLabelProvider.update(ColumnLabelProvider.java:34)
	at org.eclipse.jface.viewers.ViewerColumn.refresh(ViewerColumn.java:141)
	at org.eclipse.jface.viewers.AbstractTableViewer.doUpdateItem(AbstractTableViewer.java:391)
	at org.eclipse.jface.viewers.StructuredViewer$UpdateItemSafeRunnable.run(StructuredViewer.java:475)
	at org.eclipse.core.runtime.SafeRunner.run(SafeRunner.java:42)
	at org.eclipse.ui.internal.JFaceUtil.lambda$0(JFaceUtil.java:44)
	at org.eclipse.jface.util.SafeRunnable.run(SafeRunnable.java:173)
	at org.eclipse.jface.viewers.StructuredViewer.updateItem(StructuredViewer.java:2158)
	at org.eclipse.jface.viewers.AbstractTableViewer.internalRefreshAll(AbstractTableViewer.java:674)
	at org.eclipse.jface.viewers.AbstractTableViewer.internalRefresh(AbstractTableViewer.java:612)
	at org.eclipse.jface.viewers.AbstractTableViewer.internalRefresh(AbstractTableViewer.java:604)
	at org.eclipse.jface.viewers.AbstractTableViewer.lambda$0(AbstractTableViewer.java:566)
[...]

PR #770, opened 2019-11-19
fix(hc): handle cases where user/org information is unavailable

florianesser:
This adds handling for cases where user or organisation info is queried via the hale connect API, but the information is cannot be obtained (e.g. because the user/org was deleted in the meantime or because of missing authorization).

Previously, failure to retrieve user/org information could lead to errors in the UI.

https://github.com/halestudio/hale/issues/752


florianesser:
retest this please

PR #785, opened 2020-01-17
feat: support creating GeoPackage files and writing instances

stempler:
Basic support for writing instances to a new or existing GeoPackage
file.
If the writer encounters an existing table it will write to that table
(and not change its structure). If the writer does not find a table it
tries to create it. The writer supports feature and attribute tables.

In addition added support for attribute tables to GeoPackage schema reader and
instance reader (to complement the writer functionality).

Possible later improvements to this basic version:

- specific support for Date and DateTime data types
- support foreign key constraints
- automatically fill foreign keys from autogenerated primary keys
  (probably a prerequisite for most scenarios where foreign keys are
  used)
- support database transactions
- support geometry types that are not supported by the JTS library used
  in hale


stempler:
First example GeoPackage file produced: [new3193440899081415900.zip](https://github.com/halestudio/hale/files/4076919/new3193440899081415900.zip)

io/plugins/eu.esdihumboldt.hale.io.geopackage/src/eu/esdihumboldt/hale/io/geopackage/GeopackageInstanceWriter.java
```diff
@@ -0,0 +1,570 @@
...
+				// convenience for overwriting empty existing file
+				file.delete();
+			}
+			if (!file.exists()) {
+				GeoPackageManager.create(file);
+			}
+			// open existing file
+			geoPackage = GeoPackageManager.open(file, true);
+
+//			geoPackage.beginTransaction(); //XXX not sure how to use this -> results in a SQL exception later on (SQLITE_BUSY)
```
thorsten-reitz: There is a test for the transactions: https://github.com/ngageoint/geopackage-java/blob/master/src/test/java/mil/nga/geopackage/test/db/TransactionTest.java

io/plugins/eu.esdihumboldt.hale.io.geopackage/src/eu/esdihumboldt/hale/io/geopackage/GeopackageInstanceWriter.java
```diff
@@ -0,0 +1,570 @@
...
+				 * Set<TypeDefinition> sortedSet = getSortedSchemas(
+				 * getTargetSchema().getMappingRelevantTypes());
+				 * 
+				 * for (TypeDefinition td : sortedSet) {
+				 * writeInstances(geoPackage, instances.select(new
+				 * TypeFilter(td)), progress, reporter); }
+				 */
+			}
+
+//			geoPackage.commit();
```
thorsten-reitz: In the test cited above, they do not seem to use `commit()`

io/plugins/eu.esdihumboldt.hale.io.geopackage/src/eu/esdihumboldt/hale/io/geopackage/GeopackageInstanceWriter.java
```diff
@@ -0,0 +1,570 @@
...
+					FeatureDao features = geoPackage.getFeatureDao(tableName);
+					FeatureRow frow = features.newRow();
+
+					populateRow(frow, instance);
+
+					// set geometry
+					String geometryColumn = features.getGeometryColumnName();
+					// XXX instead of using value traverse (GeometryFinder) for
+					// geometry?
+					Object geom = new InstanceAccessor(instance).findChildren(geometryColumn)
```
thorsten-reitz: Would this be limited to certain geometry models right now?

io/plugins/eu.esdihumboldt.hale.io.geopackage/src/eu/esdihumboldt/hale/io/geopackage/GeopackageInstanceWriter.java
stempler: The key is probably disabling autocommit, I'll test that later

io/plugins/eu.esdihumboldt.hale.io.geopackage/src/eu/esdihumboldt/hale/io/geopackage/GeopackageInstanceWriter.java
stempler: Generally right now everything is limited to what the JTS WKB reader/writer can handle (as internally JTS geometries are used), which seems to be less than what the GeoPackage library can.

stempler: Generally right now everything is limited to what the JTS WKB reader/writer can handle (as internally JTS geometries are used), which seems to be less than what the GeoPackage library can.

stempler: Generally right now everything is limited to what the JTS WKB reader/writer can handle (as internally JTS geometries are used), which seems to be less than what the GeoPackage library can.

stempler: Generally right now everything is limited to what the JTS WKB reader/writer can handle (as internally JTS geometries are used), which seems to be less than what the GeoPackage library can.

io/plugins/eu.esdihumboldt.hale.io.geopackage/src/eu/esdihumboldt/hale/io/geopackage/GeopackageInstanceWriter.java
stempler: I played around a bit, but it seems it's not so easy. Could be that the problem is that I want to both create tables and insert data in the transaction (a different approach would require some refactoring). As writing w/o transaction works well, I would not see that as a priority right now.

Issue #732, opened 2019-02-15
rendering target geometries cause errors when using default stlying


JohannaOtt:
hale-studio-4.0.0.SNAPSHOT-Build 457

Only some of the target geometries in the example project attached to this ticket get displayed on specific zoom levels in the Map View and the loading ("clock") sign won't disappear.
[Geologie_GK_25 (Geologische Karte 1 25000).zip](https://github.com/halestudio/hale/files/2868116/Geologie_GK_25.Geologische.Karte.1.25000.zip)

While loading the map, errors occur in the error log - example stack trace:
`java.lang.NullPointerException
	at eu.esdihumboldt.hale.ui.views.styledmap.painter.StyledInstanceMarker.initStyle(StyledInstanceMarker.java:121)
	at eu.esdihumboldt.hale.ui.views.styledmap.painter.StyledInstanceMarker.applyFill(StyledInstanceMarker.java:383)
	at eu.esdihumboldt.hale.ui.views.styledmap.painter.StyledInstanceMarker.applyFill(StyledInstanceMarker.java:1)
	at eu.esdihumboldt.hale.ui.views.styledmap.painter.InstanceMarker.paintPolygon(InstanceMarker.java:415)
	at eu.esdihumboldt.hale.ui.views.styledmap.painter.InstanceMarker.paintGeometry(InstanceMarker.java:289)
	at eu.esdihumboldt.hale.ui.views.styledmap.painter.InstanceMarker.paintGeometry(InstanceMarker.java:195)
	at eu.esdihumboldt.hale.ui.views.styledmap.painter.InstanceMarker.doPaintMarker(InstanceMarker.java:146)
	at eu.esdihumboldt.hale.ui.views.styledmap.painter.InstanceMarker.doPaintMarker(InstanceMarker.java:1)
	at de.fhg.igd.mapviewer.marker.BoundingBoxMarker.paintMarker(BoundingBoxMarker.java:74)
	at de.fhg.igd.mapviewer.marker.BoundingBoxMarker.paintMarker(BoundingBoxMarker.java:1)
	at de.fhg.igd.mapviewer.marker.AbstractMarker.paint(AbstractMarker.java:45)
	at de.fhg.igd.mapviewer.waypoints.MarkerWaypointRenderer.paintWaypoint(MarkerWaypointRenderer.java:42)
	at de.fhg.igd.mapviewer.waypoints.MarkerWaypointRenderer.paintWaypoint(MarkerWaypointRenderer.java:1)
[...]

PR #811, opened 2020-05-07
fix : rendering of target geometries

Kate-Lyndegaard:
https://github.com/halestudio/hale/issues/732


florianesser:
Thanks @Kate-Lyndegaard for providing this "lost" bugfix from our workshop that we never actually integrated 😄

Issue #626, opened 2018-06-12
Bad performance for Merge with auto-detect enabled


stempler:
The auto-detect feature in the Merge configuration seems to lead to a very bad performance during the transformation. It seems this problem was introduced already in version 3.2.0 with this PR: https://github.com/halestudio/hale/pull/285.

This problem occurs if there is a significant amount of instances that are merged together within a Merge, because the comparisons done for the auto-detect is O(n²). An additional factor is if there are many attributes that are compared.

Example from a median sized data set (~100k instances, up to ~1000 instances merged together):
- Transformation w/o auto-detect: ~3 minutes
- Transformation w/ auto-detect: more than 3 hours

Right now the workaround is to explicitly configure properties in the Merge configuration and leave the auto-detect feature turned off.

thorsten-reitz:
Add a comment to make clear that auto-merge affects performance negatively.

PR #819, opened 2020-05-22
Add explanation to merge function

Kate-Lyndegaard:
I would definitely like to improve this first attempt at an update to the hale studio help:-)

I don't totally understand the dialog which appears after the "auto-detect function" in the merge. It is titled "Please select other equal properties to merge". For example:
I have 3 features with ID = Breuberg, and I set ID as merge key.
I set a second field, ORT as an additional property to merge in this dialog. 2 of the 3 features have ORT = Breuberg, and 1 feature has ORT = Sandbach. Only the value Breuberg appears when I map the property ORT to target. Does it chose a value to keep randomly? Or, how does it differ from the auto-detect function other than it is manual selection?

Any feedback or suggestions for examples, or screenshots is appreciated.


florianesser:
retest this please

JohannaOtt:
I would not expect any of the values in ORT to appear in the target data. My understanding of this functionality was, that (if the target attribute has cardinality max 1), the source values are only transformed to the target property in case they are all the same and are not transformed at all in case several values occur.
According to the test case you describe, this was a misunderstanding. @florianesser @stempler Could you please clarify what you expect this functionality to do?

stempler:
> Please select other equal properties to merge
I would expect that every property the user ticks here, the user is sure that the value is equal for all instances with the same keys. Thus alll but one value are ignored.

florianesser:
retest this please

common/plugins/eu.esdihumboldt.hale.common.align/help/merge/help.xhtml
```diff
@@ -0,0 +1,25 @@
...
+<html xmlns="http://www.w3.org/1999/xhtml">
+<head>
+<meta http-equiv="Content-Type" content="text/html; charset=ISO-8859-1" />
+<title>Merge</title>
+</head>
+<body>
+<div id ="help">
+	<h4>Example mapping:</h4>
+	<p>
+		The merge function is used to merge multiple source features within a single, source dataset. The merged source features create a single instance in the target dataset. In the merge configuration you must specify which instances should be merged together by specifying a merge key property. Instances that have the same value for this key property will be merged into one instance. For the other properties, the values are accumulated in the merged instance.<br>
```
stempler, line 11: I think it would be good to mention that:
- multiple key properties can be selected (and that in that case all instances that are combined have the same values for all)
- it is also possible to not select a key property, which means all instances of that type get merged into one (which should be avoided if there is a large number of source instances)

common/plugins/eu.esdihumboldt.hale.common.align/help/merge/help.xhtml
```diff
@@ -0,0 +1,25 @@
...
+	</br>
+		<img src="../../eu.esdihumboldt.hale.common.align/help/merge/mergeMapping.png" /><br>
+	</br>
+		The auto-detect properties function checks for properties where all instances to merge have the same value, and will add only that unique value to the merged instance.<br>
```
stempler, line 15: Using the auto-detect functionality is associated to performance problems right now (see https://github.com/halestudio/hale/issues/626)
Should we document that here?
I don't remember if there was any action taken on that ticket (e.g. adding a warning not to use it)

Kate-Lyndegaard:
@stempler OK, that's what I thought. I will add your additional suggestions to the documentation.

Kate-Lyndegaard:
I added @stempler's suggestions to the merge function. I also added an extended explanation to the classification function in this PR.

common/plugins/eu.esdihumboldt.hale.common.align/help/classification/help.xhtml
```diff
@@ -0,0 +1,32 @@
...
+<html xmlns="http://www.w3.org/1999/xhtml">
+<head>
+<meta http-equiv="Content-Type" content="text/html; charset=ISO-8859-1" />
+<title>Classification</title>
+</head>
+<body>
+<div id ="help">
+	<h4>Example mapping:</h4>
+	<p>
+		The classification function replaces values in the source data with new values. There are a variety of methods that can be used to classify source values. The Add source value button allows you to type values directly in the table. If the value entered is found in the source data, it will be classified to the value entered as target value. <br>
```
stempler, line 11: Maybe "The classification function maps values from the source data to new values."?
I would at least avoid the "in the source data", as that sounds like the source data is modified.

common/plugins/eu.esdihumboldt.hale.common.align/help/classification/help.xhtml
```diff
@@ -0,0 +1,32 @@
...
+	</br>
+		<img src="../../eu.esdihumboldt.hale.common.align/help/classification/sourceValue.png" /><br>
+	</br>
+		The Add source values button fills the table with source values and enumerations found in the source property.<br>
```
stempler, line 15: I think it would be good to mark button names (also menu item names, view names, etc.) so the reader can clearly see the distinction from the normal text. From what I have seen in other places usually the `<i>` tag was used for that in other places in the help.
For example:
```xml
The <i>Add source</i> values button fills ...
```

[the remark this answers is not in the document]
JohannaOtt: Ok. I just think it is a bit confusing as the button in the line above has the same name. Can we maybe rename that one?

Kate-Lyndegaard:
Sorry! Now I see what you are saying. I missed that. I will change it.

florianesser:
Please don't merge yet, we're doing a quick rebase

Issue #816, opened 2020-05-13
Update help 4.1.0


JohannaOtt:
Within this issue todos on the HS Documentation shall be documented.
This issue is valid for Release 4.1.0

**Intro:**
There are some parts of the help that are outdated - others need an update. These are the pages/things affected from my point of view and those that were already collected in a different ticket. If anyone finds other parts that need to be reviewed - please don't hesitate to add them.

**Documentation routine**

    Report as enumeration
    Report IN PROGRESS [IN PROGRESS (your acronym) ] to indicate when you start with the specific doc change 
    Report DONE [DONE (your acronym) ] to indicate when the specific doc change has been done and comitted
    For each release a new HS Doc Task issue will be created by @Claudia-WE

**List of required changes on HS Documentation**

**DONE (kl)** 1. General 
- Remove the 'Hale Plug-in for FME' part from the contents overview.
- Check screenshots and update where necessary (I think they are done on Linux - as the majority of users is working on Windows it make make sense to replace them with screenshots taken from a hale studio instance running on Windows.

**DONE (kl)** 2. http://help.halestudio.org/latest/topic/eu.esdihumboldt.hale.doc.user/html/introduction/introduction_dhp.html?cp=0_0_2
- Update information on data harmonisation panel - link leads to wetransform homepage, can be misleading

**DONE (kl)** 3. http://help.halestudio.org/latest/topic/eu.esdihumboldt.hale.doc.user/html/getting_started/help.html?cp=0_2_0
- "The online help misses some information that is provided on-the-fly in hale studio, e.g. the transformation function reference and the detailed reference for I/O providers" is not true any more. There is online help available for the detailed reference of I/O providers at http://help.halestudio.org/latest/topic/eu.esdihumboldt.hale.doc.user.ioproviders/overview/InstanceReader.html?cp=0_6_2_0_0 - please clarify with the developers as I might misunderstand something

**DONE (kl)** 4. http://help.halestudio.org/latest/topic/eu.esdihumboldt.hale.doc.user/html/getting_started/main_workflow.html?cp=0_2_1
- Link to tutorial video leads to wetransform start page

**DONE (kl)** 5. http://help.halestudio.org/latest/topic/eu.esdihumboldt.hale.doc.user/html/getting_started/install_plugins.html?cp=0_2_3
- add information on available hale pro plugins
- reword 'third-party' plugins as there are wetransform plugins as well

[...]

PR #825, opened 2020-06-19
Update doc

Kate-Lyndegaard:
This PR includes fixes for issues 6, 7, 8, 9, 10, 11, 14 in Update help #816


Kate-Lyndegaard:
@JohannaOtt Regarding issue 9- I like the current INSPIRE data model page because it is clear and addresses the most common use case for presets. I did fix the specification link.
I agree that we should add the other presets, but perhaps to a new section immediately following the INSPIRE data model page? Do you have a suggestion for a title? I can only find the City GML model and the AAA model as additional presets. Are you aware of any others?

JohannaOtt:
@Kate-Lyndegaard Ok sounds good!
Concerning the title, something like "Using presets of common schemas"?
No, I also only know about these two, but now that I think about it, I think it would make sense to include the XPlanung schemas as presets, too. @stempler @florianesser What do you think?

PR #826, opened 2020-06-19
fix: support codes w/o name in GML dictionaries

### LINKED ISSUE #824: Investigate why GML Dictionary code lists are not loaded properly
Loading a codelist like http://spec.tn-its.eu/codelists/RoadFeatureSourceCode.xml seems to suceed, but after assigning the codelist to a `href` property, the dropdown doesn't populate. Also, when inserting the values manually, the validator complains about these values.

The codelist above does not contain `name` elements, so I added them through a small transformation project, resulting in entries like this one:

```
<gml:dictionaryEntry>
    <gml:Definition gml:id="pedestrianCrossing">
      <gml:description>A specified part of a road where pedestrians have right of way to cross</gml:description>
      <gml:identifier codeSpace="http://spec.tn-its.eu/codelists/RoadFeatureTypeCode">pedestrianCrossing</gml:identifier>
      <gml:name codeSpace="http://spec.tn-its.eu/codelists/RoadFeatureTypeCode">pedestrianCrossing</gml:name>
    </gml:Definition>
</gml:dictionaryEntry>
```

However, even with this codelist, the behaviour doesn't change.

stempler:
If a code definition is found in a GML dictionary w/o a name, the
identifier is used as a fall-back name.

fixes #824

Also: fix: correctly detect code list type

Reduce the priority of the SKOS code list content type to low, as
otherwise it would always be detected as type for XML based code lists,
as it does not have any other conditions associated than being an XML
file.


stempler:
> Are all the prefs and build changes in here by intent?

Well, a new test bundle was added - there were no previously existing tests for that bundle.

thorsten-reitz:
@stempler quick question, why do we have two CI builds? Does one build and the other test?

stempler:
> quick question, why do we have two CI builds? Does one build and the other test?

I think @florianesser can answer that. AFAIK the Travis CI build should replace the Jenkins build. But I saw the Travis CI build failing a lot so my assumption was that it is not ready yet for the switch and thus both are kept right now.

florianesser:
@thorsten-reitz 
@stempler is correct. I kept the Jenkins build around to make sure that the Travis CI build behaves identically. However, the Travis CI build often fails in situations where the Jenkins build succeeds (i.e. not because of an error in the code).

I didn't yet investigate this systematically, but my impression is that Travis mainly has trouble running the tests and downloading artifacts from the internet, in particular from our Artifactory.

PR #828, opened 2020-06-26
Upgrade to Eclipse platform 2020-03 and GeoTools 21.0

florianesser:


thorsten-reitz:
280 changed files... that will be quite the review :).

florianesser:
@thorsten-reitz A lot of that are just the package change from `com.vividsolutions.jts` to `org.locationtech.jts`.

common/plugins/eu.esdihumboldt.hale.common.align.merge/META-INF/MANIFEST.MF
```diff
@@ -23,7 +23,7 @@ Import-Package: com.google.common.base;version="17.0.0",
  eu.esdihumboldt.hale.common.instance.model,
  eu.esdihumboldt.util,
  eu.esdihumboldt.util.definition,
- javax.annotation,
+ javax.annotation;version="[1.2.0,1.2.0]",
```
stempler, line 26: Are there now multiple versions flying around? Might the hard cap on the version cause problems later?

common/plugins/eu.esdihumboldt.hale.common.filter/src/eu/esdihumboldt/hale/common/filter/internal/EntityReplacementVisitor.java
```diff
@@ -78,8 +78,7 @@ private String toPropertyName(EntityDefinition entityDefinition) {
 					return c.getChild().getName().getLocalPart();
 					// TODO use version w/ namespace in ambiguous cases?
 				}).collect(Collectors.joining("."));
-		// put in quotes so it is possible to parse the resulting filterTerm
-		return '"' + name + '"';
+		return name;
```
stempler, line 81: So this is no longer necessary? Did it cause problems?

cst/plugins/eu.esdihumboldt.cst.functions.geometric/src/eu/esdihumboldt/cst/functions/geometric/NetworkExpansion.java
```diff
@@ -151,10 +151,7 @@ protected Object evaluate(String transformationIdentifier, TransformationEngine
 		GeometryProperty<Geometry> result = null;
 		if (old_geometry != null) {
 			Geometry new_geometry = null;
-			BufferParameters bufferParameters = new BufferParameters();
-			bufferParameters.setEndCapStyle(CAP_STYLE);
```
stempler, line 155: Is this the default in the new API?

common/plugins/eu.esdihumboldt.hale.common.align.merge/META-INF/MANIFEST.MF
florianesser: This is a remnant from my attempts to move to JDK 11 where I pulled in `javax:javaee-api:7.0.0` which also defined this package, but only partially. From the commit message: "This prevents inadvertently importing the package from recursive dependencies where the package cannot be easily excluded (e.g. `javax:javaee-api:7.0.0`)."

If the version of the `javax.annotation` package is updated in the future, the version pin needs to be updated, too. I didn't think that to be a problem as it should result in a compile error if 1.2.0 isn't available anymore.

Given that we abandoned the move to JDK 11 for now, I wouldn't mind reverting it now and revisiting it later.

common/plugins/eu.esdihumboldt.hale.common.filter/src/eu/esdihumboldt/hale/common/filter/internal/EntityReplacementVisitor.java
florianesser: No, it's no longer necessary as far as I can see. The filter parser in GeoTools was changed and now correctly parses the name without quotes. With the quotes, the parser now escapes them, e.g. `"aa" = 'test'` is now parsed as `""aa"" = 'test'`.

common/plugins/eu.esdihumboldt.hale.common.align.merge/META-INF/MANIFEST.MF
stempler: No need to revert in my view, was just interested in the reason.

cst/plugins/eu.esdihumboldt.cst.functions.geometric/src/eu/esdihumboldt/cst/functions/geometric/NetworkExpansion.java
florianesser: No, oversight on my part, thanks for the pointer! I'll push an update.

PR #831, opened 2020-06-26
refactor: remove Server product

florianesser:
@stempler Besides the `server/` tree I also removed some server-related snippets from the build scripts. Is it ok to remove them or do you envision some future purpose for them?


build/gradle/commitAndProductionStage.gradle
```diff
@@ -676,156 +660,6 @@ def packageClient(productOutputPath) {
...
-                include(name: '**/*.sh')
-                include(name: '**/*.so')
-                include(name: '**/jre/bin/*')
-				include(name: launcherName)
-				include(name: "${productName}.app/Contents/MacOS/${launcherName}")
-            }
-        }
-
-        // Docker image
-        // right now mainly intended for Infocenter product
```
stempler, line 760: Isn't the Infocenter product also a server build? Does it still work after removing all the build logic?
If it's not needed for the Infocenter, then it can be removed from my point of view.

florianesser: Thanks, I restored it. It is indeed needed for the Infocenter build.

PR #832, opened 2020-06-28
Add update site creation and artifact upload to Travis build

florianesser:
Adds the missing steps of the Jenkins hale build to the Travis build. Also adds missing dependency mappings for unpuzzle.


florianesser:
@stempler I've disabled the `deployArtifacts` task again because it [consistently times out](https://travis-ci.org/github/florianesser/hale/jobs/702981296) on Travis where there is a 50 minute time limit for jobs. For a complete move to Travis we'd need to find a way to either speed up or split up that task.

PR #843, opened 2020-07-01
What's new page for 4.0 and update of point of contact page

florianesser:
@Kate-Lyndegaard I assume you accidentally added the wrong screen shot on the What's New page, so I replacedit with one that shows the XPlanGML entries in the popup. Also fixed one minor problem with an unescaped ampersand, otherwise LGTM.


florianesser:
retest this please

Issue #316, opened 2017-02-10
Use Java 8 date/time types for bindings


stempler:
The use of the "old" Java date/time types causes problems, especially related to time zones, as these are implicitly set based on the current system.
It would be better to use the Java 8 date/time types that allow for explicit handling of timezones.

- [ ] Use Java 8 types in bindings for XML Schema
- [ ] Handle Java 8 types in XML simple types conversion
- [ ] Add general converters between Java 8 and old date/time types
- [ ] Use Java 8 types in bindings for all other kinds of schema, where possible/necessary
- [ ] Adapt functions that use date/time types

stempler:
A possible workaround for problems with the time zone is to set a specific timezone that should be assumed. That can for instance be done via `HALE.ini`, by adding an additional system property:

```
-Duser.timezone=UTC
```

stempler:
Note: In the Geopackage support Java 8 types were used and some converters already added.

PR #858, opened 2020-08-25
fix: don't append time zone information when writing xmlDate

florianesser:
Prevents that time zone information is added to the GML output when
writing an `xmlDate`. Before, a date-only target property would
get serialized as e.g.

    <xplan:genehmigungsDatum>2020-08-02+02:00</xplan:genehmigungsDatum>

with the actual time zone suffix `+02:00` depending on the system time zone.

With this fix, the above example gets serialized as

    <xplan:genehmigungsDatum>2020-08-02</xplan:genehmigungsDatum>

instead.

https://github.com/halestudio/hale/issues/316


stempler (approved):

JohannaOtt:
@florianesser Is there already a hale-cli SNAPSHOT available containing the changes of this PR?

florianesser:
@JohannaOtt Yes, [here](https://builds.wetransform.to/job/hale/job/hale-cli%7Epublish(devel)/)

PR #900, opened 2021-05-21
fixup! docs:Added function examples

florianesser:
Add `help/` folder created in 3b43b45c899e0d8f59fa673b483ae8da591b8b72 to binary build


JohannaOtt (approved):
I checked the example of Spatial Join where the explanation was not displayed in the help before 
grafik

PR #901, opened 2021-05-21
Prevent duplicate imports of the same schema

florianesser:
Patch to prevent duplicate import of the same schema when the schema locations differ only in the scheme part, e.g. `http://example.org/a.xsd` and `https://example.org/a.xsd`.

Also fix up the Travis CI build to also log into Docker when running the integration stage.

### DIFF
--- a/io/plugins/eu.esdihumboldt.hale.io.xsd/src/eu/esdihumboldt/hale/io/xsd/reader/XmlSchemaReader.java
+++ b/io/plugins/eu.esdihumboldt.hale.io.xsd/src/eu/esdihumboldt/hale/io/xsd/reader/XmlSchemaReader.java
@@ -21,6 +21,7 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.text.MessageFormat;
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


io/plugins/eu.esdihumboldt.hale.io.xsd/src/eu/esdihumboldt/hale/io/xsd/reader/XmlSchemaReader.java
```diff
@@ -689,15 +692,26 @@ else if (item instanceof XmlSchemaNotation) {
 				XmlSchemaExternal imp = (XmlSchemaExternal) externalItems.getItem(i);
 				XmlSchema importedSchema = imp.getSchema();
 				String location = importedSchema.getSourceURI();
-				if (!(imports.contains(location))) { // only add schemas that
+				String targetNamespace = importedSchema.getTargetNamespace();
+				if (!imports.containsKey(location)) { // only add schemas that
```
JohannaOtt, line 696: Is it an issue that they key is a String (see line 362) and the `location` it is now comparted to might be a URI?

florianesser: @JohannaOtt The variable `location` is also of type String (not URI), so there shouldn't be a problem, I think. Is that what you meant?

JohannaOtt: Sorry, misread the line. Yes, that is what I meant.

JohannaOtt (approved):

PR #902, opened 2021-06-18
fix: fix validity check of the hale»connect export dialog

florianesser:
Because permissions checks done as part of the login status update are executed
asynchronously, it is possible that the owner radio buttons are only activated
after the validity of the dialog is input is already determined. This leads to
the situation that a logged-in user can't create or update transformation
projects even if they have the correct permissions.

With this fix, the login status check and dialog validation are separated and
the dialog validation is retriggered once the asynchronous operation to
determine the user's permissions are completed.

### DIFF
--- a/io/plugins/eu.esdihumboldt.hale.io.haleconnect.ui/src/eu/esdihumboldt/hale/io/haleconnect/ui/projects/HaleConnectTarget.java
+++ b/io/plugins/eu.esdihumboldt.hale.io.haleconnect.ui/src/eu/esdihumboldt/hale/io/haleconnect/ui/projects/HaleConnectTarget.java
@@ -154,6 +154,7 @@ public class HaleConnectTarget extends AbstractTarget<HaleConnectProjectWriter> 
 						.createLoginDialog(Display.getCurrent().getActiveShell());
 				if (loginDialog != null && loginDialog.open() == Dialog.OK) {
 					HaleConnectLoginHandler.performLogin(loginDialog);
+					updateLoginStatus();
 					updateState();
 					prefillTargetProject();
 				}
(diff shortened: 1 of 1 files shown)


io/plugins/eu.esdihumboldt.hale.io.haleconnect.ui/src/eu/esdihumboldt/hale/io/haleconnect/ui/projects/HaleConnectTarget.java
```diff
@@ -392,11 +393,20 @@ private void prefillTargetProject() {
...
-				&& (newProject.getSelection() || targetProject != null));
+		boolean valid = haleConnect.isLoggedIn();
+		// Creating or updating transformation projects is only possible
+		// if the current user has sufficient permissions to create
+		// projects either in the user or in an organization context
+		valid &= ownerUser.getEnabled() || ownerOrg.getEnabled();
+		if (!newProject.getSelection()) {
+			// If the option to update an existing project is selected,
+			// a target project must be selected before proceeding.
+			valid &= targetProject != null;
```
stempler, line 406: Quick research shows that by using the `&` operator there is not support for short-circuiting so it maybe makes sense to switch to `= valid &&` instead?

stempler (approved):

io/plugins/eu.esdihumboldt.hale.io.haleconnect.ui/src/eu/esdihumboldt/hale/io/haleconnect/ui/projects/HaleConnectTarget.java
florianesser: The lack of short-circuit support definitely makes a difference in case there are function calls involved that have side effects. While I don't think this was the case here, I've changed it for clarity. Thanks for the feedback!

Issue #727, opened 2019-01-16
Proxy settings are not applied when contacting update sites


florianesser:
It looks like that in the software installation dialog hale studio tries to contact upadate sites directly even when a proxy is configured. In environments where proxy use is required this leads to hale not being able to contact update sites.

PR #903, opened 2021-06-28
fix(ui): use IProxyService to apply proxy configuration in the UI

florianesser:
Replaces the direct setting of proxy-related system properties by calls to the
proxy interface exposed by the Eclipse `IProxyService`.

This also allows the proxy settings to be applied to RCP components that are
used within hale»studio, like the software installation dialog to install
plugins.

https://github.com/halestudio/hale/issues/727
SVC-881


stempler (approved):
So I assume the proxy service does among other things set the system properties as well?

florianesser:
@stempler Exactly, this happens in the `updateXXXSystemProperties` methods of `ProxyType` which are ultimately called (e.g. [updateHttpSystemProperties](https://github.com/eclipse/eclipse.platform.team/blob/06c54b81431309ec789ee6c3b64c53d2194e1de6/bundles/org.eclipse.core.net/src/org/eclipse/core/internal/net/ProxyType.java#L408)).

PR #907, opened 2021-08-09
Fix: hale studio does not start for mac os

kapil-agnihotri:


stempler:
Can one of the admins verify this patch?

thorsten-reitz (approved):

stempler:
ok to test

build/gradle/commitAndProductionStage.gradle
```diff
@@ -390,6 +391,10 @@ def includeAdditionalFiles(artifacts, relsrcpath, productOutputPath, reldstpath,
 		    if (!Helper.buildForMac(project)) {
 		        ant.untar(src: artifactZipPath, dest: artifactInstallPath, compression: 'gzip')
 		    } else {
+				//It is certain that this code will run when creating build for mac.
+				//Therefore, the jre must be copied into the `Eclipse` folder of the bundled application.
+				artifactInstallPath = new File(productOutputPath.absolutePath ,"/Eclipse.app/Contents/Eclipse/${reldstpath}")
+				
 				// Use external tar command for OSX build since the jdk to bundle with HALE contains symbolic links
 				// This does only work on UNIX-based systems with tar installed!
 				def command = ['tar', '--preserve-permissions','-xzf', artifactZipPath.absolutePath]
```
stempler, line 400: Here using tr is not a problem on Mac?

build/gradle/commitAndProductionStage.gradle
```diff
@@ -826,6 +844,34 @@ def packageServer(productOutputPath) {
...
+
+	File file = new File(filepath)
+	def fileContent = file.getText();
+
+	def indexOfVmargs = fileContent.indexOf("-vmargs")
+	
+	def updatedContent = fileContent.substring(0, indexOfVmargs)
+	.concat("-vm")
+	.concat("\n")
+	.concat("../Eclipse/jre/jdk8u282-b08/Contents/Home/jre/bin/java")
```
stempler, line 867: Can't we get rid of the `jdk8u282-b08` here? This would mean that when the bundled JRE is changed this code needs to be changed as well. It would be better of only the reference to the JRE artifact would need to be changed,

build/gradle/commitAndProductionStage.gradle
```diff
@@ -826,6 +844,34 @@ def packageServer(productOutputPath) {
...
+ * then appends "-vm" arguments along with the JRE location bundled within the application.
+ * Finally, the changes are written back to the "HALE.ini" file.
+ * filepath HALE.ini file path
+ */
+def updateHaleIniFileWithJrePath(filepath) {
+
+	File file = new File(filepath)
+	def fileContent = file.getText();
+
+	def indexOfVmargs = fileContent.indexOf("-vmargs")
```
stempler, line 862: I think it would be a bit cleaner to read the lines, but as I also would not expect `-vmargs` to show up in a different context in the ini file I think it is OK like this.

stempler (changes requested):
Looks good to me but we should get rid of the Java version specific reference in the code (`jdk8u282-b08`)

build/gradle/commitAndProductionStage.gradle
kapil-agnihotri: No modification done in the logic of reading the file. However, made a little bit change when inserting strings to make the code a bit cleaner.

build/gradle/commitAndProductionStage.gradle
kapil-agnihotri: I have done the change to strip `jdk8u282-b08` during `tar.gz` extraction. As for all the Mac jdk bundles, root folder is the name of the jdk, so I'm stripping the root folder name and directly copying the `Contents` into `jre` folder, thereby getting rid of `jdk8u282-b08`.

build/gradle/commitAndProductionStage.gradle
kapil-agnihotri: Here using `tar` is no problem. In Mac `tar` gives problem only when used with `--transform` option.

kapil-agnihotri:
@stempler, I have fixed the review comments and the build failure issues. Please let me know your review comments.

stempler (approved):

PR #915, opened 2021-09-02
fixup!feat: support to export data using shapefile writer

kapil-agnihotri:
This commit provides a fix for exporting the Shapefiles when some
of the instances are null in the attributes which caused the
improper schema creation. The change is done to add all the
attributes even if their their data is null.
This commit also provides a fix to truncate long column names
up to 9 characters. If column names are longer than 10 characters
then the library will not export their corresponding instance data.

ING-2971


kapil-agnihotri:
added additional test case for missing instance for the attribute.

stempler (approved):
Please change the commit message to include a space after `fixup!`. Otherwise LGTM

kapil-agnihotri:
Hi @stempler,

I have updated the commit message with a space after `fixup!`, `fixup! feat: support to export data using shapefile writer`

PR #922, opened 2021-09-20
feat: enable selecting more than one file when loading schema or data

kapil-agnihotri:
This commit adds functionality to import multiple source
files for schema import and also enables to import several
 data files for the source data import.
It provides some restriction like all the selected files
should be of same format (e.g. all .shp) and does not
allow the selection of multiple csv and xml files for
schema import. However, the user can select multiple
files of same format for source data import.
This also adds auto detect functionality for importing
multiple shape files for multiple shape schemas.

### DIFF
--- a/io/plugins/eu.esdihumboldt.hale.io.jdbc.spatialite/src/eu/esdihumboldt/hale/io/jdbc/spatialite/reader/internal/SpatiaLiteInstanceReader.java
+++ b/io/plugins/eu.esdihumboldt.hale.io.jdbc.spatialite/src/eu/esdihumboldt/hale/io/jdbc/spatialite/reader/internal/SpatiaLiteInstanceReader.java
@@ -16,11 +16,15 @@
 package eu.esdihumboldt.hale.io.jdbc.spatialite.reader.internal;
 
 import java.io.File;
+import java.io.IOException;
 import java.io.InputStream;
 import java.net.URI;
 import java.util.Map;
 
+import eu.esdihumboldt.hale.common.core.io.IOProviderConfigurationException;
+import eu.esdihumboldt.hale.common.core.io.ProgressIndicator;
 import eu.esdihumboldt.hale.common.core.io.Value;
+import eu.esdihumboldt.hale.common.core.io.report.IOReport;
 import eu.esdihumboldt.hale.common.core.io.supplier.DefaultInputSupplier;
 import eu.esdihumboldt.hale.common.core.io.supplier.LocatableInputSupplier;
 import eu.esdihumboldt.hale.common.instance.io.util.InstanceReaderDecorator;
@@ -75,6 +79,19 @@ public class SpatiaLiteInstanceReader extends InstanceReaderDecorator<JDBCInstan
 	public void setSource(LocatableInputSupplier<? extends InputStream> source) {
 		this.source = source;
 		internalProvider.setSource(new SpatiaLiteJdbcIOSupplier(new File(source.getLocation())));
+	}
+
+	/**
+	 * @see eu.esdihumboldt.hale.common.core.io.ImportProvider#execute(eu.esdihumboldt.hale.common.core.io.ProgressIndicator,
+	 *      java.lang.String)
+	 */
+	@Override
+	public IOReport execute(ProgressIndicator progress, String resourceIdentifier)
+			throws IOProviderConfigurationException, IOException, UnsupportedOperationException {
+		// for now loading multiple instances from SpatiaLite DB is not
+		// supported.
+		throw new UnsupportedOperationException(
+				"The operation is not supported for SpatiaLite DB multiple files!");
 	}
 
 }
--- a/io/plugins/eu.esdihumboldt.hale.io.jdbc.spatialite/src/eu/esdihumboldt/hale/io/jdbc/spatialite/reader/internal/SpatiaLiteSchemaReader.java
+++ b/io/plugins/eu.esdihumboldt.hale.io.jdbc.spatialite/src/eu/esdihumboldt/hale/io/jdbc/spatialite/reader/internal/SpatiaLiteSchemaReader.java
@@ -16,11 +16,15 @@
 package eu.esdihumboldt.hale.io.jdbc.spatialite.reader.internal;
 
 import java.io.File;
+import java.io.IOException;
 import java.io.InputStream;
 import java.net.URI;
 import java.util.Map;
 
+import eu.esdihumboldt.hale.common.core.io.IOProviderConfigurationException;
+import eu.esdihumboldt.hale.common.core.io.ProgressIndicator;
 import eu.esdihumboldt.hale.common.core.io.Value;
+import eu.esdihumboldt.hale.common.core.io.report.IOReport;
 import eu.esdihumboldt.hale.common.core.io.supplier.DefaultInputSupplier;
 import eu.esdihumboldt.hale.common.core.io.supplier.LocatableInputSupplier;
 import eu.esdihumboldt.hale.common.schema.io.util.SchemaReaderDecorator;
@@ -75,5 +79,17 @@ public class SpatiaLiteSchemaReader extends SchemaReaderDecorator<JDBCSchemaRead
 	public void setSource(LocatableInputSupplier<? extends InputStream> source) {
 		this.source = source;
 		internalProvider.setSource(new SpatiaLiteJdbcIOSupplier(new File(source.getLocation())));
+	}
+
+	/**
+	 * @see eu.esdihumboldt.hale.common.core.io.IOProvider#execute(eu.esdihumboldt.hale.common.core.io.ProgressIndicator,
+	 *      java.lang.String)
+	 */
+	@Override
+	public IOReport execute(ProgressIndicator progress, String resourceIdentifier)
+			throws IOProviderConfigurationException, IOException, UnsupportedOperationException {
+		// for now loading multiple schemas from SpatiaLite DB is not supported.
+		throw new UnsupportedOperationException(
+				"The operation is not supported for SpatiaLite DB multiple files!");
 	}
 }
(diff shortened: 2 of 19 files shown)


ui/plugins/eu.esdihumboldt.hale.ui/src/eu/esdihumboldt/hale/ui/io/source/MultipleSchemaFilesSourceFileFieldEditor.java
```diff
@@ -0,0 +1,76 @@
...
+		Set<String> allSelectedFileExtensions = filepaths.stream()
+				.map(p -> p.substring(p.lastIndexOf("."))).collect(Collectors.toSet());
+		if (allSelectedFileExtensions.size() > 1) {
+			msg = "All files must be of the same format!";
+			return msg;
+		}
+
+		// check if selected files are of the extensions that doesn't allow
+		// multiple schema imports.
+		List<String> extensions = Arrays.asList(".csv", ".xml");
```
stempler, line 66: Ideally this would be based on information provided in a related extension, but this pragmatic approach is also fine.

common/plugins/eu.esdihumboldt.hale.common.core/src/eu/esdihumboldt/hale/common/core/io/IOProvider.java
```diff
@@ -62,6 +62,21 @@
...
+	 * @param progress the progress indicator, may be <code>null</code>
+	 * @param resourceIdentifier identifier of the provided resource. Null if a
+	 *            new resource id needs to be generated.
+	 * @return the execution report
+	 * 
+	 * @throws IOProviderConfigurationException if the I/O provider was not
+	 *             configured properly
+	 * @throws IOException if an I/O operation fails
+	 */
+	public IOReport execute(ProgressIndicator progress, String resourceIdentifier)
```
stempler, line 77: From the usage and the number of implementations that just return null in the implementation I think the `IOProvider` interface is not the right location to add this method.
Maybe instead in `ImportProvider`?

This seems to be a method that is only called in a special case. It would be good if the method documentation goes into a bit more detail on that.

common/plugins/eu.esdihumboldt.hale.common.core/src/eu/esdihumboldt/hale/common/core/io/impl/GZipEnabledImport.java
```diff
@@ -64,6 +65,15 @@ public URI getUsedLocation() {
 			return getLocation();
 		}
 
+		/**
+		 * @see eu.esdihumboldt.hale.common.core.io.supplier.LocatableInputSupplier#getUsedLocations()
+		 */
+		@Override
+		public List<URI> getUsedLocations() {
```
stempler, line 72: From the usage and the number of implementations that just return null in the implementation I think the `LocatableInputSupplier` interface is not the right location to add this method.
Maybe instead only in `FileIOSupplier`? Or even a new supplier class that is dedicated to be a source for multiple files?

ui/plugins/eu.esdihumboldt.hale.ui/plugin.xml
```diff
@@ -914,6 +914,24 @@
             name="URL"
             priority="-5">
       </source>
+      <source
+            class="eu.esdihumboldt.hale.ui.io.source.MultipleSchemaFilesSource"
+            description="Load a resource from a file"
+            icon="icons/file_obj.gif"
+            id="eu.esdihumboldt.hale.ui.io.source.multiplefiles"
+            name="MultipleFiles"
```
stempler, line 922: Is that where "From MultipleFiles" comes from in the UI? Maybe "multiple files" would be better then?

ui/plugins/eu.esdihumboldt.hale.ui/src/eu/esdihumboldt/hale/ui/io/IOWizard.java
```diff
@@ -517,133 +521,148 @@ public boolean performFinish() {
 		if (!applyConfiguration()) {
 			return false;
 		}
+		FileIOSupplier source = (FileIOSupplier) ((ImportProvider) getProvider()).getSource();
```
stempler, line 524: Is that really a safe cast?
What about non-file locations like HTTP/HTTPS or JDBC URIs?

ui/plugins/eu.esdihumboldt.hale.ui/src/eu/esdihumboldt/hale/ui/io/source/MultipleSchemaFilesSourceFileFieldEditor.java
kapil-agnihotri: > Ideally this would be based on information provided in a related extension

Do you mean this would be received in the code from some other source? What is related extension (for my curiosity)?
> but this pragmatic approach is also fine

I have left the changes as it is.

common/plugins/eu.esdihumboldt.hale.common.core/src/eu/esdihumboldt/hale/common/core/io/IOProvider.java
kapil-agnihotri: Method moved to `ImportProvider` and updated the java doc.

common/plugins/eu.esdihumboldt.hale.common.core/src/eu/esdihumboldt/hale/common/core/io/impl/GZipEnabledImport.java
kapil-agnihotri: Created a new class `FilesIOSupplier` or should I rename it to `MultipleFilesIOSupplier`?

ui/plugins/eu.esdihumboldt.hale.ui/plugin.xml
kapil-agnihotri: Yes that was the cause "From MultipleFiles", now I have updated the name.

ui/plugins/eu.esdihumboldt.hale.ui/src/eu/esdihumboldt/hale/ui/io/IOWizard.java
kapil-agnihotri: Code updated for the non-file locations.

io/plugins/eu.esdihumboldt.hale.io.jdbc.spatialite/src/eu/esdihumboldt/hale/io/jdbc/spatialite/reader/internal/SpatiaLiteInstanceReader.java
```diff
@@ -77,4 +81,14 @@ public void setSource(LocatableInputSupplier<? extends InputStream> source) {
...
 	}
 
+	/**
+	 * @see eu.esdihumboldt.hale.common.core.io.IOProvider#execute(eu.esdihumboldt.hale.common.core.io.ProgressIndicator,
+	 *      java.lang.String)
+	 */
+	@Override
+	public IOReport execute(ProgressIndicator progress, String resourceIdentifier)
+			throws IOProviderConfigurationException, IOException {
+		return null;
```
stempler, line 91: I would guess for importing multiple SpatiaLite files this would have to be implemented?

If we do not support it I would be for throwing an `UnsupportedOperationException` and adding a comment instead of just returning `null`.

io/plugins/eu.esdihumboldt.hale.io.jdbc.spatialite/src/eu/esdihumboldt/hale/io/jdbc/spatialite/reader/internal/SpatiaLiteSchemaReader.java
```diff
@@ -76,4 +80,14 @@ public void setSource(LocatableInputSupplier<? extends InputStream> source) {
...
 	}
+
+	/**
+	 * @see eu.esdihumboldt.hale.common.core.io.IOProvider#execute(eu.esdihumboldt.hale.common.core.io.ProgressIndicator,
+	 *      java.lang.String)
+	 */
+	@Override
+	public IOReport execute(ProgressIndicator progress, String resourceIdentifier)
+			throws IOProviderConfigurationException, IOException {
+		return null;
```
stempler, line 91: Same as above

common/plugins/eu.esdihumboldt.hale.common.core/src/eu/esdihumboldt/hale/common/core/io/impl/GZipEnabledImport.java
stempler: It's perfect, thank you!

io/plugins/eu.esdihumboldt.hale.io.jdbc.spatialite/src/eu/esdihumboldt/hale/io/jdbc/spatialite/reader/internal/SpatiaLiteInstanceReader.java
kapil-agnihotri: Code modified to throw `UnsupportedOperationException`.

io/plugins/eu.esdihumboldt.hale.io.jdbc.spatialite/src/eu/esdihumboldt/hale/io/jdbc/spatialite/reader/internal/SpatiaLiteSchemaReader.java
kapil-agnihotri: Code modified to throw `UnsupportedOperationException`.

PR #923, opened 2021-09-23
fixup! feat: enable selecting more than one file when loading schema

kapil-agnihotri:
Fix selection of multiple files for windows and enable export provider


stempler (approved):
LGTM, sorry for the long wait

PR #925, opened 2021-10-04
fix: import wizard relative path selection throws NPE

kapil-agnihotri:
Code provides fix to handle NPE when relative path
check box is selected during the schema/instance
import. Code is modified to fire an explicit state
changed event when relative path check box is
selected.

ING-2969


stempler (approved):
Looks good to me, but the commit message seems off - I could not find an original commit "import wizard relative path selection throws NPE", so maybe you meant to use `fix: ` ?

kapil-agnihotri:
> Looks good to me, but the commit message seems off - I could not find an original commit "import wizard relative path selection throws NPE", so maybe you meant to use `fix: ` ?

I have updated the commit message.

PR #940, opened 2021-11-12
fixup! feat: enable selecting more than one file when loading schema …

kapil-agnihotri:
…or data

Change to enable autoDetect checkbox by default when the user selects
multiple Shapefiles for data import. And by default un-check it when
only one file is selected.

ING-2969


kapil-agnihotri:
Hi @stempler, Please let me know your feedback.

stempler (approved):
LGTM

PR #948, opened 2021-12-16
feat: add ability to skip n-lines when reading csv and xls files

Flaminietta:
This commits add the ability of skipping n-lines
when reading instances from either csv and xls files.
In the UI the user can find now a new button
to select the no. of lines to be skipped.
If one wants to skip the header,
then 1 should be selected.

ING-3086


io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/reader/internal/CSVInstanceCollection.java
```diff
@@ -85,8 +85,8 @@ private void proceedToNext() {
 
 			// initialize reader if necessary
 			if (csvReader == null) {
-				boolean skipFirst = reader.getParameter(CommonSchemaConstants.PARAM_SKIP_FIRST_LINE)
-						.as(Boolean.class, false);
+				int skipN = reader.getParameter(CommonSchemaConstants.PARAM_SKIP_N_LINES)
```
stempler, line 88: I saw that although this is a different constant you use the same value `skip`. So this needs too be backwards compatible to to accepting boolean values (e.g. from cases where the old setting is passed via CLI or from projects saved with the old setting).

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/reader/internal/CSVSchemaReader.java
```diff
@@ -76,6 +76,8 @@ protected Schema loadFromSource(ProgressIndicator progress, IOReporter reporter)
 		try {
 			// initializes the first line of the table (names of the columns)
 			firstLine = reader.readNext();
+			// secondLine = reader.readNext();
+			System.out.println("Testing firstLine" + Arrays.toString(firstLine));
```
stempler, line 80: Debug remnants?

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/reader/internal/CSVSchemaReader.java
```diff
@@ -144,10 +147,11 @@ protected Schema loadFromSource(ProgressIndicator progress, IOReporter reporter)
 				property.setLocation(getSource().getLocation());
 			}
 
+			System.out.println("Testing properties" + Arrays.toString(properties));
```
stempler, line 150: Remnant from debugging?

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/reader/internal/CSVUtil.java
```diff
@@ -33,18 +34,33 @@
  */
 public class CSVUtil implements CSVConstants {
 
+//	/**
```
stempler, line 37: There shouldn't be any line comments in the Javadoc comment

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/reader/internal/CSVUtil.java
```diff
@@ -33,18 +34,33 @@
...
+		return readFirst(provider, 0);
+	}
+
 	/**
-	 * Reads only the first line of a given CSV file
-	 * 
-	 * @param provider provider to get the parameters from
-	 * @return a reader containing the first line of the CSV file
-	 * @throws IOException if an I/O operation fails
+	 * @param provider to get the parameters from
```
stempler, line 51: Maybe also add a description for this method?
For me the documentation on what is returned and on `skipLines` sounds contradictive. Are the first "x" lines skipped or retrieved?

io/plugins/eu.esdihumboldt.hale.io.xls.test/src/eu/esdihumboldt/hale/io/xls/test/XLSInstanceIOTest.java
```diff
@@ -92,7 +91,7 @@ public void test() {
 		// read the instances form the temporary XLS file
 		XLSInstanceReader reader = new XLSInstanceReader();
 		reader.setSourceSchema(schema);
-		reader.setParameter(CommonSchemaConstants.PARAM_SKIP_FIRST_LINE, Value.of(true));
+		reader.setParameter(CommonSchemaConstants.PARAM_SKIP_N_LINES, Value.of(1));
```
stempler, line 94: It would be great to not only have a test for the new functionality but also for the old one to test backwards compatibility.

io/plugins/eu.esdihumboldt.hale.io.xls/src/eu/esdihumboldt/hale/io/xls/reader/XLSInstanceReader.java
```diff
@@ -84,8 +84,7 @@ public boolean isCancelable() {
 	protected IOReport execute(ProgressIndicator progress, IOReporter reporter)
 			throws IOProviderConfigurationException, IOException {
 
-		boolean skipFirst = getParameter(CommonSchemaConstants.PARAM_SKIP_FIRST_LINE).as(
-				Boolean.class);
+		int skip_N_lines = getParameter(CommonSchemaConstants.PARAM_SKIP_N_LINES).as(Integer.class);
```
stempler, line 87: Same as mentioned previously. We need to take into account that the parameter is used in the old way and be backwards compatible.

stempler (changes requested):
Since you changed the functionality of the original parameter you would also need to update its specification in the extensions:

https://github.com/halestudio/hale/blob/e1b8f335b1092bf5631362cd8273fd8cd6c8b9ef/io/plugins/eu.esdihumboldt.hale.io.csv/plugin.xml#L55
https://github.com/halestudio/hale/blob/e1b8f335b1092bf5631362cd8273fd8cd6c8b9ef/io/plugins/eu.esdihumboldt.hale.io.xls/plugin.xml#L78

It is important for the parameter to stay backwards compatible, so places where the parameter was specified based on the old semantics do not break (like previously saved projects, transformation calls via command line, etc). An alternative approach could have been to introduce a new parameter with a different name, that takes precedence in case it is specified.

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/reader/internal/CSVUtil.java
Flaminietta: solved

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/reader/internal/CSVSchemaReader.java
Flaminietta: deleted

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/reader/internal/CSVSchemaReader.java
Flaminietta: deleted

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/reader/internal/CSVUtil.java
Flaminietta: Solved. My understanding is: the method `readFirst` reads/retrieves always just one line, either the first line after the header or the first after skipLines-lines of the file.

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/reader/internal/CSVInstanceCollection.java
Flaminietta: Solved.

io/plugins/eu.esdihumboldt.hale.io.csv.ui/src/eu/esdihumboldt/hale/io/csv/ui/TypeSelectionPage.java
```diff
@@ -106,16 +107,20 @@ public void selectionChanged(SelectionChangedEvent event) {
...
+					text.setSelection(skipNlines);
+					label1.getParent().layout();
 				}
 			}
 		});
 
-		button = new Button(page, SWT.CHECK);
-		button.setText("Skip first line");
-		button.setSelection(true);
+		// button = new Button(page, SWT.CHECK);
```
stempler, line 117: Maybe remove the commented component?

io/plugins/eu.esdihumboldt.hale.io.csv.ui/src/eu/esdihumboldt/hale/io/csv/ui/TypeSelectionPage.java
```diff
@@ -106,16 +107,20 @@ public void selectionChanged(SelectionChangedEvent event) {
...
+		// button.setText("Skip first line");
+		// button.setSelection(true);
+
+		label2 = new Label(page, SWT.NONE);
+		label2.setText("No. of lines to skip");
+		text = new Text(page, SWT.SINGLE | SWT.BORDER);
```
stempler, line 123: Would it maybe make sense to use a component that only allows numbers as input?
For instance `org.eclipse.swt.widgets.Spinner` - there are a couple of examples in the hale studio code.

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/reader/internal/CSVInstanceCollection.java
stempler: I guess the PR has not been updated yet then?

io/plugins/eu.esdihumboldt.hale.io.csv/plugin.xml
```diff
@@ -53,9 +53,9 @@
                ref="eu.esdihumboldt.hale.io.csv">
          </contentType>
          <providerParameter
-               description="States if the first line in the CSV file should be skipped (e.g. because it contains the header)"
-               label="Skip first line"
-               name="skip"
+               description="States how many lines should be skipped when reading instances from a CSV file (e.g. because of the header)"
+               label="Skip N lines"
+               name="skipN"
```
stempler, line 58: Regarding the behavior you chose now to replace the existing parameter and be backwards compatible on it. So the name needs to stay `skip`.

Maybe mention that boolean values are also supported for backwards compatibility and what their effect is?

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/reader/CommonSchemaConstants.java
```diff
@@ -25,7 +25,7 @@
 	/**
 	 * the parameter specifying the reader setting
 	 */
-	public static final String PARAM_SKIP_N_LINES = "skip";
+	public static final String PARAM_SKIP_N_LINES = "skipN";
```
stempler, line 28: Regarding the behavior you chose now to replace the existing parameter and be backwards compatible on it. So the name needs to stay `skip`.

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/reader/internal/CSVInstanceCollection.java
```diff
@@ -85,8 +85,21 @@ private void proceedToNext() {
 
 			// initialize reader if necessary
 			if (csvReader == null) {
-				int skipN = reader.getParameter(CommonSchemaConstants.PARAM_SKIP_N_LINES)
-						.as(Integer.class, 0);
+
+				int skipN = 0;
+				Boolean skipType = reader.getParameter(CommonSchemaConstants.PARAM_SKIP_N_LINES)
+						.as(boolean.class);
```
stempler, line 91: I think it would be better to use the wrapper class `Boolean` for the conversion, especially as it is expected here that `null` may be returned.

io/plugins/eu.esdihumboldt.hale.io.csv/plugin.xml
```diff
@@ -53,9 +53,9 @@
...
                optional="true">
             <parameterBinding
                   class="java.lang.Boolean">
```
stempler, line 61: The binding should then probably be Integer?

Further down the `valueDescriptor` is also not updated yet.

io/plugins/eu.esdihumboldt.hale.io.xls.test/src/eu/esdihumboldt/hale/io/xls/test/reader/XLSReaderTest.java
```diff
@@ -314,6 +338,11 @@ private Schema readXLSSchema(String sourceLocation, int sheetIndex, String typeN
 		return schemaReader.getSchema();
 	}
 
+	private InstanceCollection readXLSInstances(String sourceLocation, int sheetIndex,
+			String typeName, boolean skipN, Schema sourceSchema) throws Exception {
+		return readXLSInstances(sourceLocation, sheetIndex, typeName, skipN ? 1 : 0, sourceSchema);
```
stempler, line 343: This does not really check the backwards compatibility as parameter is passed to the reader as integer.

io/plugins/eu.esdihumboldt.hale.io.xls/plugin.xml
```diff
@@ -76,8 +76,8 @@
                ref="eu.esdihumboldt.hale.io.xls.xlsx">
          </contentType>
          <providerParameter
-               description="States if the first line in the CSV file should be skipped (e.g. because it contains the header)"
-               label="Skip first line"
+               description="States how many lines should be skipped when reading instances from a XLS file (e.g. because of the header)"
+               label="Skip N lines"
```
stempler, line 80: The name is correct here :+1: 

Over than that same things as for CSV apply:
- mention backwards compatible behavior?
- parameterBinding/valueDescriptor further below were not updated

io/plugins/eu.esdihumboldt.hale.io.xls/src/eu/esdihumboldt/hale/io/xls/reader/XLSInstanceReader.java
```diff
@@ -107,27 +105,74 @@ protected IOReport execute(ProgressIndicator progress, IOReporter reporter)
 		propAr = type.getChildren().toArray(new PropertyDefinition[type.getChildren().size()]);
 		Collection<List<String>> rows = analyser.getRows();
 
-		// skip if first row is a header
-		if (skip_N_lines == 0) {
-			// otherwise first line is also an instance
+		int skip_N_lines = 0;
+		Boolean skipType = getParameter(CommonSchemaConstants.PARAM_SKIP_N_LINES).as(boolean.class);
+
+		if (skipType == null) {
```
stempler, line 111: As in this section you can already handle all different cases, I think it would be cleaner to handle it similar to in the CSV reader, to translate the boolean values into numbers (0/1), and not treat them as different cases again further below.

io/plugins/eu.esdihumboldt.hale.io.csv.test/src/eu/esdihumboldt/hale/io/csv/CSVInstanceReaderTest.java
```diff
@@ -79,10 +79,19 @@ public void testReadSimple() throws Exception {
 		}
 
 		// read Instances ###
-		InstanceCollection instances = readCSVInstances("/data/test1.csv", typeName, true, schema,
+		InstanceCollection instances = readCSVInstances("/data/test1.csv", typeName, 1, schema,
 				null, null, null);
+		System.out.println("Test: no. of instances" + collectionSize(instances));
```
kapil-agnihotri, line 84: Debug remnants?

io/plugins/eu.esdihumboldt.hale.io.csv.test/src/eu/esdihumboldt/hale/io/csv/CSVInstanceReaderTest.java
```diff
@@ -104,6 +113,32 @@ public void testReadSimple() throws Exception {
...
+	public void testBackwardCompatibilityRead() throws Exception {
+		String typeName = "location";
+		// read Schema ###
+		Schema schema = readCSVSchema("/data/test1.csv", typeName,
+				"java.lang.String,java.lang.String,java.lang.String,java.lang.String",
+				"Name,Xcoord,Ycoord,id", null, null, null);
+
+		// test backward compatibility with skip first line as boolean
+		int numberOfLines = 3; // no. of lines = header + 2 instances
+		InstanceCollection instances_no_skip = readCSVInstances("/data/test1.csv", typeName, false,
```
kapil-agnihotri, line 133: It would be great if you could follow the java naming standards with came case `instancesNoSkip` for `instances_no_skip`. This will keep the standards same across the application.

Does `no` in the variable name represents `NO` or the `number`? Maybe then also update the name of the variable accordingly.

io/plugins/eu.esdihumboldt.hale.io.csv.test/src/eu/esdihumboldt/hale/io/csv/CSVInstanceReaderTest.java
```diff
@@ -104,6 +113,32 @@ public void testReadSimple() throws Exception {
...
+				schema, null, null, null);
+		assertEquals(numberOfLines, collectionSize(instances_no_skip));
+
+		InstanceCollection instances_skip_first = readCSVInstances("/data/test1.csv", typeName,
```
kapil-agnihotri, line 137: java naming standards with camel case.

io/plugins/eu.esdihumboldt.hale.io.csv.ui/src/eu/esdihumboldt/hale/io/csv/ui/TypeSelectionPage.java
```diff
@@ -90,8 +91,8 @@ protected void createContent(Composite page) {
 		GridData layoutData = new GridData();
 		layoutData.widthHint = 200;
 
-		label = new Label(page, SWT.NONE);
-		label.setText("Choose your Type:");
+		label1 = new Label(page, SWT.NONE);
```
kapil-agnihotri, line 94: It would be a bit more readable if you could rename `label1` and `label2` to something appropriate based on what those labels stand for.

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/reader/internal/CSVInstanceCollection.java
```diff
@@ -85,8 +85,21 @@ private void proceedToNext() {
...
+
+				if (skipType == null) {
+					skipN = reader.getParameter(CommonSchemaConstants.PARAM_SKIP_N_LINES)
+							.as(Integer.class, 0);
+				}
+				else if (skipType == true) {
```
kapil-agnihotri, line 97: Although, `if (skipType == true)` is completely acceptable, you can also shorten it to `if (skipType)`

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/reader/internal/CSVInstanceCollection.java
```diff
@@ -85,8 +85,21 @@ private void proceedToNext() {
...
+					skipN = 1;
+				}
+				else if (skipType == false) {
```
kapil-agnihotri, line 100: Here you don't need `else if` just `else` is enough as the boolean parameter skipType can have only 3 values:
- null
- true
- false
so for the third condition you don't have to explicitly do a check for false.

io/plugins/eu.esdihumboldt.hale.io.xls.test/src/eu/esdihumboldt/hale/io/xls/test/reader/XLSReaderTest.java
```diff
@@ -83,16 +83,24 @@ public void testReadSimple() throws Exception {
...
+		// read Instances - not header ###
+		InstanceCollection instances = readXLSInstances("/data/simpleOneSheet.xls", 0, typeName, 1,
+				schema);
 		assertTrue(instances.hasSize());
 		assertEquals(numberOfInstances, instances.size());
 
+		// read and skip N instances ###
+		int numberOfLines = 4; // no. of lines = header + 3 instances
+		for (int i = 0; i <= numberOfLines; i++) {
+			InstanceCollection remaining_instances = readXLSInstances("/data/simpleOneSheet.xls", 0,
```
kapil-agnihotri, line 99: Java naming standards for `remaining_instances`

io/plugins/eu.esdihumboldt.hale.io.xls.test/src/eu/esdihumboldt/hale/io/xls/test/reader/XLSReaderTest.java
```diff
@@ -114,6 +122,30 @@ public void testReadSimple() throws Exception {
...
+	 */
+	@Test
+	public void testBackwardCompatibilityRead() throws Exception {
+		// read Schema ###
+		Schema schema = readXLSSchema("/data/simpleOneSheet.xls", 0, typeName,
+				"java.lang.String,java.lang.String,java.lang.String");
+		// test backward compatibility with skip first line as boolean
+		int numberOfLines = 4; // no. of lines = header + 3 instances
+
+		InstanceCollection instances_no_skip = readXLSInstances("/data/simpleOneSheet.xls", 0,
```
kapil-agnihotri, line 140: Java naming standards for `instances_no_skip`

io/plugins/eu.esdihumboldt.hale.io.xls.test/src/eu/esdihumboldt/hale/io/xls/test/reader/XLSReaderTest.java
```diff
@@ -114,6 +122,30 @@ public void testReadSimple() throws Exception {
...
+				typeName, false, schema);
+		InstanceCollection instances_skip_first = readXLSInstances("/data/simpleOneSheet.xls", 0,
```
kapil-agnihotri, line 142: Java naming standards for `instances_skip_first`

io/plugins/eu.esdihumboldt.hale.io.xls/src/eu/esdihumboldt/hale/io/xls/reader/XLSInstanceReader.java
```diff
@@ -108,23 +105,74 @@ protected IOReport execute(ProgressIndicator progress, IOReporter reporter)
 		propAr = type.getChildren().toArray(new PropertyDefinition[type.getChildren().size()]);
 		Collection<List<String>> rows = analyser.getRows();
 
-		// skip if first row is a header
-		if (!skipFirst) {
-			// otherwise first line is also an instance
+		int skip_N_lines = 0;
```
kapil-agnihotri, line 108: Java naming standards using camelCase without `_`

io/plugins/eu.esdihumboldt.hale.io.xls/src/eu/esdihumboldt/hale/io/xls/reader/XLSInstanceReader.java
```diff
@@ -108,23 +105,74 @@ protected IOReport execute(ProgressIndicator progress, IOReporter reporter)
...
+			// iterate over all rows to create the instances
+			readRows(reporter, rows);
 		}
 
-		// iterate over all rows to create the instances
+		reporter.setSuccess(true);
+		return reporter;
+	}
+
+	/**
```
kapil-agnihotri, line 163: Missing java doc for the method.

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/reader/internal/CSVConfiguration.java
```diff
@@ -36,14 +36,14 @@
 
 	private final char escape;
 
-	private final boolean skip_first_line;
+	private final int skip_N_lines;
```
kapil-agnihotri, line 39: Java naming standard with `camelCase`. Underscore is only allowed for constants and in-between `ALL_CAPITAL_LETTERS`.

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/reader/internal/CSVConfiguration.java
```diff
@@ -54,11 +54,11 @@ public CSVConfiguration() {
 	 * @param esc the escape sign
 	 * @param skip the boolean to skip the first line or not
 	 */
-	public CSVConfiguration(char sep, char qu, char esc, boolean skip) {
+	public CSVConfiguration(char sep, char qu, char esc, int skip_N_lines) {
```
kapil-agnihotri, line 57: Variable name as per java naming standard with camelCase.

kapil-agnihotri (changes requested):

io/plugins/eu.esdihumboldt.hale.io.xls.test/src/eu/esdihumboldt/hale/io/xls/test/reader/XLSReaderTest.java
Flaminietta: This is tricky for me. I spent a lot of time being able to pass through those tests and I finally came up with this solution. In this way, I can pass to the method readXLSInstances (see lines 140 and 142) a boolean value for skipN that then is converted here into an integer. Does it not check backward compatibility in this way? Why not? 

Can you maybe suggest to me a possible solution (pseudo code)?

Flaminietta: I couldn't let your suggestion to use `Value skipN `instead of `boolean` in the definition of the method work fine.

io/plugins/eu.esdihumboldt.hale.io.csv.ui/src/eu/esdihumboldt/hale/io/csv/ui/TypeSelectionPage.java
Flaminietta: This is not something I have introduced.

io/plugins/eu.esdihumboldt.hale.io.xls/src/eu/esdihumboldt/hale/io/xls/reader/XLSInstanceReader.java
```diff
@@ -108,23 +105,71 @@ protected IOReport execute(ProgressIndicator progress, IOReporter reporter)
...
+
+		// do not skip any lines
+		if (skipNlines == 0) {
+
+			// also the header is retrieved
 			createInstanceCollection(analyser.getHeader(), reporter);
+
 			line++;
+
+			readRows(reporter, rows);
```
kapil-agnihotri, line 134: Looks like after the modification of converting `boolean` to `int`, `readRows(reporter, rows);` is used only in one place. I believe that the new method can be removed and its code can be pasted here.

kapil-agnihotri (changes requested):

io/plugins/eu.esdihumboldt.hale.io.csv.ui/src/eu/esdihumboldt/hale/io/csv/ui/TypeSelectionPage.java
```diff
@@ -48,8 +49,10 @@
 public class TypeSelectionPage extends InstanceReaderConfigurationPage implements CSVConstants {
 
 	private TypeDefinitionSelector sel;
-	private Button button;
-	private Label label;
+	private Text text;
```
stempler, line 52: Is the text component still used?

io/plugins/eu.esdihumboldt.hale.io.csv.ui/src/eu/esdihumboldt/hale/io/csv/ui/TypeSelectionPage.java
```diff
@@ -48,8 +49,10 @@
...
+	private Spinner spinner;
```
stempler, line 53: Maybe name the variable in a way that it is more clear which setting this component is associated to?

io/plugins/eu.esdihumboldt.hale.io.xls/plugin.xml
```diff
@@ -76,16 +76,19 @@
...
-               description="States if the first line in the CSV file should be skipped (e.g. because it contains the header)"
-               label="Skip first line"
+               description="States how many lines should be skipped when reading instances 
+               from a XLS file (e.g. because of the header). Also boolean values are 
+               supported for backwards compatibility. When a boolean is used, true is translated 
+               into 1 (i.e. skip the first line) and false into 0 (i.e. do not skip any lines)"
+               label="Skip N lines"
                name="skip"
-               optional="true">
+               optional="0">
```
stempler, line 85: I think this should be `true`. Didn't you use the extension editor in Eclipse? I think that would not have let you choose a wrong value.

io/plugins/eu.esdihumboldt.hale.io.xls/plugin.xml
```diff
@@ -76,16 +76,19 @@
...
             <parameterBinding
-                  class="java.lang.Boolean">
+                  class="java.lang.Integer">
             </parameterBinding>
             <valueDescriptor
                   default="false"
```
stempler, line 90: I think this relates to the actual default value, so that would be `0`

io/plugins/eu.esdihumboldt.hale.io.xls/plugin.xml
Flaminietta: I don't know what the extension editor is, can you please show me? However, in my logic, `skip` is now an integer and that is why I would use default `0`.

stempler (approved):

io/plugins/eu.esdihumboldt.hale.io.csv.test/src/eu/esdihumboldt/hale/io/csv/CSVInstanceReaderTest.java
```diff
@@ -246,11 +280,47 @@ private Schema readCSVSchema(String sourceLocation, String typeName, String para
...
+	}
+
 	private InstanceCollection readCSVInstances(String sourceLocation, String typeName,
 			boolean skipFirst, Schema sourceSchema, String seperator, String quote, String escape)
-					throws Exception {
-		return readCSVInstances(sourceLocation, typeName, skipFirst, sourceSchema, seperator, quote,
-				escape, null);
+			throws Exception {
+
+		InstanceReader instanceReader = new CSVInstanceReader();
```
kapil-agnihotri, line 308: I believe that there were no additional changes in this method and the entire method body can be replaced by the original code from the master to avoid code duplication.

```
return readCSVInstances(sourceLocation, typeName, skipFirst, sourceSchema, seperator, quote,
 				escape, null);
```

io/plugins/eu.esdihumboldt.hale.io.csv.test/src/eu/esdihumboldt/hale/io/csv/CSVInstanceReaderTest.java
```diff
@@ -246,11 +280,47 @@ private Schema readCSVSchema(String sourceLocation, String typeName, String para
 		return schemaReader.getSchema();
 	}
 
+	private InstanceCollection readCSVInstances(String sourceLocation, String typeName, int skipN,
+			Schema sourceSchema, String seperator, String quote, String escape) throws Exception {
+
```
kapil-agnihotri, line 285: I believe that the method body can be replaced with 
```
return readCSVInstances(sourceLocation, typeName, skipN, sourceSchema, seperator, quote,
 				escape, null);
```
This will call your method:
```
private InstanceCollection readCSVInstances(String sourceLocation, String typeName, int skipN,
 			Schema sourceSchema, String seperator, String quote, String escape, String decimal)
```
with decimal parameter as `null`and will avoid code duplication.

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/reader/internal/CSVConfiguration.java
```diff
@@ -85,8 +85,8 @@ public char getEscape() {
 	/**
 	 * @return the skip_first_line
```
kapil-agnihotri, line 86: As the parameter `skip_first_line` no longer exists, in the comment
 `@return skipNlines` instead of `@return the skip_first_line`?

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/reader/internal/CSVUtil.java
kapil-agnihotri: Here @stempler meant also to add a description to the method documentation. Like as below:
```
/**
* This method reads/retrieves always just one line, either the first line after the header or the first line after skipping N number of lines from the given file.
* 
* @param provider to get the parameters from
* @param skipLines how many lines have to be skipped
* @return a reader containing the first line after (no. of lines -
*         skipLines) of the CSV file
* IOException
*/
```

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/reader/internal/CSVUtil.java
```diff
@@ -36,15 +37,31 @@
...
+	 * @param skipLines how many lines have to be skipped
+	 * @return a reader containing the first line after (no. of lines -
+	 *         skipLines) of the CSV file
+	 * @throws IOException
+	 */
+	public static CSVReader readFirst(ImportProvider provider, int skipLines) throws IOException {
 		Reader streamReader = new BufferedReader(
 				new InputStreamReader(provider.getSource().getInput(), provider.getCharset()));
+//		CSVReader reader = new CSVReader(streamReader, getSep(provider), getQuote(provider),
```
kapil-agnihotri, line 60: Dead code, should be removed.

io/plugins/eu.esdihumboldt.hale.io.xls.test/src/eu/esdihumboldt/hale/io/xls/test/XLSInstanceIOTest.java
kapil-agnihotri: I hope that here the functionality with `boolean` parameter `True`was tested as I don't see any test case testing the backwards compatibility with a boolean parameter.

kapil-agnihotri (changes requested):

io/plugins/eu.esdihumboldt.hale.io.csv.test/src/eu/esdihumboldt/hale/io/csv/CSVInstanceReaderTest.java
Flaminietta: @kapil-agnihotri I don't fully understand your point. If I look at the code as it is now, it makes sense to me that all overloaded methods `readCSVInstances `return an object that is yet an instance, so I am fine with `return instanceReader.getInstances()`. What would it mean if instead I use `return readCSVInstances(sourceLocation, typeName, skipFirst, sourceSchema, seperator, quote, escape, null);`? Can you please explain it to me?

io/plugins/eu.esdihumboldt.hale.io.csv.test/src/eu/esdihumboldt/hale/io/csv/CSVInstanceReaderTest.java
Flaminietta: I understand that I can substitute the two kinds of return. But I don't understand where I can avoid code duplication.

io/plugins/eu.esdihumboldt.hale.io.csv.test/src/eu/esdihumboldt/hale/io/csv/CSVInstanceReaderTest.java
kapil-agnihotri: > What would it mean if instead I use `return readCSVInstances(sourceLocation, typeName, skipFirst, sourceSchema, seperator, quote, escape, null);`


It will
- call another overloaded method with a `null` parameter for the last argument which will again return `instanceReader.getInstances()`
- avoid code duplication
- code logic in one place easier to modify and maintain, in case of any changes in the future, then the logic needs to be modified only in one place.

io/plugins/eu.esdihumboldt.hale.io.xls.test/src/eu/esdihumboldt/hale/io/xls/test/XLSInstanceIOTest.java
Flaminietta: Added.

kapil-agnihotri (approved):

PR #949, opened 2022-01-28
feat: add option to specify leniency when using DateExtraction

Flaminietta:
This commit adds the option of specifying whether or not date/time parsing is to be lenient when using the function DateExtraction. It also modifies the description of the parameter "Date format" so as to be less prone to mistakes (e.g., quotes should not be used and the format refers to the one in the source data).

ING-3214

### DIFF
--- a/cst/plugins/eu.esdihumboldt.cst.functions.string/plugin.xml
+++ b/cst/plugins/eu.esdihumboldt.cst.functions.string/plugin.xml
@@ -50,7 +50,7 @@
             </property>
          </targetProperties>
          <functionParameter
-               description="For example: yyyy-MM-dd HH:mm:ss;&#xA; see http://docs.oracle.com/javase/6/docs/api/java/text/SimpleDateFormat.html for more details"
+               description="For example: yyyy-MM-dd HH:mm:ss. &#xA; See http://docs.oracle.com/javase/6/docs/api/java/text/SimpleDateFormat.html for more details."
                label="Source date format"
                maxOccurrence="1"
                minOccurrence="1"
@@ -62,7 +62,7 @@
             </parameterBinding>
          </functionParameter>
          <functionParameter
-               description="Lenient parsing of dates; default is &quot;true&quot;;&#xA; see https://docs.oracle.com/javase/8/docs/api/java/text/DateFormat.html for more details"
+               description="Lenient parsing of dates, default is true.&#xA; See https://docs.oracle.com/javase/8/docs/api/java/text/DateFormat.html for more details"
                label="Leniency"
                maxOccurrence="1"
                minOccurrence="0"
--- a/cst/plugins/eu.esdihumboldt.cst.functions.string/plugin.xml
+++ b/cst/plugins/eu.esdihumboldt.cst.functions.string/plugin.xml
@@ -50,8 +50,8 @@
             </property>
          </targetProperties>
          <functionParameter
-               description="For example &quot;yyyy-MM-dd HH:mm:ss&quot;&#xA;(see http://docs.oracle.com/javase/6/docs/api/java/text/SimpleDateFormat.html)"
-               label="Date format"
+               description="For example: yyyy-MM-dd HH:mm:ss;&#xA; see http://docs.oracle.com/javase/6/docs/api/java/text/SimpleDateFormat.html for more details"
+               label="Source date format"
                maxOccurrence="1"
                minOccurrence="1"
                name="dateFormat">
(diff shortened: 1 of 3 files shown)


cst/plugins/eu.esdihumboldt.cst.functions.string/plugin.xml
```diff
@@ -50,8 +50,8 @@
             </property>
          </targetProperties>
          <functionParameter
-               description="For example &quot;yyyy-MM-dd HH:mm:ss&quot;&#xA;(see http://docs.oracle.com/javase/6/docs/api/java/text/SimpleDateFormat.html)"
-               label="Date format"
+               description="For example: yyyy-MM-dd HH:mm:ss;&#xA; see http://docs.oracle.com/javase/6/docs/api/java/text/SimpleDateFormat.html for more details"
```
stempler, line 53: I guess you might also want to remove the semicolon that belonged to the second `&quot;` then?

cst/plugins/eu.esdihumboldt.cst.functions.string/src/eu/esdihumboldt/cst/functions/string/DateExtraction.java
```diff
@@ -57,6 +57,12 @@ protected Object evaluate(String transformationIdentifier, TransformationEngine
 					.format("Mandatory parameter {0} not defined", PARAMETER_DATE_FORMAT));
 		}
 
+		if (getParameters() == null || getParameters().get(PARAMETER_LENIENCY) == null
+				|| getParameters().get(PARAMETER_LENIENCY).isEmpty()) {
+			throw new TransformationException(MessageFormat
+					.format("Mandatory parameter {0} not defined", PARAMETER_LENIENCY));
```
stempler, line 63: It's not a mandatory parameter, right?

cst/plugins/eu.esdihumboldt.cst.functions.string/src/eu/esdihumboldt/cst/functions/string/DateExtraction.java
```diff
@@ -65,6 +71,13 @@ protected Object evaluate(String transformationIdentifier, TransformationEngine
 		String sourceString = variables.values().iterator().next().getValueAs(String.class);
 		SimpleDateFormat sdf = new SimpleDateFormat(dateFormat);
 
+		// default ....
+		String leniency = getParameters().get(PARAMETER_LENIENCY).get(0).as(String.class);
```
stempler, line 75: I think this would be a bit cleaner:

```
boolean leniency = getOptionalParameter(PARAMETER_LENIENCY, Value.of(true)).as(Boolean.class);
```

cst/plugins/eu.esdihumboldt.cst.functions.string/src/eu/esdihumboldt/cst/functions/string/DateExtraction.java
```diff
@@ -65,6 +71,13 @@ protected Object evaluate(String transformationIdentifier, TransformationEngine
...
+		leniency = getExecutionContext().getVariables().replaceVariables(leniency);
```
stempler, line 76: Only true/false supported, so variable replacement is not really applicable here.

kapil-agnihotri (changes requested):

cst/plugins/eu.esdihumboldt.cst.functions.string/plugin.xml
```diff
@@ -61,6 +61,19 @@
                </validator>
             </parameterBinding>
          </functionParameter>
+         <functionParameter
+               description="Lenient parsing of dates; default is &quot;true&quot;;&#xA; see https://docs.oracle.com/javase/8/docs/api/java/text/DateFormat.html for more details"
+               label="Leniency"
+               maxOccurrence="1"
+               minOccurrence="0"
```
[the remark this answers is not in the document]
stempler: Yes it should be `0` for backwards compatibility. Also, otherwise a default value does not make sense.

cst/plugins/eu.esdihumboldt.cst.functions.string/src/eu/esdihumboldt/cst/functions/string/DateExtraction.java
Flaminietta: Not anymore, right.

kapil-agnihotri (approved):

cst/plugins/eu.esdihumboldt.cst.functions.string/src/eu/esdihumboldt/cst/functions/string/DateExtraction.java
```diff
@@ -65,6 +66,14 @@ protected Object evaluate(String transformationIdentifier, TransformationEngine
 		String sourceString = variables.values().iterator().next().getValueAs(String.class);
 		SimpleDateFormat sdf = new SimpleDateFormat(dateFormat);
 
+		// by default setLenient(true)
+		boolean leniency = getOptionalParameter(PARAMETER_LENIENCY, Value.of(true))
+				.as(Boolean.class);
+
+		if (leniency == false) {
+			sdf.setLenient(false);
```
stempler, line 74: As you have a boolean now you could also just set it to the value of `leniency` in any case.

stempler (approved):

cst/plugins/eu.esdihumboldt.cst.functions.string/src/eu/esdihumboldt/cst/functions/string/DateExtraction.java
Flaminietta: This is also done and tested. I merge it.

PR #952, opened 2022-02-23
feat: add option to specify the initial value for the sequential ID generator function.

Flaminietta:
The default value is 1.

ING-3284

### DIFF
--- a/cst/plugins/eu.esdihumboldt.cst.functions.numeric/plugin.xml
+++ b/cst/plugins/eu.esdihumboldt.cst.functions.numeric/plugin.xml
@@ -95,6 +95,12 @@
                maxOccurrence="1"
                minOccurrence="1"
                name="startValue">
+            <parameterBinding
+                   class="java.lang.Integer">
+            </parameterBinding>      
+              <valueDescriptor
+                   default="1">
+              </valueDescriptor>
          </functionParameter>
          <functionParameter
                description="Identifier prefix"
--- a/cst/plugins/eu.esdihumboldt.cst.functions.numeric/plugin.xml
+++ b/cst/plugins/eu.esdihumboldt.cst.functions.numeric/plugin.xml
@@ -93,7 +93,7 @@
                description="Starting numerical value for the sequence"
                label="StartValue"
                maxOccurrence="1"
-               minOccurrence="1"
+               minOccurrence="0"
                name="startValue">
             <parameterBinding
                    class="java.lang.Integer">
--- a/cst/plugins/eu.esdihumboldt.cst.functions.numeric/plugin.xml
+++ b/cst/plugins/eu.esdihumboldt.cst.functions.numeric/plugin.xml
@@ -75,6 +75,28 @@
             </property>
          </targetProperties>
          <functionParameter
+               description="Generation sequence to use"
+               label="Sequence"
+               maxOccurrence="1"
+               minOccurrence="1"
+               name="sequence">
+            <parameterEnumeration>
+               <enumerationValue
+                     value="overall">
+               </enumerationValue>
+               <enumerationValue
+                     value="type">
+               </enumerationValue>
+            </parameterEnumeration>
+         </functionParameter>
+         <functionParameter
+               description="Starting numerical value for the sequence"
+               label="StartValue"
+               maxOccurrence="1"
+               minOccurrence="1"
+               name="startValue">
+         </functionParameter>
+         <functionParameter
                description="Identifier prefix"
                label="Prefix"
                maxOccurrence="1"
@@ -91,21 +113,6 @@
                name="suffix">
             <parameterBinding>
             </parameterBinding>
-         </functionParameter>
-         <functionParameter
-               description="Generation sequence to use"
-               label="Sequence"
-               maxOccurrence="1"
-               minOccurrence="1"
-               name="sequence">
-            <parameterEnumeration>
-               <enumerationValue
-                     value="overall">
-               </enumerationValue>
-               <enumerationValue
-                     value="type">
-               </enumerationValue>
-            </parameterEnumeration>
          </functionParameter>
       </propertyFunction>
    </extension>
--- a/cst/plugins/eu.esdihumboldt.cst.functions.numeric/src/eu/esdihumboldt/cst/functions/numeric/sequentialid/SequentialID.java
+++ b/cst/plugins/eu.esdihumboldt.cst.functions.numeric/src/eu/esdihumboldt/cst/functions/numeric/sequentialid/SequentialID.java
@@ -48,7 +48,6 @@ public class SequentialID extends AbstractSingleTargetPropertyTransformation<Tra
 			PropertyEntityDefinition resultProperty, Map<String, String> executionParameters,
 			TransformationLog log) throws TransformationException, NoResultException {
 		// get parameter values
-
 		String prefix = getOptionalParameter(PARAM_PREFIX, Value.of("")).as(String.class);
 		String suffix = getOptionalParameter(PARAM_SUFFIX, Value.of("")).as(String.class);
 
@@ -59,6 +58,10 @@ public class SequentialID extends AbstractSingleTargetPropertyTransformation<Tra
 		// assume type as default for sequence
 		String sequenceStr = getOptionalParameter(PARAM_SEQUENCE, Value.of(Sequence.type.name()))
 				.as(String.class);
+
+		// get starting value for the sequence - default is 1
+		String startValue = "1";
+		startValue = getOptionalParameter(START_VALUE, Value.of("")).as(String.class);
 
 		// select appropriate context and key
 		Sequence sequence = Sequence.valueOf(sequenceStr);
@@ -85,7 +88,7 @@ public class SequentialID extends AbstractSingleTargetPropertyTransformation<Tra
 				id = seqValue + 1;
 			}
 			else {
-				id = START_VALUE;
+				id = Integer.parseInt(startValue);
 			}
 			context.put(key, id);
 		}
--- a/cst/plugins/eu.esdihumboldt.cst.functions.numeric/src/eu/esdihumboldt/cst/functions/numeric/sequentialid/SequentialID.java
+++ b/cst/plugins/eu.esdihumboldt.cst.functions.numeric/src/eu/esdihumboldt/cst/functions/numeric/sequentialid/SequentialID.java
@@ -50,7 +50,7 @@ public class SequentialID extends AbstractSingleTargetPropertyTransformation<Tra
 		// get parameter values
 		String prefix = getOptionalParameter(PARAM_PREFIX, Value.of("")).as(String.class);
 		String suffix = getOptionalParameter(PARAM_SUFFIX, Value.of("")).as(String.class);
-		String startValue = getOptionalParameter(START_VALUE, Value.of("1")).as(String.class);
+		String startValue = getOptionalParameter(PARAM_START_VALUE, Value.of("1")).as(String.class);
 
 		// replace transformation variables in prefix, suffix and startValue
 		prefix = getExecutionContext().getVariables().replaceVariables(prefix);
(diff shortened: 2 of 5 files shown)


cst/plugins/eu.esdihumboldt.cst.functions.numeric/plugin.xml
```diff
@@ -74,6 +74,28 @@
...
+               <enumerationValue
+                     value="type">
+               </enumerationValue>
+            </parameterEnumeration>
+         </functionParameter>
+         <functionParameter
+               description="Starting numerical value for the sequence"
+               label="StartValue"
+               maxOccurrence="1"
+               minOccurrence="1"
```
stempler, line 96: It should be an optional parameter for backwards compatibility.

cst/plugins/eu.esdihumboldt.cst.functions.numeric/src/eu/esdihumboldt/cst/functions/numeric/sequentialid/SequentialID.java
```diff
@@ -60,6 +59,10 @@ protected Object evaluate(String transformationIdentifier, TransformationEngine
 		String sequenceStr = getOptionalParameter(PARAM_SEQUENCE, Value.of(Sequence.type.name()))
 				.as(String.class);
 
+		// get starting value for the sequence - default is 1
+		String startValue = "1";
+		startValue = getOptionalParameter(START_VALUE, Value.of("")).as(String.class);
```
stempler, line 64: If you use a default of an empty string, the default value will not be one, but cause an error (because an empty string cannot be converted to an integer later on).

ui/plugins/eu.esdihumboldt.hale.ui.functions.numeric/src/eu/esdihumboldt/hale/ui/functions/numeric/SequentialIDParameterPage.java
```diff
@@ -235,10 +239,43 @@ public void modifyText(ModifyEvent e) {
...
+			controlLayout.applyTo(startValue);
+
+			startValue.setText(
+					getOptionalInitialValue(START_VALUE, new ParameterValue("1")).as(String.class));
+
+			startValue.addModifyListener(new ModifyListener() {
+
+				@Override
+				public void modifyText(ModifyEvent e) {
+					updateStatus();
```
stempler, line 258: Would you expect any validation of what was entered?

cst/plugins/eu.esdihumboldt.cst.functions.numeric/src/eu/esdihumboldt/cst/functions/numeric/sequentialid/SequentialID.java
```diff
@@ -85,7 +88,7 @@ protected Object evaluate(String transformationIdentifier, TransformationEngine
 				id = seqValue + 1;
 			}
 			else {
-				id = START_VALUE;
+				id = Integer.parseInt(startValue);
```
stempler, line 91: In the UI you allow to specify project variables, so you should do
`startValue = getExecutionContext().getVariables().replaceVariables(startValue);`
beforehand.

cst/plugins/eu.esdihumboldt.cst.functions.numeric/plugin.xml
Flaminietta: I added now a couple of lines for setting the default value: 
```
<functionParameter
               description="Starting numerical value for the sequence"
               label="StartValue"
               maxOccurrence="1"
               minOccurrence="1"
               name="startValue">
            <parameterBinding
                   class="java.lang.Integer">
            </parameterBinding>      
              <valueDescriptor
                   default="1">
              </valueDescriptor>
         </functionParameter>
```
However, I don't see a different behavior of hs: when the "Generate sequential ID" window pops up, I see, as I have seen before, that the startValue is set on "1". Do I miss something?

Flaminietta: Solved! By addressing your second comment I see the right behavior with a default value of 1 if I leave the StartValue empty.

ui/plugins/eu.esdihumboldt.hale.ui.functions.numeric/src/eu/esdihumboldt/hale/ui/functions/numeric/SequentialIDParameterPage.java
Flaminietta: I honestly copied it from the above paragraph ("specify suffix") with a basic understanding of what is happening here. 

I would not expect a validation on the integer value because the int can be anything. But I would expect it to be an integer and nothing else.

What should I do?

cst/plugins/eu.esdihumboldt.cst.functions.numeric/plugin.xml
stempler: `minOccurrence` should still be `0`, because the function did already exist before, and any project where it was used before won't have that parameter.

cst/plugins/eu.esdihumboldt.cst.functions.numeric/src/eu/esdihumboldt/cst/functions/numeric/sequentialid/SequentialID.java
stempler: How do you think this was resolved?

I would suggest something like:

```
String startValue = getOptionalParameter(START_VALUE, Value.of("1")).as(String.class);
```

cst/plugins/eu.esdihumboldt.cst.functions.numeric/src/eu/esdihumboldt/cst/functions/numeric/sequentialid/SequentialID.java
stempler: How do you think this was resolved? I don't see a change.

ui/plugins/eu.esdihumboldt.hale.ui.functions.numeric/src/eu/esdihumboldt/hale/ui/functions/numeric/SequentialIDParameterPage.java
stempler: Right now the user should enter an integer or use a project variable.
But they can enter any string - so it would be quite easy to enter something that is not valid and leads to an error.

Doing some kind of validation (e.g. further below where the example is generated) could help - but I think it is also fine to not do it - then the user just needs to deal with the potential transformation error.

Another way to support the user would be to make sure that if they use a wrong value (which you can't for sure check in the function editor always, as they could use project variables), the error message from the transformation is in such a way that it is easily clear to users what the problem is and how they can solve it.

Flaminietta: I've tried out my solution for the validation of `startValue`, please check it in my latest commit. My logic:

- in `SequentialIDParameterPage.java`, `startValue` is of type string; I don't want to modify its type and I don't see an easy way to validate the content here, for a string that - for its intrinsic nature  - can be anything;
- I rather think that the validation can happen where the string is parsed into an integer, that is in `SequentialID.java`;
- I use the logic `try{} catch() {}` to modify the error message;
- I believe that the error can only be generated by the user inputting a string that does not represent an integer (any sequence of chars or a float given as a string)
- I verified that - if in the app one gives a sequence of chars or a float number - then one gets an `Instance validation warning` with the wished message.

cst/plugins/eu.esdihumboldt.cst.functions.numeric/src/eu/esdihumboldt/cst/functions/numeric/sequentialid/SequentialID.java
```diff
@@ -78,16 +79,21 @@ protected Object evaluate(String transformationIdentifier, TransformationEngine
...
 			if (seqValue != null) {
 				id = seqValue + 1;
 			}
 			else {
-				id = START_VALUE;
+				try {
+					id = Integer.parseInt(startValue);
+					context.put(key, id);
+				} catch (Exception E) {
+					System.out.println("Error with the input value of " + startValue
```
stempler, line 93: I would rather be for letting this fail rather than catching the error and continuing.

I think it should work to throw an exception here with an adapted message and the original exception as cause.

cst/plugins/eu.esdihumboldt.cst.functions.numeric/src/eu/esdihumboldt/cst/functions/numeric/sequentialid/SequentialIDConstants.java
```diff
@@ -45,10 +45,8 @@
 
 	/**
 	 * The start value for sequences.
-	 * 
-	 * TODO make this configurable?
 	 */
-	public static final int START_VALUE = 1;
+	public static final String START_VALUE = "startValue";
```
stempler, line 49: As you changed the semantics maybe also change the name of the constant? e.g. to `PARAM_START_VALUE` following the pattern for the other function parameter names?

cst/plugins/eu.esdihumboldt.cst.functions.numeric/src/eu/esdihumboldt/cst/functions/numeric/sequentialid/SequentialID.java
```diff
@@ -78,16 +79,21 @@ protected Object evaluate(String transformationIdentifier, TransformationEngine
...
-				id = START_VALUE;
+				try {
+					id = Integer.parseInt(startValue);
+					context.put(key, id);
+				} finally {
+					System.out.println("ERROR with the input value of startValue  " + startValue
+							+ " for ID creation: the input value should be an integer");
+				}
 			}
-			context.put(key, id);
```
stempler, line 90: This line needs to still apply both for if and else case as before - otherwise you won't have a sequence but always the start value used.

cst/plugins/eu.esdihumboldt.cst.functions.numeric/src/eu/esdihumboldt/cst/functions/numeric/sequentialid/SequentialID.java
```diff
@@ -78,16 +79,21 @@ protected Object evaluate(String transformationIdentifier, TransformationEngine
...
 			if (seqValue != null) {
 				id = seqValue + 1;
 			}
 			else {
-				id = START_VALUE;
+				try {
+					id = Integer.parseInt(startValue);
+					context.put(key, id);
+				} finally {
+					System.out.println("ERROR with the input value of startValue  " + startValue
```
stempler, line 93: Like this, this will be printed regardless of if there was an error or not.
Also, simply printing to the console like that is not how logging is handled within a transformation.

As I said, rethrowing the exception is probably the easiest way here, e.g.

```
} catch(Exception e) {
  throw new TransformationException("My improved error message", e);
}
```

cst/plugins/eu.esdihumboldt.cst.functions.numeric/src/eu/esdihumboldt/cst/functions/numeric/sequentialid/SequentialID.java
```diff
@@ -78,16 +79,21 @@ protected Object evaluate(String transformationIdentifier, TransformationEngine
 			break;
 		}
 
-		long id;
+		long id = 1;
```
stempler, line 82: I think it would make sense to remove the initialization here as that is never used.

stempler (approved):

Flaminietta:
Also cosmetic has its value, I made it again :-)

PR #953, opened 2022-03-08
feat: add option to save on CSV preserving the original order of columns

Flaminietta:
Add an option to save transformed data to a CSV file preserving the original number and order of columns as in the source schema.

ING-3287


Flaminietta:
In the second commit 74ac1b0:

- inserts a new button `useSchema` to allow the user to decide if the order of the exported columns for the transformed data is the one read from the source schema (i.e. the order is preserved) or from the instances (as before)
- adds an extra unit test to distinguish between the simple assign test with `useSchema` true and false
- solves the issue for CSV files

TO CHECK:
- why in the test `testWriteSimpleSchema` the column with no values inside (i.e. b1) is not dropped?

TODO:
- same for XLS?

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/writer/AbstractTableInstanceWriter.java
```diff
@@ -58,8 +60,69 @@ public boolean isCancelable() {
 	protected Map<String, Object> getPropertyMap(Instance instance, List<String> headerRow,
 			boolean solveNestedProperties) {
 		// properties of current instance
-		Iterable<QName> allProperties = instance.getPropertyNames();
+		Iterable<QName> allProperties = DefinitionUtil.getAllProperties(instance.getDefinition())
```
stempler, line 63: This looks like the method below is basically duplicated here. I think ideally would be to remove this one so it is always explicit if `useSchema` is enabled or not.

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/writer/internal/CSVInstanceWriter.java
```diff
@@ -146,10 +145,13 @@ public boolean isPassthrough() {
 	}
 
 	// write current instance to csv file
-	private void writeLine(boolean solveNestedProperties, List<String> headerRow,
-			Instance instance, CSVWriter writer) {
+	private void writeLine(boolean solveNestedProperties, List<String> headerRow, Instance instance,
+			CSVWriter writer) {
+		boolean useSchema = getParameter(InstanceTableIOConstants.USE_SCHEMA).as(Boolean.class,
```
stempler, line 150: Should be an optional parameter to be backwards compatible.

Though as we said the original behavior is a bug, I guess the default should then be `true`.

Remember that when you add a new parameter for an I/O provider you also need to define it in the related extension in `plugin.xml`.

io/plugins/eu.esdihumboldt.hale.io.csv.test/src/eu/esdihumboldt/hale/io/csv/writer/CSVInstanceWriterTest.java
```diff
@@ -76,42 +77,112 @@ public void testWriteSimpleSchema() throws Exception {
...
 		List<String[]> rows = reader.readAll();
-		//
+
 		reader.close();
 
 		assertEquals("Not enough rows.", numberOfRows, rows.size());
 
+		// Check no. of columns ###
+		int countCols = rows.get(0).length;
+		// TEST PRINT I see that, even if b1 has no values in instance1.xml,
```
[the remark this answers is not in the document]
Flaminietta: How can I then reproduce the scenario `no value`, for which the column is dropped? When does it happen in practice?

cst/plugins/eu.esdihumboldt.cst.test/src/testdata/simpleassign/instance1.xml
```diff
@@ -4,12 +4,12 @@
  xsi:schemaLocation="http://www.example.org/t1/ t1.xsd">
     <tns:t1 id="id0">
         <a1>a10</a1>
-        <b1>b10</b1>
+        <b1></b1>
```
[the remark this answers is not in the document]
Flaminietta: Should I create a specific test to reproduce then the `no value` case? If yes, how? (See also comment above).
I'll take care to revert this change back in instance1.xml.

io/plugins/eu.esdihumboldt.hale.io.csv.ui/src/eu/esdihumboldt/hale/io/csv/ui/InstanceExportConfigurationPage.java
```diff
@@ -103,6 +104,8 @@ public void disable() {
 	public boolean updateConfiguration(InstanceWriter provider) {
 		provider.setParameter(InstanceTableIOConstants.SOLVE_NESTED_PROPERTIES,
```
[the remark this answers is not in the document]
Flaminietta: I also don't know what this parameter is for and I have seen also comments in the code asking the same question (`CSVInstanceWriter.java` lines 79-81)

However, how do you deduce that this is not used in case useSchema=true?

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/writer/AbstractTableInstanceWriter.java
Flaminietta: OK

io/plugins/eu.esdihumboldt.hale.io.csv.test/src/eu/esdihumboldt/hale/io/csv/writer/CSVInstanceWriterTest.java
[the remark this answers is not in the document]
stempler: You need a schema with an optional first level property, that you don't populate in the data.

A good way to define a schema and instances for test purposes is using the related Groovy builders, as for example here in line 232 to 252:
https://github.com/halestudio/hale/blob/44e54a9dfd067bdf51db427a50de9803213ffb1c/io/plugins/eu.esdihumboldt.hale.io.geopackage.test/src/eu/esdihumboldt/hale/io/geopackage/GeopackageInstanceWriterTest.groovy#L232

io/plugins/eu.esdihumboldt.hale.io.csv.ui/src/eu/esdihumboldt/hale/io/csv/ui/InstanceExportConfigurationPage.java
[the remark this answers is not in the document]
stempler: I think I was wrong and it can apply even with `useSchema`. Though I'm not sure what the result is because I did not analyse the `iterateBuild` function.

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/writer/internal/CSVInstanceWriter.java
Flaminietta: Where can I find the right plugin.xml? I have explored some directories but I am not sure I have found the right one. I need to note it down as a recurrent step ;-)

stempler: The bundle where the `CSVInstanceWriter` is defined should also hold the declaration of it as I/O provider in the extensions.

io/plugins/eu.esdihumboldt.hale.io.xls/src/eu/esdihumboldt/hale/io/xls/writer/XLSInstanceWriter.java
```diff
@@ -120,14 +124,15 @@ else if (getContentType().getId().equals("eu.esdihumboldt.hale.io.xls.xlsx")) {
...
+		writeRow(row,
+				super.getPropertyMap(instance, headerRowStrings, solveNestedProperties, useSchema));
 
 		while (instanceIterator.hasNext()) {
 			Instance nextInst = instanceIterator.next();
 			if (nextInst.getDefinition().equals(definition)) {
 				row = sheet.createRow(rowNum++);
-				writeRow(row,
-						super.getPropertyMap(nextInst, headerRowStrings, solveNestedProperties));
+				writeRow(row, super.getPropertyMap(nextInst, headerRowStrings,
```
stempler, line 134: The method declaration is
```
protected Map<String, Object> getPropertyMap(Instance instance, List<String> headerRow,
			boolean useSchema, boolean solveNestedProperties)
```

So I think you pass the arguments the wrong way.

Flaminietta: sorry!

io/plugins/eu.esdihumboldt.hale.io.csv.test/src/eu/esdihumboldt/hale/io/csv/writer/CSVInstanceWriterTest.java
```diff
@@ -76,42 +77,112 @@ public void testWriteSimpleSchema() throws Exception {
...
+
+		assertEquals("Not enough rows.", numberOfRows, rows.size());
+
+		// Check no. of columns ###
+		int countCols = rows.get(0).length;
+		assertEquals("Not enough columns written in the output file.", numberOfColumns, countCols);
+		// Check the order of the columns ###
+		String[] originalColumns = propertyNames.split(",");
+		int j = 0;
+		for (String s : originalColumns) {
```
kapil-agnihotri, line 166: Since you are not using `s` in the loop nor the order from the originalColumn and only its length, the loop can be modified as: 

```
for (int j = 0; j < originalColumns.length; j++) {
                      
                    }
```

At first looking at the loop, it appears as if you would be using `s` inside it but as it is not the case so I would advise to change the loop to use `originalColumns.length`.

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/writer/internal/CSVInstanceWriter.java
Flaminietta: I would then insert the new parameter in this section of the plugin:
```
<provider
            class="eu.esdihumboldt.hale.io.csv.writer.internal.CSVInstanceWriter"
            id="eu.esdihumboldt.hale.io.csv.writer.instance"
            name="CSV file">
         <contentType
               ref="eu.esdihumboldt.hale.io.csv">
         </contentType>
         <providerParameter
               description="Schema type which instances should be exported."
               label="Export type"
               name="selectedExportType"
               optional="false"
               valueDescriptor="eu.esdihumboldt.hale.io.csv.TypenameParameterDescriptor">
            <parameterBinding
                  class="java.lang.String">
            </parameterBinding>
         </providerParameter>
         <providerParameter
               description="Separator character that delimits columns in the CSV file"
               label="Separator character"
               name="separator"
               optional="true">
            <parameterBinding
                  class="java.lang.String">
            </parameterBinding>
            <valueDescriptor
                  defaultDescription="Defaults to tabulator (\t)">
            </valueDescriptor>
         </providerParameter>
         <providerParameter
               description="Character used to quote CSV entries"
               label="Quote character"
               name="quote"
               optional="true">
            <parameterBinding
                  class="java.lang.String">
            </parameterBinding>
            <valueDescriptor
                  default="&quot;">
            </valueDescriptor>
         </providerParameter>
         <providerParameter
               description="Character used to escape special characters, e.g. the quote character"
               label="Escape character"
               name="escape"
               optional="true">
            <parameterBinding
                  class="java.lang.String">
            </parameterBinding>
            <valueDescriptor
                  default="\">
            </valueDescriptor>
         </providerParameter>
      </provider>
   </extension>
```
but why there is no `solveNestedProperties` in here too?

io/plugins/eu.esdihumboldt.hale.io.csv.test/src/eu/esdihumboldt/hale/io/csv/writer/CSVInstanceWriterTest.java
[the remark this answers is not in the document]
Flaminietta: TODO

io/plugins/eu.esdihumboldt.hale.io.csv.test/src/eu/esdihumboldt/hale/io/csv/writer/CSVInstanceWriterTest.java
Flaminietta: I think `s` it is used, look in the assert at line 173:
```
for (String s : originalColumns) {
			assertEquals("The order of the columns is not equal to the original source file", s,
					rows.get(0)[j]);
			j++;
		}
```

io/plugins/eu.esdihumboldt.hale.io.csv.test/META-INF/MANIFEST.MF
```diff
@@ -7,8 +7,13 @@ Fragment-Host: eu.esdihumboldt.hale.io.csv;bundle-version="2.5.0"
...
+ eu.esdihumboldt.hale.common.instance.groovy,
+ eu.esdihumboldt.hale.common.schema.groovy,
  eu.esdihumboldt.hale.common.test,
+ eu.esdihumboldt.util.groovy.sandbox,
  org.junit;version="4.8.1",
  org.junit.rules;version="4.11.0"
 Bundle-Vendor: data harmonisation panel
 Automatic-Module-Name: eu.esdihumboldt.hale.io.csv.test
+Require-Bundle: eu.esdihumboldt.util.groovy;bundle-version="4.2.0",
+ groovy-all;bundle-version="2.4.14"
```
stempler, line 19: The bundle that should be included as dependency for Groovy should be `groovy` instead of `groovy-all`.

io/plugins/eu.esdihumboldt.hale.io.xls.test/META-INF/MANIFEST.MF
```diff
@@ -12,14 +12,16 @@ Require-Bundle: org.junit,
  eu.esdihumboldt.hale.common.instance;bundle-version="2.9.0",
  eu.esdihumboldt.hale.common.schema;bundle-version="2.9.0",
  eu.esdihumboldt.hale.io.xls;bundle-version="2.9.0",
- eu.esdihumboldt.hale.common.schema.persist;bundle-version="2.9.0"
+ eu.esdihumboldt.hale.common.schema.persist;bundle-version="2.9.0",
+ groovy-all;bundle-version="2.4.14"
```
stempler, line 16: The bundle that should be included as dependency for Groovy should be `groovy` instead of `groovy-all`.

io/plugins/eu.esdihumboldt.hale.io.xls.test/plugin.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<?eclipse version="3.4"?>
+<plugin>
+   <extension
+         point="eu.esdihumboldt.hale.io.csv.propertyType">
```
stempler, line 5: Probably this file was added accidentally? If so, then the build.properties above also does not need to be changed.

io/plugins/eu.esdihumboldt.hale.io.xls.test/META-INF/MANIFEST.MF
```diff
@@ -12,14 +12,16 @@ Require-Bundle: org.junit,
...
 Import-Package: com.google.common.io;version="15.0.0",
  eu.esdihumboldt.cst.test,
  eu.esdihumboldt.hale.common.instance.groovy,
  eu.esdihumboldt.hale.common.schema.groovy,
  eu.esdihumboldt.hale.common.schema.model,
  eu.esdihumboldt.hale.common.schema.model.constraint.type,
  eu.esdihumboldt.hale.common.test,
+ eu.esdihumboldt.hale.io.csv.ui,
```
stempler, line 24: There should not ever be a dependency to a UI bundle from a non UI bundle.

io/plugins/eu.esdihumboldt.hale.io.xls.test/src/eu/esdihumboldt/hale/io/xls/test/writer/XLSInstanceWriterTestUtil.groovy
```diff
@@ -0,0 +1,48 @@
...
+
+package eu.esdihumboldt.hale.io.xls.test.writer
+
+import eu.esdihumboldt.hale.common.instance.groovy.InstanceBuilder
+import eu.esdihumboldt.hale.common.instance.model.InstanceCollection
+import eu.esdihumboldt.hale.common.schema.groovy.SchemaBuilder
+import eu.esdihumboldt.hale.common.schema.model.Schema
+
+/**
+ * TODO Type description
```
stempler, line 24: Resolve TODO?

io/plugins/eu.esdihumboldt.hale.io.csv.test/src/eu/esdihumboldt/hale/io/csv/writer/CSVInstanceWriterTestUtil.groovy
```diff
@@ -0,0 +1,49 @@
...
+
+package eu.esdihumboldt.hale.io.csv.writer
+
+import eu.esdihumboldt.hale.common.instance.groovy.InstanceBuilder
+import eu.esdihumboldt.hale.common.instance.model.InstanceCollection
+import eu.esdihumboldt.hale.common.schema.groovy.SchemaBuilder
+import eu.esdihumboldt.hale.common.schema.model.Schema
+
+/**
+ * TODO Type description
```
stempler, line 24: Resolve TODO?

io/plugins/eu.esdihumboldt.hale.io.csv/plugin.xml
```diff
@@ -156,7 +156,16 @@
             name="CSV file">
          <contentType
                ref="eu.esdihumboldt.hale.io.csv">
-         </contentType>
+         </contentType>        
+         <providerParameter
```
stempler, line 160: Please add a `valueDescriptor` element that specifies what the default value is.

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/writer/AbstractTableInstanceWriter.java
```diff
@@ -79,45 +90,30 @@ public boolean isCancelable() {
...
+					if (property instanceof Group && shouldBeDisplayed(property)) {
+						checkValue((Group) property, headerRow, row, cellValue);
+					}
+					else {
+						addProperty(headerRow, row, property, cellValue);
+					}
+
+				}
+			}
+			else {
```
stempler, line 106: To retain the previous behavior in case `useSchema` is not enabled, shouldn't this be `else if (useSchema)` then?

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/writer/AbstractTableInstanceWriter.java
```diff
@@ -79,45 +90,30 @@ public boolean isCancelable() {
...
+				String cellValue = "";
+				Object property = null;
+				if (shouldBeDisplayed(property)) {
+					cellValue = qname.getLocalPart();
+				}
+
+				// if property is an OInstance or OGroup, it's a nested property
+				if (solveNestedProperties && property instanceof Group) {
```
stempler, line 114: As it is clear that `property` is `null` you can greatly simplify the whole `if/else` block by just replacing it with this:

```
addProperty(headerRow, row, property, cellValue);
```

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/writer/internal/CSVInstanceWriter.java
stempler: > Should be an optional parameter to be backwards compatible.

This is still true. You should use `getOptionalParameter(...).

io/plugins/eu.esdihumboldt.hale.io.xls/src/eu/esdihumboldt/hale/io/xls/writer/XLSInstanceWriter.java
```diff
@@ -112,6 +112,9 @@ else if (getContentType().getId().equals("eu.esdihumboldt.hale.io.xls.xlsx")) {
 
 		headerRowStrings = new ArrayList<String>();
 
+		boolean useSchema = getParameter(InstanceTableIOConstants.USE_SCHEMA).as(Boolean.class,
```
stempler, line 115: Same as in the CSV writer - should use `getOptionalParameter(...)`

io/plugins/eu.esdihumboldt.hale.io.csv/plugin.xml
stempler, line 160: Adding the same parameter is missing for the XLS writer.

io/plugins/eu.esdihumboldt.hale.io.xls/src/eu/esdihumboldt/hale/io/xls/writer/XLSInstanceWriter.java
Flaminietta: The method `getOptionalParameter` is undefined for the type `CSVInstanceWriter` or `XLSInstanceWriter`. I think that for using this 'new' method I should then also add it into `AbstractIOProvider`, maybe like this:

         /**
	 * @return the parameter or a default value
	 * 
	 * @see IOProvider#getParameter(String)
	 */
	@Override
	public Value getOptionalParameter(String name) {
		Value value = parameters.get(name);
		if (value == null)
			value = Value.of(true);
		return value;
	}

isn't it? Am I making it more complex than what it really is?

kapil-agnihotri: I believe, what Simon is referring to is as done in the below example, you don't need to override the method, just fetch the parameter value and it is null then return the default value. https://github.com/halestudio/hale/blob/44e54a9dfd067bdf51db427a50de9803213ffb1c/common/plugins/eu.esdihumboldt.hale.common.align/src/eu/esdihumboldt/hale/common/align/transformation/function/impl/AbstractTransformationFunction.java#L129

kapil-agnihotri (commented):
LGTM, you just need to resolve the build errors, multiple tests for csvinstanceWriterTest are failing due to mistmatch in the expected and the actual data.

kapil-agnihotri:
retest this please

io/plugins/eu.esdihumboldt.hale.io.xls/src/eu/esdihumboldt/hale/io/xls/writer/XLSInstanceWriter.java
stempler: I'm sorry, I mixed that up with the transformation functions, fine as it is.

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/writer/internal/CSVInstanceWriter.java
stempler: I'm sorry, I mixed that up with the transformation functions, fine as it is.

io/plugins/eu.esdihumboldt.hale.io.csv/plugin.xml
```diff
@@ -165,6 +165,10 @@
             <parameterBinding
                   class="java.lang.Boolean">
             </parameterBinding>
+            <valueDescriptor
+                  default="&quot;true&quot;"
```
stempler, line 169: Should be the literal value without the extra escaped quotes.

io/plugins/eu.esdihumboldt.hale.io.xls/plugin.xml
```diff
@@ -109,6 +109,19 @@
...
+         <providerParameter
+               description="Use the source schema for the order and number of exported columns."
+               label="Use the source schema for the exported columns"
+               name="useSchema"
+               optional="true">
+            <parameterBinding
+                  class="java.lang.Boolean">
+            </parameterBinding>
+            <valueDescriptor
+                  default="&quot;true&quot;"
```
stempler, line 121: Should be the literal value without the extra escaped quotes.

stempler (approved):
For me just a minor thing remaining, see comments related to `valueDescriptor`.

io/plugins/eu.esdihumboldt.hale.io.xls.test/src/eu/esdihumboldt/hale/io/xls/test/writer/XLSInstanceWriterTest.java
```diff
@@ -175,6 +224,18 @@ private void checkHeader(Sheet sheet, List<String> headerNames) throws Exception
 		}
 	}
 
+	private void checkHeaderOrder(Sheet sheet, List<String> headerNames) throws Exception {
+
+		Row header = sheet.getRow(sheet.getFirstRowNum());
+
```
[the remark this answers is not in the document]
Flaminietta: I believe this is checked at line 109 `checkHeader(sheet, header);` where

Flaminietta:
> 

I re-run CSVInstanceWriterTest and cannot find the bug that now is causing the test to fail:
Screenshot 2022-03-23 at 09 55 13

stempler:
> I re-run CSVInstanceWriterTest and cannot find the bug that now is causing the test to fail:

> eu.esdihumboldt.hale.io.csv.writer.internal.CSVInstanceWriter.insertHeader(CSVInstanceWriter.java:186)

Looking at the code it looks like the error can happen there if the header is empty. Could that be the case?

Flaminietta:
> 

Yes, I think there might still be a problem of logic in the if-statement in AbstractTableInstanceWriter. I am looking at it.

PR #954, opened 2022-03-22
feat(geopackage): better support loading with XML/GML schema

stempler:
When using an XML/GML schema that resembles the Geopackage structure,
so far some information would get lost or not be represented correctly.

The changes here support this use case and thus allow for converting
Geopackage to XML/GML given a suited XML schema.

ING-2671

Also:
- Updated target platform to include (1) the AssertJ library and (2) an update to equinox-test to be able to run individual tests based on the test product (see updated documentation [here](https://github.com/halestudio/hale/wiki/Tests#run-a-single-test-class-or-method))
- When reading Geopackages prefer determining the CRS information from the Geopackage file instead of from the schema


florianesser (approved):
LGTM

kapil-agnihotri (approved):

PR #957, opened 2022-03-25
Add option to automatically add the codespace during export.

Flaminietta:
Add an option to automatically add the codespace to the GML:identifier during export on GML.

ING-3285


Flaminietta:
@stempler @kapil-agnihotri 
I follow the same strategy as for the last ticket by letting you follow my approach in a draft PR. In this way, you can comment and correct me anytime, if needed.

To this point, I was able to add a new button in the dialogue page when exporting data to GML to let the user decide if the codespace should be automatically written.

My struggle now is to understand where exactly this parameter should be read and used for that to happen. My guess is that a String should be now defined in `GMLInstanceWriter`. But from there I get lost. I see potential links with WFS and the InspireInstanceWriter. Can you help?

stempler:
@Flaminietta Did you already check my comment on the ticket where I linked a commit where a somewhat similar thing was done for `nilReason` handling?

kapil-agnihotri:
retest this please

io/plugins/eu.esdihumboldt.hale.io.gml/plugin.xml
```diff
@@ -480,6 +480,20 @@
...
+         <providerParameter
+               description="Automatically add codespace to gml:identifier during GML export."
+               label="Add codespace to gml:identifier"
+               name="xml.add.codespace"
+               optional="true">
+            <parameterBinding
+                  class="java.lang.Boolean">
+            </parameterBinding>
+            <valueDescriptor
+                  default="true"
```
stempler, line 492: Isn't the value you set automatically INSPIRE specific? If it is enabled by default from now that also impacts non-INSPIRE datasets.

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriter.java
```diff
@@ -1597,6 +1606,11 @@ private void writeProperties(Group parent, Collection<? extends ChildDefinition<
 				&& ((GroupPropertyDefinition) parent.getDefinition())
 						.getConstraint(ChoiceFlag.class).isEnabled();
 
+		// DEBUGGING
+		if (parent.getDefinition().getIdentifier().equals("identifier")) {
```
stempler, line 1610: Remove?

Do you know that you can also use conditional breakpoints in Java?

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriter.java
```diff
@@ -1617,6 +1631,13 @@ private void writeProperties(Group parent, Collection<? extends ChildDefinition<
 							}
 						}
 
+						// special case handling: auto add codespace
+						if (getParameter(PARAM_ADD_CODESPACE).as(Boolean.class, true)) {
```
stempler, line 1635: If you check the parameter setting, if a codespace should be automatically added here, what is the function of the `addCodespace` parameter?

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriter.java
```diff
@@ -1632,6 +1653,16 @@ private void writeProperties(Group parent, Collection<? extends ChildDefinition<
 								// default
 								writeAttribute(values[0], propDef);
 							}
+							// special case handling: automatically add
+							// codespace to gml:identifier
+							if ("addCodespace".equals(propDef.getName().getLocalPart())
```
stempler, line 1658: So you expect a property "addCodespace" here? Why?

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriter.java
```diff
@@ -1765,9 +1796,15 @@ private void writeElement(Object value, PropertyDefinition propDef, IOReporter r
 				// write no elements if there is a value or only a nil reason
 				boolean writeElements = !hasValue && !hasOnlyNilReason;
 				boolean isNil = !writeElements && (!hasValue || value == null);
+				boolean isCodespace = !hasValue && !hasOnlyNilReason; // Not
```
stempler, line 1799: What information should this transport?

stempler (changes requested):

Flaminietta:
With the 6b9f4cb commit, I think that the codeSpace is automatically and correctly added to the gml:identifier in the exported data to gml. I also think that:

1. I check if no value is set for codeSpace in lines 1634, 1652 and 1803
2. I check for the parent to be a gml:identifier in line 1803
3. I am not sure what to do to check that the gml:identifier is an INSPIRE type

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriter.java
Flaminietta: ?

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriter.java
Flaminietta: No....

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriter.java
Flaminietta: Maybe it's fine now?

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriter.java
```diff
@@ -1752,6 +1785,8 @@ private void writeElement(Object value, PropertyDefinition propDef, IOReporter r
 			boolean hasValue = propDef.getPropertyType().getConstraint(HasValueFlag.class)
 					.isEnabled();
 
+			boolean isIdentifier = propDef.getDisplayName().equals("identifier");
```
stempler, line 1788: For identifying if this is the correct property, the local part and the namespace should be checked.
The `StreamGmlWriter` already has a property you can use to identify the current GML namespace: https://github.com/halestudio/hale/blob/1a25d6c351035ec4c3e162fdf7db4dcc1ee4d07c/io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriter.java#L260

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriter.java
```diff
@@ -1765,9 +1800,16 @@ private void writeElement(Object value, PropertyDefinition propDef, IOReporter r
 				// write no elements if there is a value or only a nil reason
 				boolean writeElements = !hasValue && !hasOnlyNilReason;
 				boolean isNil = !writeElements && (!hasValue || value == null);
+				boolean isCodespace = isIdentifier || value == null; // Not
```
stempler, line 1803: If it is about determining if a code space should be added, I guess it only makes sense to add the codespace if there is an actual value for the identifier. Otherwise you end up with an identifier that has a `codeSpace` but no value, which I don't think is the expected result.

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriter.java
```diff
@@ -1637,7 +1646,30 @@ private void writeProperties(Group parent, Collection<? extends ChildDefinition<
 						if (values.length > 1) {
 							// TODO warning?!
 						}
-					}
+					} // end if for nilReason
+
+					if (child.getName().getLocalPart().toString().equals("codeSpace")
```
stempler, line 1651: I checked the definition of the schema in hale studio and the property uses an empty namespace, so could use a more accurate check by comparing the full qualified name to a `new QName("codeSpace")`

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriter.java
```diff
@@ -1540,7 +1545,7 @@ protected void writeMember(Instance instance, TypeDefinition type, IOReporter re
 //		Name elementName = GmlWriterUtil.getElementName(type);
 //		writer.writeStartElement(elementName.getNamespaceURI(), elementName.getLocalPart());
 
-		writeProperties(instance, type, true, false, report);
+		writeProperties(instance, type, true, false, true, report);
```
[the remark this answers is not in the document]
Flaminietta: How do I determine whether an instance type is an INSPIRE type or not?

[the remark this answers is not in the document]
Flaminietta: @stempler, my plan is to implement two helper functions to identify 1) if the type is of type INSPIRE (done); 2) to check if the parent is of type `gml:identifier`.

I cannot find the right way of implementing the second function.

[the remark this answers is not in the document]
Flaminietta: In the meanwhile, I have tried other possible solutions. I did not find yet the right one.

Flaminietta:
@stempler @kapil-agnihotri after a long debugging I get an output file that looks maybe right BUT it works like that only if I comment the _if statement_ in lines 1655-1658 of _StreamGmlWriter_:

```
if (getParameter(PARAM_ADD_CODESPACE).as(Boolean.class, true) && isInspireType
							&& parentIsGmlIdentifier) {
						System.out.println("pass");
					       allowWrite = true;
```
then making the two parameters _isInspireType_ and _parentIsGmlIdentifier_ look unuseful. Why?



Screenshot 2022-04-28 at 18 35 54

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/writer/internal/GmlWriterUtil.java
```diff
@@ -129,6 +129,35 @@ public static boolean isID(TypeDefinition type) {
 		}
 	}
 
+	/**
+	 * Determines if the given type is an INSPIRE type
+	 * 
+	 * @param ProdDef the type definition
+	 * @return if the type is an INSPIRE type
+	 */
+	public static boolean isInspireType(PropertyDefinition propDef) {
```
stempler, line 138: We want to determine if the feature type of the instance, where the identifier is defined is an INSPIRE type.

I did not yet get to the place where this method is called, but as a `PropertyDefinition` is passed here, it is basically ruled out that your actually pass the instance type here.
The parameter should be of type `TypeDefinition`.

On the type you should also check it's super types, if any uses the INSPIRE namespace, as it could be a type from an INSPIRE extension.

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriter.java
```diff
@@ -1602,6 +1604,14 @@ private void writeProperties(Group parent, Collection<? extends ChildDefinition<
 
 			if (child.asProperty() != null) {
 				PropertyDefinition propDef = child.asProperty();
+
+				boolean isInspireType = GmlWriterUtil.isInspireType(propDef);
```
stempler, line 1608: You can't determine if the parent of the property is an INSPIRE type by checking the property.

You have the parent group (`parent`) available in this method. What you could do is check if it is an instance of the `Instance` class, and in that case extract the parent `TypeDefinition`. That you could check if it is an INSPIRE type.

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/writer/internal/GmlWriterUtil.java
```diff
@@ -129,6 +129,35 @@ public static boolean isID(TypeDefinition type) {
...
+	}
+
+	/**
+	 * Determines if the given parent is a gml:identifier
+	 * 
+	 * @param QName the QName
+	 * @return true if the parent is a gml:identifier
+	 */
+
+	public static boolean isGmlIdentifier(QName QName) {
```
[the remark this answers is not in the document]
Flaminietta: I have checked and seen that the value of `gmlNs` for my test dataset is always `http://www.opengis.net/gml/3.2`. What should I check exactly? That it contains `gml`?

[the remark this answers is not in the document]
kapil-agnihotri: this `http://www.opengis.net/gml/3.2` itself is gml namespace. 
you might have to check:

```
gmlNs.equals(qName.getNamespaceURI())
```
you can also do a null check for `gmlNs` if it can be null. However, I don't think that it will be `null` 

reference: https://docs.oracle.com/javase/7/docs/api/javax/xml/namespace/QName.html#getNamespaceURI()

Flaminietta:
Still does not work as expected. I never pass 

```
if (getParameter(PARAM_ADD_CODESPACE).as(Boolean.class, true) && isInspireType
							&& parentIsGmlIdentifier && values == null
							&& child.getName().equals(gmlIdentifier))
```

because (maybe) when all of the above conditions are true `attribute` and i`sAttribute` are `false` in the outer if-statement.

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriter.java
```diff
@@ -1602,6 +1604,18 @@ private void writeProperties(Group parent, Collection<? extends ChildDefinition<
...
+
+				boolean isInspireType = false;
+				if (parent instanceof Instance) {
+
+					isInspireType = GmlWriterUtil
+							.isInspireType(((Instance) parent).getDefinition());
+
+				}
+
+				boolean parentIsGmlIdentifier = GmlWriterUtil.isGmlIdentifier(child.getName(),
```
stempler, line 1616: For this, [this comment](https://github.com/halestudio/hale/pull/957#discussion_r843604048) is still relevant.
As you said in your comment [here](https://github.com/halestudio/hale/pull/957#issuecomment-1113422733), you determine if the current child processed is a GML identifier. This is then information that needs to be passed on when processing the children of `child`.
The information on if the parent is within an INSPIRE type also needs to be passed on,
Otherwise you don't have the necessary information present where `codeSpace` is actually handled.
Accessing the parent of the `codeSpace` attribute does not help, because the parent is the GML identifier.

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriter.java
```diff
@@ -1637,7 +1651,27 @@ private void writeProperties(Group parent, Collection<? extends ChildDefinition<
 						if (values.length > 1) {
 							// TODO warning?!
 						}
-					}
+					} // end if special case for nilReason
+
+					// special case handling: automatically add codespace to
+					// gml:identifier
+					QName gmlIdentifier = new QName("codeSpace");
```
stempler, line 1658: The naming of the variable is very misleading.

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriter.java
```diff
@@ -1766,11 +1801,17 @@ private void writeElement(Object value, PropertyDefinition propDef, IOReporter r
 				boolean writeElements = !hasValue && !hasOnlyNilReason;
 				boolean isNil = !writeElements && (!hasValue || value == null);
 
-				// write all children
-				writeProperties(group, group.getDefinition(), writeElements, isNil, report);
+				/*
+				 * // write all children writeProperties(group,
```
stempler, line 1805: Changing the logic here looks to me like something that breaks a lot of things. Child properties should be handled even if there is no value present directly in the parent.

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriter.java
stempler: In the method, there is only one parent, and you are looking at its children.
If the parent is the feature type, then you can find the GML identifier as child.
But when processing the GML identifier it is another method invocation, so then the parent is the GML identifier and you can find codeSpace as one of its children.
That's why I said in the original comment, you need to pass some information on, as they are not available in the context of the method.

Flaminietta:
@stempler with the last changes the functionality seems to work.

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriter.java
stempler: @Flaminietta the test failures probably come from this, but there are also other issues. I'll prepare an update of the PR.

stempler:
@Flaminietta I pushed some changes to your branch, you can inspect them in the most recent commit.
Do you want to do some testing?
In any case before merging I would be for squashing all commits to one.

io/plugins/eu.esdihumboldt.hale.io.gml.test/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriter2Test.groovy
```diff
@@ -68,6 +67,102 @@ class StreamGmlWriter2Test {
 		TestUtil.startConversionService();
 	}
 
+	/**
+	 * Test if a codespace attribute is automatically added to a GML identifier within an INSPIRE type.
```
stempler, line 71: @Flaminietta  A good additional test would be, to test that the codeSpace is not added automatically if the parent type is not an INSPIRE type.

stempler (approved):
@Flaminietta Please squash the commits locally or during the merge and use an appropriate message for the new commit. Thank you!

PR #958, opened 2022-03-30
fixup! feat: add option to specify the initial value for the sequential ID generator function.

Flaminietta:
The starting integer value of any ID sequence must be larger than 0.

ING-3284

### DIFF
--- a/cst/plugins/eu.esdihumboldt.cst.functions.numeric/src/eu/esdihumboldt/cst/functions/numeric/sequentialid/SequentialID.java
+++ b/cst/plugins/eu.esdihumboldt.cst.functions.numeric/src/eu/esdihumboldt/cst/functions/numeric/sequentialid/SequentialID.java
@@ -86,15 +86,15 @@ public class SequentialID extends AbstractSingleTargetPropertyTransformation<Tra
 				id = seqValue + 1;
 			}
 			else {
-				if (startValue.charAt(0) == '0') {
-					throw new TransformationException(
-							"ERROR: the starting value for ID creation cannot have leading 0s or be equal to 0");
-				}
 				try {
 					id = Integer.parseInt(startValue);
+					if (id < 0) {
+						throw new TransformationException(
+								"ERROR: the starting value for ID creation should be a positive integer");
+					}
 				} catch (Exception e) {
 					throw new TransformationException(
-							"ERROR: the starting value for ID creation should be an integer larger than 0");
+							"ERROR: the starting value for ID creation should be an integer");
 				}
 			}
 			context.put(key, id);
--- a/cst/plugins/eu.esdihumboldt.cst.functions.numeric/src/eu/esdihumboldt/cst/functions/numeric/sequentialid/SequentialID.java
+++ b/cst/plugins/eu.esdihumboldt.cst.functions.numeric/src/eu/esdihumboldt/cst/functions/numeric/sequentialid/SequentialID.java
@@ -93,9 +93,9 @@ public class SequentialID extends AbstractSingleTargetPropertyTransformation<Tra
 								"ERROR: the starting value for ID creation should be larger than 0");
 					}
 				} catch (Exception e) {
-					throw new TransformationException(
-							"ERROR: the input value of startValue  " + startValue
-									+ " for ID creation: the input value should be an integer");
+					throw new TransformationException("ERROR: the input value of startValue  "
+							+ startValue
+							+ " for ID creation: the input value should be an integer larger than 0");
 				}
 			}
 			context.put(key, id);
--- a/cst/plugins/eu.esdihumboldt.cst.functions.numeric/src/eu/esdihumboldt/cst/functions/numeric/sequentialid/SequentialID.java
+++ b/cst/plugins/eu.esdihumboldt.cst.functions.numeric/src/eu/esdihumboldt/cst/functions/numeric/sequentialid/SequentialID.java
@@ -86,16 +86,15 @@ public class SequentialID extends AbstractSingleTargetPropertyTransformation<Tra
 				id = seqValue + 1;
 			}
 			else {
+				if (startValue.charAt(0) == '0') {
+					throw new TransformationException(
+							"ERROR: the starting value for ID creation cannot have leading 0s or be equal to 0");
+				}
 				try {
 					id = Integer.parseInt(startValue);
-					if (id == 0) {
-						throw new TransformationException(
-								"ERROR: the starting value for ID creation should be larger than 0");
-					}
 				} catch (Exception e) {
-					throw new TransformationException("ERROR: the input value of startValue  "
-							+ startValue
-							+ " for ID creation: the input value should be an integer larger than 0");
+					throw new TransformationException(
+							"ERROR: the starting value for ID creation should be an integer larger than 0");
 				}
 			}
 			context.put(key, id);
--- a/cst/plugins/eu.esdihumboldt.cst.functions.numeric/src/eu/esdihumboldt/cst/functions/numeric/sequentialid/SequentialID.java
+++ b/cst/plugins/eu.esdihumboldt.cst.functions.numeric/src/eu/esdihumboldt/cst/functions/numeric/sequentialid/SequentialID.java
@@ -23,7 +23,9 @@ import com.google.common.collect.ListMultimap;
 import eu.esdihumboldt.hale.common.align.model.impl.PropertyEntityDefinition;
 import eu.esdihumboldt.hale.common.align.transformation.engine.TransformationEngine;
 import eu.esdihumboldt.hale.common.align.transformation.function.PropertyValue;
+import eu.esdihumboldt.hale.common.align.transformation.function.TransformationException;
 import eu.esdihumboldt.hale.common.align.transformation.function.impl.AbstractSingleTargetPropertyTransformation;
+import eu.esdihumboldt.hale.common.align.transformation.function.impl.NoResultException;
 import eu.esdihumboldt.hale.common.align.transformation.report.TransformationLog;
 import eu.esdihumboldt.hale.common.core.io.Value;
 
@@ -44,7 +46,7 @@ public class SequentialID extends AbstractSingleTargetPropertyTransformation<Tra
 	protected Object evaluate(String transformationIdentifier, TransformationEngine engine,
 			ListMultimap<String, PropertyValue> variables, String resultName,
 			PropertyEntityDefinition resultProperty, Map<String, String> executionParameters,
-			TransformationLog log) throws RuntimeException {
+			TransformationLog log) throws TransformationException, NoResultException {
 		// get parameter values
 		String prefix = getOptionalParameter(PARAM_PREFIX, Value.of("")).as(String.class);
 		String suffix = getOptionalParameter(PARAM_SUFFIX, Value.of("")).as(String.class);
@@ -85,23 +87,23 @@ public class SequentialID extends AbstractSingleTargetPropertyTransformation<Tra
 			}
 			else {
 				if (Integer.parseInt(startValue) == 0) {
-					throw new RuntimeException(
-							"The starting value for ID creation should be larger than 0");
+					throw new TransformationException(
+							"ERROR: the starting value for ID creation should be larger than 0");
+					// anyways, in the UI the user cannot use 0s or special
+					// characters for startValue
 				}
 				try {
 					id = Integer.parseInt(startValue);
 				} catch (Exception e) {
-					throw new RuntimeException(
-							"ERROR with the input value of startValue  " + startValue
+					throw new TransformationException(
+							"ERROR: the input value of startValue  " + startValue
 									+ " for ID creation: the input value should be an integer");
 				}
 			}
 			context.put(key, id);
 		}
 
-		if (prefix.isEmpty() && suffix.isEmpty())
-
-		{
+		if (prefix.isEmpty() && suffix.isEmpty()) {
 			return id;
 		}
 		else {
--- a/cst/plugins/eu.esdihumboldt.cst.functions.numeric/src/eu/esdihumboldt/cst/functions/numeric/sequentialid/SequentialID.java
+++ b/cst/plugins/eu.esdihumboldt.cst.functions.numeric/src/eu/esdihumboldt/cst/functions/numeric/sequentialid/SequentialID.java
@@ -86,14 +86,12 @@ public class SequentialID extends AbstractSingleTargetPropertyTransformation<Tra
 				id = seqValue + 1;
 			}
 			else {
-				if (Integer.parseInt(startValue) == 0) {
-					throw new TransformationException(
-							"ERROR: the starting value for ID creation should be larger than 0");
-					// anyways, in the UI the user cannot use 0s or special
-					// characters for startValue
-				}
 				try {
 					id = Integer.parseInt(startValue);
+					if (id == 0) {
+						throw new TransformationException(
+								"ERROR: the starting value for ID creation should be larger than 0");
+					}
 				} catch (Exception e) {
 					throw new TransformationException(
 							"ERROR: the input value of startValue  " + startValue
--- a/cst/plugins/eu.esdihumboldt.cst.functions.numeric/src/eu/esdihumboldt/cst/functions/numeric/sequentialid/SequentialID.java
+++ b/cst/plugins/eu.esdihumboldt.cst.functions.numeric/src/eu/esdihumboldt/cst/functions/numeric/sequentialid/SequentialID.java
@@ -23,9 +23,7 @@ import com.google.common.collect.ListMultimap;
 import eu.esdihumboldt.hale.common.align.model.impl.PropertyEntityDefinition;
 import eu.esdihumboldt.hale.common.align.transformation.engine.TransformationEngine;
 import eu.esdihumboldt.hale.common.align.transformation.function.PropertyValue;
-import eu.esdihumboldt.hale.common.align.transformation.function.TransformationException;
 import eu.esdihumboldt.hale.common.align.transformation.function.impl.AbstractSingleTargetPropertyTransformation;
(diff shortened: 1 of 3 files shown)


Flaminietta:
@kapil-agnihotri, I have only still one doubt about my fixup: to let the code compile correctly, you'll see that in the class `SequentialID` and method `evaluate` I have changed `throws Exception` with `RuntimeException`. I am not sure that is fully OK.

cst/plugins/eu.esdihumboldt.cst.functions.numeric/src/eu/esdihumboldt/cst/functions/numeric/sequentialid/SequentialID.java
```diff
@@ -46,7 +44,7 @@
 	protected Object evaluate(String transformationIdentifier, TransformationEngine engine,
 			ListMultimap<String, PropertyValue> variables, String resultName,
 			PropertyEntityDefinition resultProperty, Map<String, String> executionParameters,
-			TransformationLog log) throws TransformationException, NoResultException {
+			TransformationLog log) throws RuntimeException {
```
kapil-agnihotri, line 47: Any specific reason to change `throws TransformationException, NoResultException ` by ` throws RuntimeException`?

cst/plugins/eu.esdihumboldt.cst.functions.numeric/src/eu/esdihumboldt/cst/functions/numeric/sequentialid/SequentialID.java
```diff
@@ -86,18 +84,24 @@ protected Object evaluate(String transformationIdentifier, TransformationEngine
 				id = seqValue + 1;
 			}
 			else {
+				if (Integer.parseInt(startValue) == 0) {
```
kapil-agnihotri, line 87: I think that the changes that were proposed in the ticket will not meet by these code changes.
If I understand the requirement correctly then Kate and Thorsten mentioned that the ID should not start with leading 0's, i.e. 0011, 0001, 0000, etc.
Moreover, with your existing changes that happens, i.e. `leading zeros are stripped in the data`. The only thing you need too modify/do 
- is in the `StartValue` text box where the user enters the string: user should not be able to start a value from 0. If a user does this then you need to show some kind of error message in the dialog box saying `a start value cannot start with 0`. 
- `ID example` label to not show any example that has start value starting with 0. (if you do the first steps then this will automatically be handled.) 
With the changes that you have done

kapil-agnihotri (changes requested):

cst/plugins/eu.esdihumboldt.cst.functions.numeric/src/eu/esdihumboldt/cst/functions/numeric/sequentialid/SequentialID.java
Flaminietta: I can go back to TransformationException, NoResultException and add InvalidInputException

cst/plugins/eu.esdihumboldt.cst.functions.numeric/src/eu/esdihumboldt/cst/functions/numeric/sequentialid/SequentialID.java
Flaminietta: I don't agree. I have tested my implementation: if one put 0 as `startValue` then the transformation ends with an error and in `Details` one can see the exception message `"ERROR: the starting value for ID creation should be larger than 0"`.

However, maybe I can also implement your suggestions (in the bullet points) for a more robust fixup.

What do you think?

cst/plugins/eu.esdihumboldt.cst.functions.numeric/src/eu/esdihumboldt/cst/functions/numeric/sequentialid/SequentialID.java
```diff
@@ -86,18 +84,24 @@ protected Object evaluate(String transformationIdentifier, TransformationEngine
...
+					throw new RuntimeException(
+							"The starting value for ID creation should be larger than 0");
+				}
 				try {
 					id = Integer.parseInt(startValue);
 				} catch (Exception e) {
-					throw new TransformationException(
+					throw new RuntimeException(
```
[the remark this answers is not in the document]
Flaminietta: I'll revert it.

cst/plugins/eu.esdihumboldt.cst.functions.numeric/src/eu/esdihumboldt/cst/functions/numeric/sequentialid/SequentialID.java
```diff
@@ -86,18 +84,24 @@ protected Object evaluate(String transformationIdentifier, TransformationEngine
...
 							"ERROR with the input value of startValue  " + startValue
 									+ " for ID creation: the input value should be an integer");
 				}
 			}
 			context.put(key, id);
 		}
 
-		if (prefix.isEmpty() && suffix.isEmpty()) {
+		if (prefix.isEmpty() && suffix.isEmpty())
+
```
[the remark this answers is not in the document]
Flaminietta: I'll correct it.

cst/plugins/eu.esdihumboldt.cst.functions.numeric/src/eu/esdihumboldt/cst/functions/numeric/sequentialid/SequentialID.java
```diff
@@ -86,11 +86,17 @@ protected Object evaluate(String transformationIdentifier, TransformationEngine
 				id = seqValue + 1;
 			}
 			else {
+				if (Integer.parseInt(startValue) == 0) {
```
kapil-agnihotri, line 89: `if (Integer.parseInt(startValue) == 0) {` will throw a NumberFormatException if it is not able to parse the string value to Integer. Also, as it is not in a try catch block your code will explode.
Although, we know that we have done the validations in the UI to make sure that no string value is transferred to this function, I would still use `id`, i.e. `if(id == 0)` for this validation in the `try catch` block below.

ui/plugins/eu.esdihumboldt.hale.ui.functions.numeric/src/eu/esdihumboldt/hale/ui/functions/numeric/SequentialIDParameterPage.java
```diff
@@ -377,13 +377,22 @@ protected String generateExample() {
 		String suffix = (this.suffix == null) ? ("") : (this.suffix.getText());
 		String startValue = (this.startValue == null) ? ("") : (this.startValue.getText());
 
+		String specialCharactersString = "!@#$%&*()'+,-./:;<=>?[]^_`{|}";
+
+		for (int i = 0; i < startValue.length(); i++) {
+			char ch = startValue.charAt(i);
+			if (specialCharactersString.contains(Character.toString(ch))
+					|| startValue.startsWith("0")) {
+				return "StartValue cannot have leading 0s or contain special characters";
```
kapil-agnihotri, line 386: 1. What about alphabets. Are they allowed? are decimals allowed? confirm it first from Thorsten or Kate.
A cleaner way would be to use pattern matcher refer
- https://stackoverflow.com/a/15111450 or
- https://www.geeksforgeeks.org/how-to-check-if-string-contains-only-digits-in-java/ -- point `3 Using Regular Expression:`

2. When you return an error message are you sure the finish button is not enabled?

Flaminietta: I know there are two ways to check for forbidden chars, with or without regex. I have decided without it because I find it more intuitive. Can it stay?

kapil-agnihotri: > I know there are two ways to check for forbidden chars, with or without regex. I have decided without it because I find it more intuitive. Can it stay?

I know this approach is more intuitive and you can clearly see a list of things that are excluded. Also the complexity of this approach and with the one using Patternmatcher is same O(N). However, there are drawbacks of this approach:
- in the line `char ch = startValue.charAt(i);` you are first getting a character from a string and then converting that character back to string with `Character.toString(ch)` to check if it is present in the special characters string. So, this code looks a little extra effort.
- You have to manually add all the special characters that you want to exclude. What if you miss one? For instance you missed `\`.  Debugging the missed one would be a nightmare. 
- If later you decide to exclude alphabets then you would need to add `A, B, C...upto Z` and `a, b, c.... upto z`. or else you need to add some other condition with `||` to validate for alphabets. That will make your code look clumsy.
- So your if condition will look something like this: 

```
if (specialCharactersString.contains(Character.toString(ch)) || startValue.startsWith("0") || Character.isAlphabetic(startValue.tolowercase()))
```
					
But still if you prefer to use the one without regex then it is absolutely fine.

ui/plugins/eu.esdihumboldt.hale.ui.functions.numeric/src/eu/esdihumboldt/hale/ui/functions/numeric/SequentialIDParameterPage.java
```diff
@@ -377,13 +377,22 @@ protected String generateExample() {
 		String suffix = (this.suffix == null) ? ("") : (this.suffix.getText());
 		String startValue = (this.startValue == null) ? ("") : (this.startValue.getText());
 
+		String specialCharactersString = "!@#$%&*()'+,-./:;<=>?[]^_`{|}";
```
florianesser, line 380: @Flaminietta The method `generateExample()` doesn't seem to be the right place to check the validity of the user input. A better place to do that would be the `ModifyListener` of the input field (or a method called from there).

ui/plugins/eu.esdihumboldt.hale.ui.functions.numeric/src/eu/esdihumboldt/hale/ui/functions/numeric/SequentialIDParameterPage.java
florianesser: What's the goal here? If it is to check whether the input can be parsed as an integer, it might be easier to just `try` to parse it and do the error handling in the `catch` block.

Regarding the leading zeros: I'd suggest to implement the example generation in a way so that a check for leading zeros in not needed, e.g. by using the parsed `Integer` value of the input.

Also, shouldn't `PARAM_START_VALUE` be modelled as an `Integer` instead of a `String`?

cst/plugins/eu.esdihumboldt.cst.functions.numeric/src/eu/esdihumboldt/cst/functions/numeric/sequentialid/SequentialID.java
```diff
@@ -88,9 +88,13 @@ protected Object evaluate(String transformationIdentifier, TransformationEngine
 			else {
 				try {
 					id = Integer.parseInt(startValue);
+					if (id == 0) {
```
stempler, line 91: To get some background: Why is it not OK if the starting number is zero?

The condition will ignore values lower than zero - that seems to be not OK though according to the message below?

ui/plugins/eu.esdihumboldt.hale.ui.functions.numeric/src/eu/esdihumboldt/hale/ui/functions/numeric/SequentialIDParameterPage.java
```diff
@@ -298,6 +303,28 @@ public void modifyText(ModifyEvent e) {
 	 */
 	protected void updateStatus() {
 		boolean complete = true;
+		setErrorMessage(null);
+
+		try {
+
+			Integer.parseInt(startValue.getText());
+
+			if (startValue.getText().charAt(0) == '0') {
```
stempler, line 312: This only checks the first character in the string.

This very is different from the check that is used in the function implementation. It would probably make sense to use the same condition in both places so it is consistent?

ui/plugins/eu.esdihumboldt.hale.ui.functions.numeric/src/eu/esdihumboldt/hale/ui/functions/numeric/SequentialIDParameterPage.java
```diff
@@ -383,8 +411,20 @@ protected String generateExample() {
 
 		prefix = variables.replaceVariables(prefix);
 		suffix = variables.replaceVariables(suffix);
-		startValue = variables.replaceVariables(startValue);
 
-		return prefix + startValue + suffix;
+		int ID = 0;
+
+		if (startValue != "") {
```
stempler, line 417: In Java you can't check for an empty string like this, this compares the references, not the contents.

You could use something like
`if (startValue != null && !startValue.isEmpty()) {`

cst/plugins/eu.esdihumboldt.cst.functions.numeric/src/eu/esdihumboldt/cst/functions/numeric/sequentialid/SequentialID.java
Flaminietta: After a few clarifications with Kate, Thorsten and Florian, I've got that ID should start with any integer larger than 0. No initial special characters are allowed. Also, IDs starting for example with 01, 001, 0001...should not be allowed because leading 0s are used only for padding.

thorsten-reitz: Exactly, leading 0es (padding) should not be added, since the primary use case was simple integer IDs. I think it can in principle also start with 0, but I think in GeoPackages I did indeed always start with 1.

ui/plugins/eu.esdihumboldt.hale.ui.functions.numeric/src/eu/esdihumboldt/hale/ui/functions/numeric/SequentialIDParameterPage.java
Flaminietta: I believe that in both places I am checking that the generated ID (and thus the ID example) does not start with 0.

In the function in fact we check if startValue==0:

```
id = Integer.parseInt(startValue);
if (id == 0) {
		throw new TransformationException(
		"ERROR: the starting value for ID creation should be larger than 0");
		}
```

ui/plugins/eu.esdihumboldt.hale.ui.functions.numeric/src/eu/esdihumboldt/hale/ui/functions/numeric/SequentialIDParameterPage.java
Flaminietta: Thank you, this goes directly in my personal collection ;-) https://app.getguru.com/card/cq8pea7i/Some-tricks-for-newbies

cst/plugins/eu.esdihumboldt.cst.functions.numeric/src/eu/esdihumboldt/cst/functions/numeric/sequentialid/SequentialID.java
stempler: So maybe change the condition to `if (id < 0) {` ?

ui/plugins/eu.esdihumboldt.hale.ui.functions.numeric/src/eu/esdihumboldt/hale/ui/functions/numeric/SequentialIDParameterPage.java
stempler: Still the fact remains that the checks are very different, with different results for different input strings (e.g. if someone uses leading zeros in the input string).

cst/plugins/eu.esdihumboldt.cst.functions.numeric/src/eu/esdihumboldt/cst/functions/numeric/sequentialid/SequentialID.java
Flaminietta: Now there is the same condition as in the example generation. I check for any String, whose first character is 0, e.g. 0, 001, 0001 etc. They are all not allowed.

cst/plugins/eu.esdihumboldt.cst.functions.numeric/src/eu/esdihumboldt/cst/functions/numeric/sequentialid/SequentialID.java
```diff
@@ -86,12 +86,15 @@ protected Object evaluate(String transformationIdentifier, TransformationEngine
 				id = seqValue + 1;
 			}
 			else {
+				if (startValue.charAt(0) == '0') {
```
stempler, line 89: If the condition you want to check is related to numbers I would always prefer handling it as a number.
What about negative integers? I guess they still should not be accepted?
I would still see no reason though, why zero could not be a valid start value.

When you convert the string to an integer something like leading zeros does not matter anymore anyhow, because you work with a number representation and not a string anymore.

Flaminietta:
I hope the last commit addresses all the last points discussed:

- ID can be 0
- ID cannot be < 0
- ID cannot start with special chars
- if the starting value has leading 0s, the example shows that they are dropped
- the logic is implemented after parsing the string into an int

stempler (approved):
LGTM, one minor comment

kapil-agnihotri (commented):
LGTM, one comment regarding the commit messages. I hope you will `squash` all these commit messages into one and and will `reword` the header, body, and the footer before merging.

kapil-agnihotri (approved):

PR #961, opened 2022-04-07
fix(geopackage): allow loading Geopackage files w/ non standard file extension

stempler:
Previously, loading a Geopackage would only work if the file has a
`.gpkg` or `.gpkx` extension.

ING-2671

Also: fix(geopackage): queryFilter is a parameter for the reader


florianesser (approved):
LGTM

PR #967, opened 2022-05-05
feat: export data in json format

kapil-agnihotri:
Changes to export data in JSON format from
any kind of input data.

ING-3291


stempler (changes requested):

io/plugins/eu.esdihumboldt.hale.io.json/src/eu/esdihumboldt/hale/io/json/JsonInstanceWriter.java
```diff
@@ -40,9 +46,18 @@ public boolean isCancelable() {
 	protected IOReport execute(ProgressIndicator progress, IOReporter reporter)
 			throws IOProviderConfigurationException, IOException {
 		progress.begin("Generating JSON", ProgressIndicator.UNKNOWN);
+
+		InstanceToJson instanceToJson = new InstanceToJson(true);
```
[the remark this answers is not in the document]
kapil-agnihotri: > Ideally you would make it an optional constructor argument of this class
1. We don't have `optional arguments` in Java. Do you mean that I should create a default construct with no params and initialise the property to create a valid `GeoJson` to `false`? 
- I think the constructor that takes a boolean parameter is already a clean way to handle this. 

> GeoJSONInstanceWriter can just extend it and only needs to set it to true in the constructor.
2. Why should `GeoJSONInstanceWriter extends InstanceToJson` class? If we extend like this then we cannot extend `AbstractInstanceWriter` unless we would like to create another abstract class that will extend either `InstanceToJson` or `AbstractInstanceWriter` and then extending this `AbstractAbstractinstanceWriter` class into `GeoJSONInstanceWriter`. 
- I believe by `extend`, you meant to create a class variable like `InstanceToJson instanceToJson` and initialise this variable in the default constructor of `GeoJSONInstanceWriter` class with default value as `true`. This can also be achieved by the constructor that takes the boolean argument `public InstanceToJson(boolean geoJson) {.... `.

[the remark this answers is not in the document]
stempler: > Ideally you would make it an optional constructor argument of this class

With "this class" I mean `JsonInstanceWriter`.
With "optional constructor argument" I meant basically adding a constructor with a boolean argument to `JsonInstanceWriter`, optional because we need to retain the default constructor.

> Why should GeoJSONInstanceWriter extends InstanceToJson class?

I meant for `GeoJSONInstanceWriter` to extend `JsonInstanceWriter`.

io/plugins/eu.esdihumboldt.hale.io.json/src/eu/esdihumboldt/hale/io/json/writer/InstanceToJson.java
```diff
@@ -0,0 +1,701 @@
...
+
+	/**
+	 * Write an instance to Json.
+	 *
+	 * @param jsonGen the Json generator
+	 * @param instance the instance to write
+	 * @param log the log
+	 * @throws IOException
+	 */
+//	public void writeInstance(JsonGenerator jsonGen, Instance instance, SimpleLog log)
```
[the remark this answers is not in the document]
kapil-agnihotri: Commented code removed

io/plugins/eu.esdihumboldt.hale.io.json/src/eu/esdihumboldt/hale/io/json/JsonInstanceWriter.java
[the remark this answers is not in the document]
kapil-agnihotri: Changes done to:
1. made `useGeoJsonFeatures` as an optional constructor argument defaulting it to `false`
2. extended `GeoJSONInstanceWriter ` with `JsonInstanceWriter`

io/plugins/eu.esdihumboldt.hale.io.json/src/eu/esdihumboldt/hale/io/json/JsonInstanceWriter.java
```diff
@@ -40,9 +46,18 @@ public boolean isCancelable() {
...
+
 		try {
-			JacksonMapper mapper = new JacksonMapper();
-			mapper.streamWriteCollection(getTarget(), getInstances(), reporter);
+			LocatableOutputSupplier<? extends OutputStream> out = getTarget();
+
+			try (BufferedWriter writer = new BufferedWriter(
+					new OutputStreamWriter(out.getOutput(), Charset.forName("UTF-8")))) {
+				String jsonString = instanceToJson.toJsonString(getInstances(), reporter);
```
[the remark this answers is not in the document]
kapil-agnihotri: Now using `writer` implementation similar to that of `toJsonString` and I have not deleted the code from `toJsonString`.

io/plugins/eu.esdihumboldt.hale.io.json/src/eu/esdihumboldt/hale/io/json/JsonInstanceWriter.java
```diff
@@ -15,14 +15,20 @@
...
+import java.nio.charset.Charset;
 
 import eu.esdihumboldt.hale.common.core.io.IOProviderConfigurationException;
 import eu.esdihumboldt.hale.common.core.io.ProgressIndicator;
 import eu.esdihumboldt.hale.common.core.io.report.IOReport;
 import eu.esdihumboldt.hale.common.core.io.report.IOReporter;
 import eu.esdihumboldt.hale.common.core.io.report.impl.IOMessageImpl;
+import eu.esdihumboldt.hale.common.core.io.supplier.LocatableOutputSupplier;
 import eu.esdihumboldt.hale.common.instance.io.impl.AbstractInstanceWriter;
+import eu.esdihumboldt.hale.io.json.writer.InstanceToJson;
```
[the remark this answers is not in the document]
kapil-agnihotri: Changes done:
1. replaced the functionality in `GeoJSONInstanceWriter`
2. removed `JacksonMapper`
3. Deprecated all `GeoJSONConfig*` classes
4. removed  UI configuration page
5. removed definition of obsolete parameter for `GeoJSONInstanceWriter`

io/plugins/eu.esdihumboldt.hale.io.json/src/eu/esdihumboldt/hale/io/json/GeoJSONInstanceWriter.java
```diff
@@ -22,20 +22,32 @@
...
 
+	/**
+	 * @param geoJson use geoJson feature when exporting to geoJson
+	 */
+	public GeoJSONInstanceWriter() {
+		super(true);
+	}
+
 	@Override
 	public boolean isCancelable() {
```
stempler, line 52: We can remove this method, no additional value here compared to the super class.

io/plugins/eu.esdihumboldt.hale.io.json/src/eu/esdihumboldt/hale/io/json/GeoJSONInstanceWriter.java
```diff
@@ -45,20 +57,16 @@ public boolean isCancelable() {
 	protected IOReport execute(ProgressIndicator progress, IOReporter reporter)
```
stempler, line 57: We can remove this method, no additional value here compared to the super class.

io/plugins/eu.esdihumboldt.hale.io.json/src/eu/esdihumboldt/hale/io/json/JsonInstanceWriter.java
```diff
@@ -31,6 +38,29 @@
  */
 public class JsonInstanceWriter extends AbstractInstanceWriter {
 
+	InstanceToJson instanceToJson;
```
stempler, line 41: Make final (and private?)?

io/plugins/eu.esdihumboldt.hale.io.json/src/eu/esdihumboldt/hale/io/json/JsonInstanceWriter.java
```diff
@@ -31,6 +38,29 @@
...
+	boolean useGeoJsonFeatures = false;
```
stempler, line 42: Make final (and private?) and remove initialization (as it is initialized in the constructors)?

io/plugins/eu.esdihumboldt.hale.io.json/src/eu/esdihumboldt/hale/io/json/writer/IgnoreNamespaces.java
```diff
@@ -0,0 +1,83 @@
...
+	 * This method will not add any prefix to the namespace and will ignore the
+	 * call
+	 *
+	 * @param prefixes a map of namespaces to prefixes
+	 * 
+	 * @see eu.esdihumboldt.hale.io.json.writer.NamespaceManager#addPrefixes(java.util.Map)
+	 */
+	@Override
+	public void addPrefixes(Map<String, String> prefixes) {
+		prefixes.forEach((namespace, prefix) -> {
```
stempler, line 78: No need to loop if we don't do anything, right?

io/plugins/eu.esdihumboldt.hale.io.json/src/eu/esdihumboldt/hale/io/json/writer/NamespaceManager.java
```diff
@@ -0,0 +1,100 @@
...
+import com.google.common.collect.HashBiMap;
+
+/**
+ * Abstract class to manage Namespaces. This class can be further extended to
+ * manage and include namespace prefixes in the exported Json or geoJson
+ * formats. Not Thread safe.
+ * 
+ * @author Kapil Agnihotri
+ */
+public abstract class NamespaceManager {
```
stempler, line 32: I think a it better would have been to introduce an interface that only holds the methods that `InstanceToJson` actually uses. That would have made the `IgnoreNamespaces` implementation simpler and it would not need to carry the unused map around.

But the solution as you implemented is also totally fine :+1:

kapil-agnihotri: First, I made it an interface but then thought of making it abstract 
1. to move the common code like (`getNamespaces()` and `getNamespace()`) to one place. 
2. I had an understanding that `IgnoreNamespaces` would anyways will need to return a map for `getnamespace*()`. Otherwise returning null would then lead to unchecked NPEs and a bit changes in the code.

stempler: Makes sense. Then I did not look at the details closely enough. Thanks!

stempler: Thinking again about it, the sole methods used seem to be `getPrefix` and `getNamespaces`. `getNamespaces` is used to add the namespace definitions to the file in case of GeoJson.
But if we ignore the namespaces, it does not make sense to define any. So `IgnoreNamespaces` could just return an empty map or `null` for `getNamespaces`, and no namespace definitions should be written to the output in that case.

io/plugins/eu.esdihumboldt.hale.io.json/src/eu/esdihumboldt/hale/io/json/GeoJSONInstanceWriter.java
kapil-agnihotri: True, now writing the `TypeName` in the progress message using `getDefaultTypeName`.

io/plugins/eu.esdihumboldt.hale.io.json/src/eu/esdihumboldt/hale/io/json/JsonInstanceWriter.java
kapil-agnihotri: Yes, missed writing it. Sorry.

io/plugins/eu.esdihumboldt.hale.io.json/src/eu/esdihumboldt/hale/io/json/JsonInstanceWriter.java
kapil-agnihotri: Yes as the default constructor will call the parameterised one so no need to initialise it. initialisation removed.

io/plugins/eu.esdihumboldt.hale.io.json/src/eu/esdihumboldt/hale/io/json/writer/IgnoreNamespaces.java
kapil-agnihotri: Doing nothing with the new implementation.

io/plugins/eu.esdihumboldt.hale.io.json/src/eu/esdihumboldt/hale/io/json/writer/NamespaceManager.java
kapil-agnihotri: True, We don't need to return any name space definition if we are ignoring namespace. I created an interface and now returning either empty string for prefix or `null` (I guess returning an empty map is unnecessary just to avoid a null check it will also consume some memory). Also, now I added a null check when writing the namespaces incase of GeoJson.

PR ready for review.

stempler (approved):

PR #968, opened 2022-05-10
refactor: merge schemas for HSD export

stempler:
When exporting multiple schemas to HSD as XML or Json, merge any schemas
to one schema.
This works around the restriction, that only a single schema can be
imported from an HSD file.

### DIFF
--- a/common/plugins/eu.esdihumboldt.hale.common.schema.persist/src/eu/esdihumboldt/hale/common/schema/persist/hsd/MergeSchemas.java
+++ b/common/plugins/eu.esdihumboldt.hale.common.schema.persist/src/eu/esdihumboldt/hale/common/schema/persist/hsd/MergeSchemas.java
@@ -1,0 +1,100 @@
+/*
+ * Copyright (c) 2022 wetransform GmbH
+ * 
+ * All rights reserved. This program and the accompanying materials are made
+ * available under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation, either version 3 of the License,
+ * or (at your option) any later version.
+ * 
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this distribution. If not, see <http://www.gnu.org/licenses/>.
+ * 
+ * Contributors:
+ *     wetransform GmbH <http://www.wetransform.to>
+ */
+
+package eu.esdihumboldt.hale.common.schema.persist.hsd;
+
+import java.net.URI;
+import java.util.Collections;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+import com.google.common.collect.Streams;
+
+import eu.esdihumboldt.hale.common.schema.model.Schema;
+import eu.esdihumboldt.hale.common.schema.model.impl.DefaultSchemaSpace;
+
+/**
+ * Helper class for merging multiple schemas for export.
+ * 
+ * @author Simon Templer
+ */
+public class MergeSchemas {
+
+	private static class MergedSchema extends DefaultSchemaSpace implements Schema {
+
+		private final String namespace;
+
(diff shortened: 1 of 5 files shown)


common/plugins/eu.esdihumboldt.hale.common.schema.persist/src/eu/esdihumboldt/hale/common/schema/persist/hsd/MergeSchemas.java
```diff
@@ -0,0 +1,100 @@
...
+			return this.namespace;
+		}
+
+	}
+
+	/**
+	 * Namespace to use for combined schemas if no common namespace is
+	 * available.
+	 */
+	public static final String DEFAULT_COMBINED_NAMESPACE = HaleSchemaConstants.NS + "/combine";
```
kapil-agnihotri, line 62: What happens if the combined schema is then again merged with some other schema? For instance importing it again in HS with some other schema
- Is this scenario possible? 
- Will this lead to `"http://www.esdi-humboldt.eu/hale/schema/combined/combined" namespace?

common/plugins/eu.esdihumboldt.hale.common.schema.persist/src/eu/esdihumboldt/hale/common/schema/persist/hsd/MergeSchemas.java
```diff
@@ -0,0 +1,100 @@
...
+
+		if (mergeAll) {
+			// we merge to one namespace
+			String namespace;
+			if (counts.size() == 1) {
+				// if there is only one namespace, use it
+				namespace = counts.keySet().iterator().next();
+			}
+			else {
+				namespace = DEFAULT_COMBINED_NAMESPACE;
```
kapil-agnihotri, line 87: `DEFAULT_COMBINED_NAMESPACE` is used only in this class. Can it be made private?

kapil-agnihotri (approved):

common/plugins/eu.esdihumboldt.hale.common.schema.persist/src/eu/esdihumboldt/hale/common/schema/persist/hsd/MergeSchemas.java
```diff
@@ -0,0 +1,100 @@
...
+ * 
+ * @author Simon Templer
+ */
+public class MergeSchemas {
+
+	private static class MergedSchema extends DefaultSchemaSpace implements Schema {
+
+		private final String namespace;
+
+		public MergedSchema(String namespace, Iterable<? extends Schema> schemas) {
```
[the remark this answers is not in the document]
stempler: For HSD the issue with `Schema` is only that it has, in addition to the types that are defined, a `namespace` property.
The original behavior was to retain the schemas as they are loaded in hale, but the issue there is that it is not supported to load multiple schemas with the existing mechanism.

In HSD the schema namespace does not play a role, other than that it is persisted. It is a different matter when for instance trying to create an XML schema from the schema, but that is a different problem then, that needs to be solved somewhere else.

common/plugins/eu.esdihumboldt.hale.common.schema.persist/src/eu/esdihumboldt/hale/common/schema/persist/hsd/MergeSchemas.java
stempler: Good point. For the HSD export it would not matter though, because the namespaces of the schemas that are combined are not retained other than their usage in the defined types. And that is where the combined namespace is not used. So I think we are safe with the current approach.

PR #969, opened 2022-05-11
NPE exporting multiple geometries to Shapefiles

kapil-agnihotri:


io/plugins/eu.esdihumboldt.hale.io.shp/src/eu/esdihumboldt/hale/io/shp/writer/ShapefileInstanceWriter.java
```diff
@@ -543,18 +543,22 @@ private Map<String, Map<String, List<SimpleFeature>>> createFeatures(
 					writePropertiesInstanceData(schemaFbMap, instance, type, localPart);
 
 					// create list of simple features.
-					Set<String> geometryKeys = schemaFbMap.get(localPart).keySet();
-					for (String key : geometryKeys) {
+					// fix in case of multiple geometries in data
```
stempler, line 546: @kapil-agnihotri Was the issue here really multiple geometries (which is not supported in Shapefiles, right?) or rather different geometry types?
As @JohannaOtt 's data was from Shapefiles originally, I would not thing that it is possible that it contained multiple geometries per feature type.

Adding a unit test with a respective test case would also help understanding the issue better.

kapil-agnihotri: > Was the issue here really multiple geometries (which is not supported in Shapefiles, right?) or rather different geometry types?

Yes, it was the issue with the different geometry types. We already supported multiple geometries in a single schema but this case was little different where each instance has different geometry type but single geometry. I have added a specific test case too for this. 

I was referring this case with multiple geometries too but it should be called `with different geometry types` because every instance in the data contains only one geometry object.

stempler (approved):

PR #971, opened 2022-05-18
fixup! feat: export data in json format

kapil-agnihotri:
Correct the winding order as counterclockwise based
on the right-hand rule, i.e., exterior rings are  counterclockwise,
and holes are clockwise.

ING-3291


kapil-agnihotri:
Hi @stempler, please let me know your feedback.

PR #976, opened 2022-07-08
Remove the GroupPropertyDefinitions from path when using "Add condition context"

Flaminietta:
The function "Add condition context" does not work if in the filter path are included also GroupPropertyDefinitions (e.g. "choice", "choice_1" or the like). This PR filters them out.

ING-3458


Flaminietta:
I have addressed your suggestions.

ui/plugins/eu.esdihumboldt.hale.ui.filter/src/eu/esdihumboldt/hale/ui/filter/cql/CQLFilterField.java
```diff
@@ -95,11 +96,18 @@ protected String selectVariable() {
...
 			for (int i = 0; i < dialog.getObject().getPropertyPath().size(); i++) {
-				if (i != 0)
-					var.append(".");
-				var.append(dialog.getObject().getPropertyPath().get(i).getChild().getName()
-						.getLocalPart().toString());
+				ChildContext child = dialog.getObject().getPropertyPath().get(i);
+
+				if (child.getChild().asGroup() == null) {
+					if (i != 0)
+						var.append(".");
```
stempler, line 103: > The case that you don't handle here however is that the first element (index 0) could have been a group/choice. In that case you would also not want to add a dot.

This is not solved yet.
If the  first element with index 0 was a group/choice then you would start the string with a dot, which is not correct.

stempler (changes requested):

ui/plugins/eu.esdihumboldt.hale.ui.filter/src/eu/esdihumboldt/hale/ui/filter/cql/CQLFilterField.java
```diff
@@ -92,23 +92,26 @@ protected String selectVariable() {
...
+		String stringVar = null;
 		if (dialog.open() == PropertyDefinitionDialog.OK && dialog.getObject() != null
 				&& dialog.getObject().getType().getName().toString().length() >= 1) {
 			StringBuilder var = new StringBuilder();
 			for (int i = 0; i < dialog.getObject().getPropertyPath().size(); i++) {
 				ChildContext child = dialog.getObject().getPropertyPath().get(i);
 
 				if (child.getChild().asGroup() == null) {
-					if (i != 0)
+					if (i != 0 && !stringVar.isEmpty())
```
stempler, line 103: `stringVar` will not be initialized yet.
You can do a check on the StringBuilder (`var`) directly. There should be a method where you can check its current length or something like that.

stempler (approved):

PR #980, opened 2022-08-01
Groovy Merge UI: one single variable or a list is proposed

Flaminietta:
When using the help wizard during a Groovy Merge to specify a Groovy script (to build a target instance from a merged source), the access variable from the source structure automatically proposes to access one single value or all values in a list.

ING-3488

### DIFF
--- a/ui/plugins/eu.esdihumboldt.hale.ui.functions.groovy/src/eu/esdihumboldt/hale/ui/functions/groovy/internal/JoinTypeStructureTray.java
+++ b/ui/plugins/eu.esdihumboldt.hale.ui.functions.groovy/src/eu/esdihumboldt/hale/ui/functions/groovy/internal/JoinTypeStructureTray.java
@@ -40,7 +40,6 @@ import eu.esdihumboldt.hale.ui.util.viewer.tree.TreePathProviderAdapter;
  */
 public class JoinTypeStructureTray extends TypeStructureTray {
 
-	private static boolean isMerge = false;
 	private ParameterValue param = null;
 
 	/**
@@ -53,7 +52,7 @@ public class JoinTypeStructureTray extends TypeStructureTray {
 	 */
 	public JoinTypeStructureTray(ParameterValue param, TypeProvider types,
 			SchemaSpaceID schemaSpace) {
-		super(types, schemaSpace, isMerge);
+		super(types, schemaSpace, false);
 		this.param = param;
 
 	}
--- a/ui/plugins/eu.esdihumboldt.hale.ui.functions.groovy/src/eu/esdihumboldt/hale/ui/functions/groovy/internal/JoinTypeStructureTray.java
+++ b/ui/plugins/eu.esdihumboldt.hale.ui.functions.groovy/src/eu/esdihumboldt/hale/ui/functions/groovy/internal/JoinTypeStructureTray.java
@@ -40,6 +40,7 @@ import eu.esdihumboldt.hale.ui.util.viewer.tree.TreePathProviderAdapter;
  */
 public class JoinTypeStructureTray extends TypeStructureTray {
 
+	private static boolean isMerge = false;
 	private ParameterValue param = null;
 
 	/**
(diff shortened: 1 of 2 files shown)


Flaminietta:
I am not sure whether my solution is general enough. 
As discussed, I have found that I obtain the desired outcome if `canOccurMultipleTimes = true` in TypeStructureTray.java. Now `canOccurMultipleTimes = true` in case of a Merge and/or a Retype.
Doubts:
- is that really so, that `canOccurMultipleTimes = true` for EVERY Merge/Retype?
-  why  does`if !(VARIABLES_TYPE_NAME.equals(type.getName()))` determine if one is using Merge/Retype?

ui/plugins/eu.esdihumboldt.hale.ui.functions.groovy/src/eu/esdihumboldt/hale/ui/functions/groovy/internal/TypeStructureTray.java
```diff
@@ -393,14 +396,15 @@ private String createSourceSample(TreePath path, Collection<? extends TypeDefini
 			}
 			else {
 				// assuming Retype/Merge
```
stempler, line 398: _Explanation:_
This comment is related to that previously if was identified that this (a) is not a join and (b) this is not a property transformation.
It can't be a Groovy Create either because that has no source.

ui/plugins/eu.esdihumboldt.hale.ui.functions.groovy/src/eu/esdihumboldt/hale/ui/functions/groovy/internal/TypeStructureTray.java
```diff
@@ -475,7 +479,7 @@ else if (def instanceof TypeDefinition) {
 			StringBuilder example = new StringBuilder();
 
 			boolean canOccurMultipleTimes = false;
-			if (path.getFirstSegment() instanceof TypeDefinition) {
+			if (path.getFirstSegment() instanceof TypeDefinition || RetypeOrMerge == true) {
```
stempler, line 482: We really only want this if it is a Merge, not a Retype.

ui/plugins/eu.esdihumboldt.hale.ui.functions.groovy/src/eu/esdihumboldt/hale/ui/functions/groovy/internal/TypeStructureTray.java
```diff
@@ -109,6 +110,8 @@ public interface TypeProvider {
 
 	}
 
+	static boolean isMerge = false;
```
stempler, line 113: Static is a bad idea. Like this, once you had one case with a GroovyMergePage this would always be true, even when you edit a Groovy Retype etc.

You could instead pass the variable to the `TypeStructureTray` constructor.

stempler (changes requested):

ui/plugins/eu.esdihumboldt.hale.ui.functions.groovy/src/eu/esdihumboldt/hale/ui/functions/groovy/internal/JoinTypeStructureTray.java
```diff
@@ -40,6 +40,7 @@
  */
 public class JoinTypeStructureTray extends TypeStructureTray {
 
+	private static boolean isMerge = false;
```
stempler, line 43: In the join type structure tray there is no variable needed, we know that because it is for the Join, it is not a Merge.

ui/plugins/eu.esdihumboldt.hale.ui.functions.groovy/src/eu/esdihumboldt/hale/ui/functions/groovy/internal/JoinTypeStructureTray.java
```diff
@@ -50,8 +51,9 @@ public class JoinTypeStructureTray extends TypeStructureTray {
 	 * @param types the type provider
 	 * @param schemaSpace the schema space
 	 */
-	public JoinTypeStructureTray(ParameterValue param, TypeProvider types, SchemaSpaceID schemaSpace) {
-		super(types, schemaSpace);
+	public JoinTypeStructureTray(ParameterValue param, TypeProvider types,
+			SchemaSpaceID schemaSpace) {
+		super(types, schemaSpace, isMerge);
```
stempler, line 56: Thus we can just pass `false` here.

stempler (approved):

kapil-agnihotri (approved):

Issue #970, opened 2022-05-16
Add option to Retype function with structural name to exclude top-level properties from copying


thorsten-reitz:
Currently, it is not possible to make a Retype with structural rename enabled, when one needs to also apply property level functions for  specific behaviour.

With this adaptation to the retype function, one additional parameter would be added to the Function parameters page that allows the user to select properties which should be excluded from the copying.

JohannaOtt:
Closed with https://github.com/halestudio/hale/pull/981

Issue #909, opened 2021-08-17
Add option to overwrite structural Retype mapping for single elements


JohannaOtt:
Currently, when using a structural Retype mapping, there is no option to change single elements.
Sometimes, there are use cases, where you want to keep the biggest part of the source data, but only want to change a few elements.
In these cases it would be helpful to be able to "overwrite" the mapping resulting from the structural Retype with an explicite mapping on property level.
So there should be something like an internal prioritization between the additional mapping on attribute level (higher priority) and the property "mappings" (not explicitely created) from the structural Retype (lower priority).

thorsten-reitz:
I agree with this, it would be quite powerful to have such a feature.

thorsten-reitz:
I have actually tested this in the last weeks, and have only encountered one case where it doesn't work: When there is a type level groovy operation (Groovy Merge, Groovy Retype). In all other cases (check the end repo for examples) this worked nicely.

thorsten-reitz:
Related way of solving this: https://github.com/halestudio/hale/issues/970

JohannaOtt:
Closed with https://github.com/halestudio/hale/pull/981

PR #981, opened 2022-08-16
feat: allow excluding root properties from structural Retype

stempler:
When using Retype with enabled structural rename you can mark
properties to exclude, that will not be copied.

Limitations:

- the UI page for configuring the properties to exclude shows up
  regardless if structural rename is enabled or not
- as per design only root properties at the first level in the source
  type can be excluded


thorsten-reitz (approved):

JohannaOtt:
@stempler Would that solve https://github.com/halestudio/hale/issues/909 and https://github.com/halestudio/hale/issues/970?

stempler:
@JohannaOtt I guess so

PR #995, opened 2022-10-21
refactor: move topojson support to separate bundle

stempler:
To reduce dependencies of JSON bundle and only include Topojson dependencies if necessary.
Avoids the need to a Jitpack as Maven repository if hale is used with Maven dependencies but without Topojson.

ING-3131


florianesser (approved):
LGTM

PR #1001, opened 2022-11-14
build: Use --add-exports= to prevent malformed HALE.ini

florianesser:
With the previous `--add-exports`, the build process would generate the following segment in HALE.ini:

```
--add-exports
java.base/sun.nio.ch=ALL-UNNAMED
java.base/jdk.internal.ref=ALL-UNNAMED
```

This leads to hale»studio being unstartable at least on Linux where it gives the following exception:

```
$ ./HALE
Error: Could not find or load main class java.base.jdk.internal.ref=ALL-UNNAMED
Caused by: java.lang.ClassNotFoundException: java.base.jdk.internal.ref=ALL-UNNAMED
```

With the `--add-exports=` notation, the following segment in HALE.ini is generated instead which allows hale»studio to start correctly:

```
--add-exports=java.base/sun.nio.ch=ALL-UNNAMED
--add-exports=java.base/jdk.internal.ref=ALL-UNNAMED
```

ING-3267

### DIFF
--- a/build/templates/products/client.yaml
+++ b/build/templates/products/client.yaml
@@ -17,8 +17,8 @@ HALE:
   - -Xms256M
   - -XX:MinHeapFreeRatio=40
   - -XX:MaxHeapFreeRatio=55
-  - -add-exports java.base/sun.nio.ch=ALL-UNNAMED 
-  - -add-exports java.base/jdk.internal.ref=ALL-UNNAMED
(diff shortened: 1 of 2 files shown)


kapil-agnihotri (approved):
Thanks a lot @florianesser.

stempler (approved):

PR #1003, opened 2022-11-21
build: set multiModuleProjectDirectory for updateSite

kapil-agnihotri:
ING-3267


build/gradle/updateSite.gradle
```diff
@@ -159,7 +159,7 @@ task buildUpdateSite(dependsOn: [ createUpdateSiteFeature, generatePomFiles ]) {
 			]
 		}
 	    pg.generateParentPomFile(parentPomConfig)
-		
+		System.setProperty( MavenCli.MULTIMODULE_PROJECT_DIRECTORY, projectRoot);
```
stempler, line 162: Indentation seems off

stempler (approved):

build/gradle/updateSite.gradle
kapil-agnihotri: No idea why is it like this here, in my workspace this is exactly below `pg.` 

```
pg.generateParentPomFile(parentPomConfig) -> Line:column (161:9)
System.setProperty()                      -> Line:column (162:9)
```

kapil-agnihotri: I tried to add a blank like but still it is the same here in GitHub. However, in GitHub desktop it looks properly indented. 
Screenshot 2022-11-21 at 15 09 17

stempler: Probably inconsistent use of indentation in the file then. Then leave it as it is I would say.

PR #1006, opened 2022-12-09
fix: upload project to hale connect

kapil-agnihotri:
Compromise on immutability for OwnerFilterEntry after java 17

ING-3701

### DIFF
--- a/io/plugins/eu.esdihumboldt.hale.io.haleconnect.ui/src/eu/esdihumboldt/hale/io/haleconnect/ui/projects/ChooseHaleConnectProjectWizardPage.groovy
+++ b/io/plugins/eu.esdihumboldt.hale.io.haleconnect.ui/src/eu/esdihumboldt/hale/io/haleconnect/ui/projects/ChooseHaleConnectProjectWizardPage.groovy
@@ -73,7 +73,7 @@ import groovy.transform.CompileStatic
 public class ChooseHaleConnectProjectWizardPage extends ConfigurationWizardPage<HaleConnectProjectConfig> {
 
 	private static final ALogger log = ALoggerFactory.getLogger(ChooseHaleConnectProjectWizardPage.class);
-	private static final OwnerFilterEntry NULL_FILTER = new OwnerFilterEntry(owner: null as Owner[], label: "All projects");
+	private static final OwnerFilterEntry NULL_FILTER = new OwnerFilterEntry(owner: null , label: "All projects");
 	private static final List<String> LOADING_TABLE_INPUT = ["Loading..."]
 
 	private final HaleConnectService haleConnect;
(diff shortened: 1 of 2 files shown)


io/plugins/eu.esdihumboldt.hale.io.haleconnect.ui/src/eu/esdihumboldt/hale/io/haleconnect/ui/projects/OwnerFilterEntry.groovy
```diff
@@ -16,14 +16,17 @@
...
  * Entry in the owner filter dropdown of LoadHaleConnectProjectWizardPage
+ *
+ * After migration to java17, if this class is marked as @Immutable then 
+ * at runtime an exception is thrown indicating that clone() is not applicable to Owner class.
+ * Therefore, using @Canonical which provides similar functionality but immutability. 
  * 
  * @author Florian Esser
  */
-@Immutable
+@Canonical
```
kapil-agnihotri, line 29: Can we compromise on immutability here?
https://docs.groovy-lang.org/2.4.7/html/gapi/groovy/transform/Immutable.html

kapil-agnihotri:
retest this please

kapil-agnihotri:
@stempler PR ready for review.

io/plugins/eu.esdihumboldt.hale.io.haleconnect.ui/src/eu/esdihumboldt/hale/io/haleconnect/ui/projects/OwnerFilterEntry.groovy
stempler: Based on the documentation I would guess that since `clone` is used for arrays this causes a problem, since `Owner` does not implement it.

I would see two options:

- let `Owner` implement `Cloneable` (e.g. via [@AutoClone](https://docs.groovy-lang.org/next/html/gapi/groovy/transform/AutoClone.html))
- use a collection type (e.g. `List`) instead of an array for `owner`

I would prefer using `@AutoClone` if that works.

stempler (commented):
I think it is OK to remove the immutability here, though I think this is a bug in Groovy, since using

io/plugins/eu.esdihumboldt.hale.io.haleconnect.ui/src/eu/esdihumboldt/hale/io/haleconnect/ui/projects/ChooseHaleConnectProjectWizardPage.groovy
```diff
@@ -73,7 +73,7 @@ import groovy.transform.CompileStatic
 public class ChooseHaleConnectProjectWizardPage extends ConfigurationWizardPage<HaleConnectProjectConfig> {
 
 	private static final ALogger log = ALoggerFactory.getLogger(ChooseHaleConnectProjectWizardPage.class);
-	private static final OwnerFilterEntry NULL_FILTER = new OwnerFilterEntry(owner: null as Owner[], label: "All projects");
+	private static final OwnerFilterEntry NULL_FILTER = new OwnerFilterEntry(owner: Arrays.asList(null) , label: "All projects");
```
stempler, line 76: I think this is different than the original. The original used `null` for the array itself, you create a list here with a `null` inside.

stempler (changes requested):
LGTM except for `NULL_FILTER`

io/plugins/eu.esdihumboldt.hale.io.haleconnect.ui/src/eu/esdihumboldt/hale/io/haleconnect/ui/projects/ChooseHaleConnectProjectWizardPage.groovy
kapil-agnihotri: True, updated the code.

stempler (approved):

PR #1012, opened 2023-01-03
build: include Nashorn JavaScript engine for MSI build

florianesser:
Adds the Nashorn JavaScript engine and its dependencies. The engine is needed to run the `<script>` Ant tasks that are part of the Windows MSI build. Nashorn was previously part of the JDK but was removed with the release of Java 15.

ING-3370


kapil-agnihotri (approved):

stempler (approved):
LGTM

PR #1029, opened 2023-06-07
feat: download wfs files to tmp folder for bucket service

kapil-agnihotri:
ING-3609


io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/reader/internal/wfs/WfsBackedGmlInstanceCollection.java
```diff
@@ -406,12 +448,27 @@ private void proceedOrClose() {
...
+				builder = new URIBuilder(nextUri);
 			} catch (URISyntaxException e) {
 				throw new IllegalStateException(e.getMessage(), e);
 			}
 
+			// Download all the files into a folder for Bucket service
+			if (tmpdir != null) {
+				try {
+					URL url = nextUri.toURL();
+					String fileName = tmpdir + url.getPath() + UUID.randomUUID();
```
stempler, line 463: Can the path always be used as file name? I'm sure it for instance may contain `/` and other characters not allowed in a file name.

Related to the UUID - this will mean the file name will be different every time. Did you already think about how to deal with that on the bucket-service side related to if there are already existing files (e.g. avoid duplicate instances being loaded if a dataset is updated automatically from a WFS)

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/reader/internal/wfs/WfsBackedGmlInstanceCollection.java
```diff
@@ -406,12 +448,27 @@ private void proceedOrClose() {
...
+					// TODO: may be download files in a thread
+					FileUtils.copyURLToFile(url, new File(fileName));
```
stempler, line 465: To avoid downloading the same data twice I think it would make sense to overwrite `nextUri` with the URI to the downloaded file.

io/plugins/eu.esdihumboldt.hale.io.gml.ui/src/eu/esdihumboldt/hale/io/gml/ui/GmlReaderSettingsPage.java
```diff
@@ -70,6 +70,7 @@ public boolean updateConfiguration(StreamGmlReader provider) {
 				Value.of(requestPaginationEnabled.getSelection()));
 		provider.setParameter(StreamGmlReader.PARAM_FEATURES_PER_WFS_REQUEST,
 				Value.of(featuresPerRequest.getText()));
+		provider.setParameter(StreamGmlReader.PARAM_TMP_DIR_PATH, null);
```
stempler, line 73: Why is this necessary? It should be an optional parameter, right?

kapil-agnihotri: Yes, agreed.

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/reader/internal/wfs/WfsBackedGmlInstanceCollection.java
kapil-agnihotri: In bucket service, I see another problem of how to send these multiple files into a single file as expected [here](https://github.com/wetransform/bucket-service/blob/4ec4b7190eeb472d1cddb95b1c0f43ef0dbda7c1/play/app/wfs/WfsImporter.scala#L160) . 
I cannot write all these different files into a single file. Should I return these several files as a list instead? However, in the end it is still expected as a [seq of Futures of single downloaded file](https://github.com/wetransform/bucket-service/blob/4ec4b7190eeb472d1cddb95b1c0f43ef0dbda7c1/play/app/controllers/Buckets.scala#L1072). If I need to return several files for a single wfs url then I should modify [this code](https://github.com/wetransform/bucket-service/blob/4ec4b7190eeb472d1cddb95b1c0f43ef0dbda7c1/play/app/controllers/Buckets.scala#L1072) as well to handle the scenario where a single wfsEndpoint can have more than one downloaded file. 

Or I was thinking that I can look at the GmlInstanceWriter and StreamGmlWriter to see how it handles the case of pagination and features per request and then replicate the same writing to the files?

Please let me know if you have any further suggestions.

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/reader/internal/wfs/WfsBackedGmlInstanceCollection.java
```diff
@@ -412,6 +454,24 @@ private void createNextIterator() {
 				throw new IllegalStateException(e.getMessage(), e);
 			}
 
+			// Download all the files into a folder for Bucket service
+			if (tmpdir != null) {
+				try {
+					URL url = nextUri.toURL();
+					StringBuilder fileName = new StringBuilder(tmpdir);
+					fileName.append("/").append(url.getPath().replaceAll("/", "."))
```
stempler, line 462: In principle OK for me, important is that it works fine with whatever strategy you use for handling it on the bucket-service side.

Did you check how the name was derived before when downloading via WFS in the bucket-service?

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/reader/internal/wfs/WfsBackedGmlInstanceCollection.java
```diff
@@ -412,6 +454,24 @@ private void createNextIterator() {
...
+							.append(UUID.randomUUID());
```
stempler, line 463: I think a sequential numbering would be better than a random suffix. Then a user could identify, which was the first part, second part, etc.

Also, I think the file extension could matter, especially for the handling in the bucket-service later.

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/reader/internal/wfs/WfsBackedGmlInstanceCollection.java
kapil-agnihotri: The file was always provided from [bucket-service](https://github.com/wetransform/bucket-service/blob/baf3eddb03133661f078ccf9071d40dcc4b00e90/play/app/wfs/WfsImporter.scala#LL144C16-L144C16) to hale studio as `FileIOSupplier`. Which is [derived here](https://github.com/wetransform/bucket-service/blob/baf3eddb03133661f078ccf9071d40dcc4b00e90/play/app/controllers/Buckets.scala#LL1102C12-L1102C12) from queryMap and then reduced in the length.

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/reader/internal/wfs/WfsBackedGmlInstanceCollection.java
kapil-agnihotri: Done, now appending `totalFeaturesProcessed` and `.gml` as extension.

stempler (approved):

PR #1033, opened 2023-06-29
feat/ING-3911

emanuelaepure10:
A new format .cpg has been added to the creatin of the .shp files.
The new files contains just the encoding text, such as “UTF-8”.

ING-3911


io/plugins/eu.esdihumboldt.hale.io.shp/src/eu/esdihumboldt/hale/io/shp/writer/ShapefileInstanceWriter.java
```diff
@@ -102,31 +102,44 @@ public boolean isCancelable() {
...
-			String filePath = Paths.get(location).getParent().toString();
-
-			filesWritten = writeInstances(instances, progress, reporter, location);
-
-			if (filesWritten.size() > 1) {
-				List<URI> uris = filesWritten.stream().map(f -> {
-					File file = new File(filePath + "/" + f + ShapefileConstants.SHP_EXTENSION);
-					return file.toURI();
-				}).collect(Collectors.toList());
+		/**
```
stempler, line 107: This looks strange - two new methods defined in another method? Does that compile for you?
Also, the removed code looks quite important to the functionality to me.

I see the there is an existing class for testing writing Shapefiles (see [here](https://github.com/halestudio/hale/blob/master/io/plugins/eu.esdihumboldt.hale.io.shp.test/src/eu/esdihumboldt/hale/io/shp/ShapefileInstanceWriterTest.groovy)). It would be great if you could extend the test cases to verify that the CPG file is written and contains the expected content.

emanuelaepure10: This is compiling and it's working.

The second new method is called from 2 places: once when you have more than one source and once when you have just one source data/feature type. 
There was no code removed. Please check bellow and you will see that you can still find it. The for has been changes to accommodate the creation of .cpg as well. 

I will extend the test to verify the .cpg

stempler: > This is compiling and it's working.

Well, I reviewed the version before you added the additional commit, so I could only take into account what I could see here.

It would be great if you can combine the two commits, since the first one seems to have been corrupt in some way then.

io/plugins/eu.esdihumboldt.hale.io.shp/src/eu/esdihumboldt/hale/io/shp/writer/ShapefileInstanceWriter.java
```diff
@@ -112,10 +116,18 @@ protected IOReport execute(ProgressIndicator progress, IOReporter reporter)
...
+				for (String f : filesWritten) {
 					File file = new File(filePath + "/" + f + ShapefileConstants.SHP_EXTENSION);
-					return file.toURI();
-				}).collect(Collectors.toList());
+					uris.add(file.toURI());
+
+					String cpgFileName = filePath + "/" + f + ShapefileConstants.CPG_EXTENSION;
+					writeCodePageFile(cpgFileName);
+					file = new File(cpgFileName);
+					uris.add(file.toURI());
```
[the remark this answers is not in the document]
kapil-agnihotri: This will contain [name of the file](https://github.com/halestudio/hale/blob/a970f6e19cac38578fb00eb75baa31808df22423/io/plugins/eu.esdihumboldt.hale.io.shp/src/eu/esdihumboldt/hale/io/shp/writer/ShapefileInstanceWriter.java#L693C8-L693C8) without extension.

io/plugins/eu.esdihumboldt.hale.io.shp.test/src/eu/esdihumboldt/hale/io/shp/ShapefileInstanceWriterTest.groovy
```diff
@@ -1443,4 +1456,23 @@ class ShapefileInstanceWriterTest {
 			assertEquals(2, num)
 		}
 	}
+
+	@Test
+	void testCPGFileCreation() {
```
kapil-agnihotri, line 1461: This test, tests the functionality of writing just a standalone file. Would it be possible to create another test where you can confirm creation of `cpg` file in the complete process?
You can refer to [this test case](https://github.com/halestudio/hale/blob/a970f6e19cac38578fb00eb75baa31808df22423/io/plugins/eu.esdihumboldt.hale.io.shp.test/src/eu/esdihumboldt/hale/io/shp/ShapefileInstanceWriterTest.groovy#L208) and may be replicate it and when you have [reference to file](https://github.com/halestudio/hale/blob/a970f6e19cac38578fb00eb75baa31808df22423/io/plugins/eu.esdihumboldt.hale.io.shp.test/src/eu/esdihumboldt/hale/io/shp/ShapefileInstanceWriterTest.groovy#L232C40-L232C40) probably you can there also assert for the creation of `cpg` file.

io/plugins/eu.esdihumboldt.hale.io.shp/src/eu/esdihumboldt/hale/io/shp/writer/ShapefileInstanceWriter.java
```diff
@@ -472,6 +478,8 @@ private Map<String, Map<String, ShapefileDataStore>> createSchema(URI location,
 	 * - filename_geometryType.shp if multiple geometries.<br>
 	 * - filename.shp single schema and geom.
 	 * 
+	 * Create the CPG file starting from the Shapefile names
```
kapil-agnihotri, line 481: Missed moving this comment to the new method `writeCodePageFile` during refactoring?

io/plugins/eu.esdihumboldt.hale.io.shp/src/eu/esdihumboldt/hale/io/shp/writer/ShapefileInstanceWriter.java
```diff
@@ -724,4 +732,30 @@ private List<String> writeToFile(
...
 	}
 
+	/**
+	 * 
+	 * @param filePath Path of the file to be written with just one line of the
+	 *            encoding
+	 * @return File that has been created filled with the text
+	 * @throws IOException exception in any.
+	 */
+	public File writeCodePageFile(String filePath) throws IOException {
```
kapil-agnihotri, line 742: Maybe you don't have to return the File as it is not being used anywhere?

io/plugins/eu.esdihumboldt.hale.io.shp/src/eu/esdihumboldt/hale/io/shp/writer/ShapefileInstanceWriter.java
```diff
@@ -724,4 +732,30 @@ private List<String> writeToFile(
...
+		File file = new File(filePath);
+		FileWriter fr = new FileWriter(file);
```
kapil-agnihotri, line 744: `fw` for fileWriter instead of `fr` which may stand for fileReader?

io/plugins/eu.esdihumboldt.hale.io.shp/src/eu/esdihumboldt/hale/io/shp/writer/ShapefileInstanceWriter.java
```diff
@@ -724,4 +732,30 @@ private List<String> writeToFile(
...
+		try {
+			fr.write(getCharset() != null ? getCharset().toString()
+					: getDefaultCharset().toString());
+		} catch (IOException e) {
+			throw e;
+		} finally {
+			try {
+				fr.close();
+			} catch (IOException e) {
+				throw new IOException("An error occurred while creating/writing the file: "
```
kapil-agnihotri, line 754: This exception will be thrown when closing the file so the text should be modified. 

I think this particular error message should move to the above catch block (you may also mention `cpg` somewhere in the text to be more specific for the origin of the exception).

io/plugins/eu.esdihumboldt.hale.io.shp.test/src/eu/esdihumboldt/hale/io/shp/ShapefileInstanceWriterTest.groovy
```diff
@@ -117,6 +117,19 @@ class ShapefileInstanceWriterTest {
 		return reader.getInstances();
 	}
 
+	@Test
+	public void whenReadWithBufferedReader_thenCorrect()
```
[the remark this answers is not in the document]
emanuelaepure10: I didn't create this method. 
Only method I have created in this class is testCPGFileCreation().

io/plugins/eu.esdihumboldt.hale.io.shp.test/src/eu/esdihumboldt/hale/io/shp/ShapefileInstanceWriterTest.groovy
```diff
@@ -117,6 +117,19 @@ class ShapefileInstanceWriterTest {
...
+	throws IOException {
+		String expected_value = "UTF-8";
+		String file ="src/test/resources/fileTest.txt";
```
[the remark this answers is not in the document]
emanuelaepure10: I didn't write this text. I don't know why is seen as addition.

io/plugins/eu.esdihumboldt.hale.io.shp.test/src/eu/esdihumboldt/hale/io/shp/ShapefileInstanceWriterTest.groovy
emanuelaepure10: Sure!

io/plugins/eu.esdihumboldt.hale.io.shp.test/src/eu/esdihumboldt/hale/io/shp/ShapefileInstanceWriterTest.groovy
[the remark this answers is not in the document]
stempler: If it is not relevant then I would suggest to remove this method.

io/plugins/eu.esdihumboldt.hale.io.shp.test/src/eu/esdihumboldt/hale/io/shp/ShapefileInstanceWriterTest.groovy
stempler: > But I think that this test is already done in one of the already created test, example simpleSchemaWriteTest(), This test would create the .shp and .cpg files.

Yes, I think it would also be fine to extend existing tests to also check if the CPG file was written and the content is correct.

io/plugins/eu.esdihumboldt.hale.io.shp.test/src/eu/esdihumboldt/hale/io/shp/ShapefileInstanceWriterTest.groovy
```diff
@@ -1443,4 +1456,24 @@ class ShapefileInstanceWriterTest {
...
+		def tmpDir = Files.createTempDirectory("CPGFileTest")
+		def tmpFile = Files.createTempFile(tmpDir, 'new', '.cpg')
+
+		try {
+			ShapefileInstanceWriter writer = new ShapefileInstanceWriter();
+			writer.writeCodePageFile(tmpFile.toFile().getAbsolutePath());
+			assertTrue(tmpFile.toFile().exists());
+
+			String expected_value = writer.getCharset().toString();
+			BufferedReader bufferedReader = new BufferedReader(new FileReader(tmpFile.toFile().getAbsolutePath()));
```
stempler, line 1471: Just an FYI - since we use Groovy here, there is a simple way to get the text content of a file:
https://docs.groovy-lang.org/latest/html/groovy-jdk/java/io/File.html#getText()

io/plugins/eu.esdihumboldt.hale.io.shp/src/eu/esdihumboldt/hale/io/shp/writer/ShapefileInstanceWriter.java
```diff
@@ -724,4 +730,29 @@ private List<String> writeToFile(
...
+	 * 
+	 * @param filePath Path of the file to be written with just one line of the
+	 *            encoding
+	 * @throws IOException exception in any.
+	 */
+	public void writeCodePageFile(String filePath) throws IOException {
+		File file = new File(filePath);
+		FileWriter fileWriter = new FileWriter(file);
+		try {
+			fileWriter.write(getCharset() != null ? getCharset().toString()
```
stempler, line 743: Now that you use the writer setting to document the encoding in the CPG file, you also need to make sure that the writer does write the Shapefile with this encoding. May assumption would be that this can be done [similar to in the reader](https://github.com/halestudio/hale/blob/37d0458c0868ccf5ff0adf4cb44bb9b7fb07776a/io/plugins/eu.esdihumboldt.hale.io.shp/src/eu/esdihumboldt/hale/io/shp/reader/internal/ShapeInstanceReader.java#L105).

io/plugins/eu.esdihumboldt.hale.io.shp.test/src/eu/esdihumboldt/hale/io/shp/ShapefileInstanceWriterTest.groovy
stempler: A bonus would be adding a test where a Shapefile with special characters is written and read again, with a specific charset configured for both, to verify that charset settings for writer and reader work as expected.

io/plugins/eu.esdihumboldt.hale.io.shp.test/src/eu/esdihumboldt/hale/io/shp/ShapefileInstanceWriterTest.groovy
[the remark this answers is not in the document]
kapil-agnihotri: Neither we 😉. 
Let's delete it if it is not required.

io/plugins/eu.esdihumboldt.hale.io.shp.test/src/eu/esdihumboldt/hale/io/shp/ShapefileInstanceWriterTest.groovy
kapil-agnihotri: Yes, it must be. But still it would be good to verify either in the same test or in an additional test.

io/plugins/eu.esdihumboldt.hale.io.shp/src/eu/esdihumboldt/hale/io/shp/writer/ShapefileInstanceWriter.java
```diff
@@ -724,4 +730,29 @@ private List<String> writeToFile(
 		return filesWritten;
 	}
 
+	/**
+	 *
```
kapil-agnihotri, line 734: I meant that you can move the comment : `Create the CPG file starting from the Shapefile names` here to this newly added method as it belongs here.

io/plugins/eu.esdihumboldt.hale.io.shp/src/eu/esdihumboldt/hale/io/shp/writer/ShapefileInstanceWriter.java
```diff
@@ -724,4 +730,35 @@ private List<String> writeToFile(
...
+	public void writeCodePageFile(String cpgFilePath) throws IOException {
+		File cpgFile = new File(cpgFilePath);
+		FileWriter fileWriter = new FileWriter(cpgFile);
+
+		ShapefileDataStore store = new ShapefileDataStore(
+				new File(cpgFilePath.replace(ShapefileConstants.CPG_EXTENSION,
+						ShapefileConstants.SHP_EXTENSION)).toURL());
+
+		try {
+			fileWriter.write(store.getCharset() != null ? store.getCharset().toString()
```
stempler, line 749: Looking at the code for [ShapefileDataStore](https://github.com/geotools/geotools/blob/5e3616c984b432573e2e58ad51b133f2580ddb0e/modules/plugin/shapefile/src/main/java/org/geotools/data/shapefile/ShapefileDataStore.java#L98C32-L98C32) this would then always be the default value from there (ISO 8859-1).

I preferred the solution from before, using the writer setting - this would then have opened the possibility for users to specify the desired encoding. As mentioned before, that would have required making sure that this charset is used when writing the Shapefile, which should be easily possible where the Shapefile data store for writing the respective file is obtained.

With how it is done here now, it is very intransparent which charset is used and why, and we solely rely on that the default is used here and also where the Shapefile is actually written.

io/plugins/eu.esdihumboldt.hale.io.shp.test/src/eu/esdihumboldt/hale/io/shp/ShapefileInstanceWriterTest.groovy
```diff
@@ -184,11 +186,25 @@ class ShapefileInstanceWriterTest {
...
 		}
 	}
 
+	private static testCpgFileAgainstRelatedSpapefile(File shpFile) {
+		ShapefileDataStore store = new ShapefileDataStore(shpFile.toURL());
+		String cpgFilePath = shpFile.getAbsolutePath().replace(ShapefileConstants.SHP_EXTENSION, ShapefileConstants.CPG_EXTENSION)
+
+		def cpgFile = new File(cpgFilePath)
+		if (cpgFile.exists()) {
+			assertTrue(store.getCharset().toString().equals(cpgFile.text))
```
stempler, line 202: As mentioned in the comment on the writer class, the `store.getCharset()` here will always just return the default value for the charset that is specified in the `ShapefileDataStore` class, and thus this check only works because anyway always the same charset is used for writing the files.

io/plugins/eu.esdihumboldt.hale.io.shp.test/src/eu/esdihumboldt/hale/io/shp/ShapefileInstanceWriterTest.groovy
```diff
@@ -184,11 +186,25 @@ class ShapefileInstanceWriterTest {
...
 			writeInstances(tmpFile.toFile(), schema, instances, configurator)
 			handler.accept(tmpFile.toFile())
+
+			testCpgFileAgainstRelatedSpapefile(tmpFile.toFile())
 		} finally {
 			tmpDir.deleteDir()
 		}
 	}
 
+	private static testCpgFileAgainstRelatedSpapefile(File shpFile) {
```
kapil-agnihotri, line 196: Misspelled word `Shapefile` in `testCpgFileAgainstRelatedSpapefile`?

stempler:
@kapil-agnihotri Any idea related to the build failure?
Since the last build was yesterday it should not be related to the Artifactory downtime. I also would not expect the problem that we saw in hale-cli, since that was Gradle related (though if we can't find anything else it would be worth to also look into changing the artifactory repo, if it is part of the repos that Maven uses here).

kapil-agnihotri:
The error doesn't look like from our repositories but from groovy.jfrog.io artifactory.
The error states that `Failure to find org.osgi:org.osgi.service.prefs:jar:1.1.2 in https://groovy.jfrog.io/artifactory/plugins-release `. I checked manually in the [artifactory](https://groovy.jfrog.io/ui/native/plugins-release/org/osgi/) and the jar is not available now. 
I created [PR](https://github.com/halestudio/hale/pull/1035) by bringing mavenCentral on top in the repository list and the build now works.

kapil-agnihotri:
retest this please

io/plugins/eu.esdihumboldt.hale.io.shp/src/eu/esdihumboldt/hale/io/shp/writer/ShapefileInstanceWriter.java
stempler: > Now that you use the writer setting to document the encoding in the CPG file, you also need to make sure that the writer does write the Shapefile with this encoding.

I would expect this to require something like

```
newDataStore.setCharset(getCharset());
```

to be done [here](https://github.com/halestudio/hale/blob/68175508ef152681bfb0ac4f9432074f1cf7701c/io/plugins/eu.esdihumboldt.hale.io.shp/src/eu/esdihumboldt/hale/io/shp/writer/ShapefileInstanceWriter.java#L463).

emanuelaepure10: I see. In the [reader is done ](https://github.com/halestudio/hale/blob/68175508ef152681bfb0ac4f9432074f1cf7701c/io/plugins/eu.esdihumboldt.hale.io.shp/src/eu/esdihumboldt/hale/io/shp/reader/internal/ShapeInstanceReader.java#L105) properly but not in the writer.  I'll update the code.

stempler (approved):

stempler:
@emanuelaepure10 
Not easily understandable from the stacktrace in the failing tests what is going wrong here. Maybe you can try to debug one of these locally? (Or does the problem not exist when you run them locally?)

Advice on running unit tests in hale studio can be found here (sadly the "Run/debug as JUnit Plug-in test" in Eclipse often is quite slow to start depending on the Eclipse version, so there is an alternative described):
https://github.com/halestudio/hale/wiki/Tests#run-a-single-test-class-or-method

emanuelaepure10:
@stempler I get 15 tests complete with success.
Thank you very much for the link.

io/plugins/eu.esdihumboldt.hale.io.shp.test/src/eu/esdihumboldt/hale/io/shp/ShapefileInstanceWriterTest.groovy
```diff
@@ -293,16 +313,18 @@ class ShapefileInstanceWriterTest {
 					def jts = the_geom.geometry
 					assert jts instanceof Point
 					def name = inst.p.name.value()
-					assert name
-					switch (name) {
+
+					String dencodedString = new String(name.getBytes(shapeFileDataStore.getCharset()));
```
kapil-agnihotri, line 317: Did you mean to write `decodedString`? In other methods as well  `decoded` is misspelt.

io/plugins/eu.esdihumboldt.hale.io.shp.test/src/eu/esdihumboldt/hale/io/shp/ShapefileInstanceWriterTest.groovy
```diff
@@ -368,8 +391,9 @@ class ShapefileInstanceWriterTest {
 
 					assert jts instanceof MultiPolygon
 					def name = inst.p.name.value()
-					assert name
-					switch (name) {
+					String dencodedString = new String(name.getBytes(shapeFileDataStore.getCharset()));
+					assert dencodedString
+					switch (dencodedString) {
```
kapil-agnihotri, line 396: It would be great if here as well in the default case you can mention `decodedString` instead of `typeName`

io/plugins/eu.esdihumboldt.hale.io.shp.test/src/eu/esdihumboldt/hale/io/shp/ShapefileInstanceWriterTest.groovy
```diff
@@ -439,8 +464,9 @@ class ShapefileInstanceWriterTest {
 						def jts = the_geom.geometry
 						assert jts
 						def name = inst.p.name.value()
-						assert name
-						switch (name) {
+						String dencodedString = new String(name.getBytes(shapeFileDataStore.getCharset()));
+						assert dencodedString
+						switch (dencodedString) {
```
kapil-agnihotri, line 469: It would be great if here as well in the default case you can mention `decodedString` instead of `typeName`

io/plugins/eu.esdihumboldt.hale.io.shp.test/src/eu/esdihumboldt/hale/io/shp/ShapefileInstanceWriterTest.groovy
```diff
@@ -1178,8 +1205,9 @@ class ShapefileInstanceWriterTest {
 					def jts = geom.geometry
 					assert jts instanceof Point
 					def name = inst.p.name.value()
-					assert name
-					switch (name) {
+					String dencodedString = new String(name.getBytes(shapeFileDataStore.getCharset()));
+					assert dencodedString
+					switch (dencodedString) {
```
kapil-agnihotri, line 1210: It would be great if here as well in the default case you can mention `decodedString` instead of `typeName`

io/plugins/eu.esdihumboldt.hale.io.shp.test/src/eu/esdihumboldt/hale/io/shp/ShapefileInstanceWriterTest.groovy
```diff
@@ -1358,8 +1387,9 @@ class ShapefileInstanceWriterTest {
 
 					assert jts instanceof MultiPolygon
 					def name = inst.p.name.value()
-					assert name
-					switch (name) {
+					String dencodedString = new String(name.getBytes(shapeFileDataStore.getCharset()));
+					assert dencodedString
+					switch (dencodedString) {
```
kapil-agnihotri, line 1392: It would be great if here as well in the default case you can mention `decodedString` instead of `typeName`

io/plugins/eu.esdihumboldt.hale.io.shp.test/src/eu/esdihumboldt/hale/io/shp/ShapefileInstanceWriterTest.groovy
```diff
@@ -1432,8 +1463,9 @@ class ShapefileInstanceWriterTest {
...
```
kapil-agnihotri, line 1468: It would be great if here as well in the default case you can mention `decodedString` instead of `typeName`

io/plugins/eu.esdihumboldt.hale.io.shp.test/src/eu/esdihumboldt/hale/io/shp/ShapefileInstanceWriterTest.groovy
emanuelaepure10: Yes :-) I'll rename them

PR #1035, opened 2023-07-04
build: change priority list of pluginRepositories

kapil-agnihotri:
groovy-jfrog.io artifactory no longer contains jar: org.osgi:org.osgi.service.prefs:jar:1.1.2, so priority list of pluginRepositories is now changed to include maven central on top


stempler (approved):
Thanks a lot for the fix!

PR #1036, opened 2023-07-05
Feat_maint/4.x: extend shapefile writer to create .cpg file

emanuelaepure10:
Same functionalities developed on the 5.x have been reproduce for the v 4.x.
 
Create .cpg file/files contain only the encoding used to write the shapefile/s related.
Update the related test class to check if the file has been created and contains the right encoding code.


stempler:
@emanuelaepure10 Are the changes the same as the ones for master and build locally? Then I would say we can ignore the failing build because that targets the new version.

emanuelaepure10:
@stempler same changes as master.

emanuelaepure10:
@stempler I have create a new PR contains the missing import that made this PR failing https://github.com/halestudio/hale/pull/1038

PR #1038, opened 2023-07-10
Update feat_maint_4.x: extend shapefile writer to create .cpg file

emanuelaepure10:
The new PR contains the update with the missing import that makes the following PR failing:
Reference old failed PR: https://github.com/halestudio/hale/pull/1036

### DIFF
--- a/io/plugins/eu.esdihumboldt.hale.io.shp.test/src/eu/esdihumboldt/hale/io/shp/ShapefileInstanceWriterTest.groovy
+++ b/io/plugins/eu.esdihumboldt.hale.io.shp.test/src/eu/esdihumboldt/hale/io/shp/ShapefileInstanceWriterTest.groovy
@@ -26,6 +26,7 @@ import java.time.Instant
 import java.time.LocalDate
 import java.util.function.Consumer
 
+import org.geotools.data.shapefile.ShapefileDataStore
 import org.junit.Test
 import org.locationtech.jts.geom.Coordinate
 import org.locationtech.jts.geom.Geometry


kapil-agnihotri:
I think it is not required to create a PR with all the changes again as the [old PR](https://github.com/halestudio/hale/commit/a4c4246abaa688c2dbb92b04ef85713cd8dd0537) already had these changes. 
In the new commit you would only need to add the import statement. 
Git shows all the files with the same changes that you already checked in to maint/4.x branch.

emanuelaepure10:
I've seen that a good practice it is to create a new branch and refer it from the new one.
But if it's fine for you to update just the old one I can do that now or in the next ones (hopefully wont be the case).

kapil-agnihotri:
> In the new commit you would only need to add the import statement.

May be my statement was confusing. With the new commit I meant that a new commit in this PR.

If you could update this branch that you probably created from your fork repository (without updating your local maint/4.x)  with the [hale/4.x branch](https://github.com/halestudio/hale/commits/maint/4.x), you will see that there will be conflicts. As this branch is not upto date with the [hale/maint/4.x branch](https://github.com/halestudio/hale/commits/maint/4.x). 

When this branch is upto date with the latest hale/maint/4.x then there would only be one change that will correspond to addition of below import statement and will not contain all the changes that you already merged to hale/maint/4.x as a part of [#1036](https://github.com/halestudio/hale/pull/1036))

```
import org.geotools.data.shapefile.ShapefileDataStore
```

kapil-agnihotri (commented):
LGTM.
As per developer guidelines in [Guru](https://app.getguru.com/card/iXrR4qoT/Developer-Guidelines?query=bucket-service), you need to mention ticket number in the footer of the commit message.

kapil-agnihotri:
From the Guru card:

```
In case a bug in a newly introduced feature or another kind of change only becomes apparent after this change was merged, but before a release was created, it is not desired for this fix to show up as a separate entry in the changelog.
Thus it should reference the original change by adding the prefix fixup! to the original commit message.
```

So, I believe, that a new commit message that you added in this PR (which is exactly same as the previous PR should be replaced as per developer Guidelines. As this is a fix to your already committed changes before release. Please refer developer guidelines for more details. 

For the ticket number: 

```
The footer should contain a closing reference to an issue if any, or generally a reference to an issue, 
for instance to a JIRA issue.
```

As we replicate all the GitHub issues to JIRA for hale studio, so we reference to JIRA issues in the commit messages instead of GitHub issues.

kapil-agnihotri:
I believe that the commit message can still be updated as per our guidelines:

```
fixup! feat: extend shapefile writer to create .cpg file

Add missing import class into the test ShapefileInstanceWriterTest.groovy

ING-3911
```

I don't think that there is any necessity to reference to the previous PR url (`Reference previous PR: https://github.com/halestudio/hale/pull/1036`)  because the subject line of this commit itself states that this is a fixup for commit `feat: extend shapefile writer to create .cpg file`. Also, it is not so complicated fix where someone has to look back to the old PR to understand it but if you still feel that it is required then it can be added.
 
Also, we usually do not copy complete url of the ING ticket instead just the format as mentioned in the Guru card.

PR #1039, opened 2023-07-11
Release candidate 5.0.1

emanuelaepure10:
Has been used the workflow 
https://github.com/halestudio/hale/blob/master/RELEASE.md
to update and tag the release 5.0.1


app/plugins/eu.esdihumboldt.hale.app.cli.commands/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: hale CLI Default Commands
 Bundle-SymbolicName: eu.esdihumboldt.hale.app.cli.commands;singleton:=true
-Bundle-Version: 5.0.0.qualifier
+Bundle-Version: 5.1.0.qualifier
```
stempler, line 5: I guess the versions should match the release versions? i.e. 5.0.1 instead of 5.1.0?

app/plugins/eu.esdihumboldt.hale.app.cli/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: hale CLI application
 Bundle-SymbolicName: eu.esdihumboldt.hale.app.cli;singleton:=true
-Bundle-Version: 5.0.0.qualifier
+Bundle-Version: 5.1.0.qualifier
```
kapil-agnihotri, line 5: Here s well `5.0.1`. I believe it would be great to change this first in master using the groovy script and then merging to the release branch. 

I believe you can run the below script on master and then create a PR.

```
./updateversionnumbers.groovy -o 5.1.0 -n 5.0.1
```

Once the PR is merged to the master, then master can be merged to the release branch so that we have a well maintained commit history in master which then will be merged to the release branch. 

But we can also wait for @stempler's feedback if he has any other proposal.

emanuelaepure10: And than we need to change back the master to 5.1.0?
Why we just don't keep 5.0.1 as long as we have the next version and we change it when the release will arrive when we will know exactly the version number?

stempler:
I merged the changes to the release branch. Let's see if the release build succeeds and the results look as expected.

stempler:
Build was successful: https://builds.wetransform.to/job/hale/job/hale~publish(release)/27/

PR #1042, opened 2023-07-12
basic framework for reading Json/GeoJson data

stempler:


emanuelaepure10 (approved):

stempler:
@emanuelaepure10 What would you prefer - should we merge it or would you rather like to use these changes only locally for now?

emanuelaepure10:
@stempler I actually have no preference. 
I would incline more to keep it because the code it's part of the issue ING-3128 that needs to be completed.

emanuelaepure10:
@stempler I think that merging it makes our life easier into developing the issue ahead

Issue #1032, opened 2023-06-20
Export more than one type when exporting to an Excel file


thorsten-reitz:
Currently, it is only possible to export one type into an Excel file ("Choose your Type you want to export"). This should be changed to allow the export of multiple types to multiple tabs of an Excel file.

PR #1043, opened 2023-07-12
feat: export multiple feature type to Excel

emanuelaepure10:
HALE studio supports selecting multiple feature types for export to Excel. 
Each selected feature type is added as a separate sheet in a single Excel workbook.

https://github.com/halestudio/hale/issues/1032


io/plugins/eu.esdihumboldt.hale.io.xls.test/src/eu/esdihumboldt/hale/io/xls/test/writer/XLSInstanceWriterTest.java
```diff
@@ -209,6 +208,49 @@ public void testWriteNotNestedProperties() throws Exception {
...
+		writer.setParameter(InstanceTableIOConstants.USE_SCHEMA, Value.of(false));
+		writer.setParameter(InstanceTableIOConstants.EXPORT_EMPTY_FEATURETYPES, Value.of(true));
+
+		File tmpFile = tmpFolder.newFile("excelWith2Sheets.xls");
+		// write instances to a temporary XLS file
+		writer.setTarget(new FileIOSupplier(tmpFile));
+		writer.setContentType(contentType);
+
+		for (InstanceCollection instanceCollection : examples) {
+			writer.setInstances(instanceCollection);
```
stempler, line 240: `setInstances` in this case goes back to this implementation: https://github.com/halestudio/hale/blob/5d335696fd7b8f7e2114d3423ea5d2cacf2445f2/common/plugins/eu.esdihumboldt.hale.common.instance/src/eu/esdihumboldt/hale/common/instance/io/impl/AbstractInstanceWriter.java#L88

By setting it multiple times you just replace what was set before.

Instead you would need to use something like this to create a combined instance collection:
https://github.com/halestudio/hale/blob/5d335696fd7b8f7e2114d3423ea5d2cacf2445f2/common/plugins/eu.esdihumboldt.hale.common.instance/src/eu/esdihumboldt/hale/common/instance/model/impl/MultiInstanceCollection.java#L42

io/plugins/eu.esdihumboldt.hale.io.xls.ui/src/eu/esdihumboldt/hale/io/xls/ui/XLSInstanceExportConfigurationPage.java
```diff
@@ -0,0 +1,200 @@
...
+import eu.esdihumboldt.hale.common.schema.model.TypeDefinition;
+import eu.esdihumboldt.hale.io.csv.InstanceTableIOConstants;
+import eu.esdihumboldt.hale.io.csv.ui.InstanceExportConfigurationPage;
+
+/**
+ * Configuration page for exporting Excel
+ * 
+ * @author Emanuela Epure
+ */
+public class XLSInstanceExportConfigurationPage extends InstanceExportConfigurationPage {
```
stempler, line 51: Why extend the CSV page?
To me it seems you overwrite all logic defined in that page without calling the super implementation, so basically nothing from the super class is actually used?

io/plugins/eu.esdihumboldt.hale.io.xls/src/eu/esdihumboldt/hale/io/xls/writer/XLSInstanceWriter.java
```diff
@@ -73,86 +74,118 @@ protected IOReport execute(ProgressIndicator progress, IOReporter reporter)
...
 
+		boolean emptyFeatureTypes = getParameter(InstanceTableIOConstants.EXPORT_EMPTY_FEATURETYPES)
+				.as(Boolean.class, false);
+
 		// get the parameter to get the type definition
 		String exportType = getParameter(InstanceTableIOConstants.EXPORT_TYPE).as(String.class);
-		QName selectedTypeName = null;
+		ArrayList<QName> selectedFeatureTypes = new ArrayList<>();
+
+		String[] array = exportType.split(",");
```
stempler, line 84: The definition and documentation of the parameter should be adapted to the new behavior:
https://github.com/halestudio/hale/blob/5d335696fd7b8f7e2114d3423ea5d2cacf2445f2/io/plugins/eu.esdihumboldt.hale.io.xls/plugin.xml#L94

io/plugins/eu.esdihumboldt.hale.io.xls/src/eu/esdihumboldt/hale/io/xls/writer/XLSInstanceWriter.java
```diff
@@ -73,86 +74,118 @@ protected IOReport execute(ProgressIndicator progress, IOReporter reporter)
 		boolean solveNestedProperties = getParameter(
 				InstanceTableIOConstants.SOLVE_NESTED_PROPERTIES).as(Boolean.class, false);
 
+		boolean emptyFeatureTypes = getParameter(InstanceTableIOConstants.EXPORT_EMPTY_FEATURETYPES)
```
stempler, line 77: The newly introduce parameter should be defined and documented in the extension, e.g. similar to here:
https://github.com/halestudio/hale/blob/5d335696fd7b8f7e2114d3423ea5d2cacf2445f2/io/plugins/eu.esdihumboldt.hale.io.xls/plugin.xml#L112

io/plugins/eu.esdihumboldt.hale.io.xls.ui/src/eu/esdihumboldt/hale/io/xls/ui/XLSInstanceExportConfigurationPage.java
emanuelaepure10: At this moment I'm just using 2 buttons name from the top page, but as no super() is used I could extent the page to AbstractConfigurationPage<InstanceWriter, IOWizard<InstanceWriter>> and have the buttons inside XLSInstanceExportConfigurationPage.

io/plugins/eu.esdihumboldt.hale.io.xls/src/eu/esdihumboldt/hale/io/xls/writer/XLSInstanceWriter.java
stempler: By the way, the easiest way to add/edit the extensions is using the dedicated editor in Eclipse instead of changing the XML directly. That is also the least error prone.

io/plugins/eu.esdihumboldt.hale.io.xls.ui/src/eu/esdihumboldt/hale/io/xls/ui/XLSInstanceExportConfigurationPage.java
stempler: I think just for reusing field declarations it does not make sense to extend the other class.
If there is similar functionality you could however see if there is potential for sharing the logic between the two classes, e.g. by moving the common functionality to a shared method that can be used by both.

io/plugins/eu.esdihumboldt.hale.io.xls/plugin.xml
```diff
@@ -144,6 +155,15 @@
             </parameterBinding>
          </providerParameter>
       </provider>
+      <provider
```
stempler, line 158: I guess this was added by accident?

io/plugins/eu.esdihumboldt.hale.io.xls/plugin.xml
```diff
@@ -122,6 +123,16 @@
                   defaultDescription="Defaults to true">
             </valueDescriptor>
          </providerParameter>
+         <providerParameter
+               description="Don&apos;t add feature types without data to the exported Excel"
+               label="Ignore feature types without data"
+               name="ignoreEmptyFeaturetypes"
+               optional="false">
```
stempler, line 130: I think it should be optional with a default value.

io/plugins/eu.esdihumboldt.hale.io.xls/plugin.xml
```diff
@@ -122,6 +123,16 @@
...
+            <parameterBinding
+                  class="java.lang.Boolean">
+            </parameterBinding>
+            <valueDescriptor></valueDescriptor>
```
stempler, line 134: The default value can be documented here, similar to the `useSchema` parameter.

io/plugins/eu.esdihumboldt.hale.io.xls/plugin.xml
emanuelaepure10: Yes. It was just a test. I'll remove it.

stempler (approved):

PR #1047, opened 2023-07-31
feat: implement geojson/json instance reader

emanuelaepure10:
GeoJSON/JSON datasets can be directly imported
Still to do:
- correct the total data of an feature type - at the moment shows the sum of all the data related to all the feature types in the project
- check the numbers on the array and geometry imported
- find a way to understand from JSON which instance is part of which feature type

ING-3128


io/plugins/eu.esdihumboldt.hale.io.json/src/eu/esdihumboldt/hale/io/json/internal/JsonInstanceConstants.java
```diff
@@ -15,12 +15,14 @@
...
 
+import eu.esdihumboldt.hale.io.json.JsonInstanceReader;
+
 /**
  * Constants used to generate JSON from instances.
  * 
  * @author Simon Templer
  */
-public interface InstanceJsonConstants {
+public interface JsonInstanceConstants {
```
stempler, line 25: Any specific reason for renaming this class? It comes originally from the Instance to Json conversion and is still used there as well.

io/plugins/eu.esdihumboldt.hale.io.json/src/eu/esdihumboldt/hale/io/json/internal/JsonInstanceConstants.java
```diff
@@ -37,4 +39,23 @@ public interface InstanceJsonConstants {
 	 */
 	public static final String PREFIX_INSTANCE_JSON = "hj";
 
+	/**
```
stempler, line 42: These are constants related to the writer implementation and not to the generic Instance to Json or Json to Instance conversion, so it would be preferrable to not have them added here.

io/plugins/eu.esdihumboldt.hale.io.json/src/eu/esdihumboldt/hale/io/json/writer/NamespaceManager.java
```diff
@@ -13,7 +13,7 @@
  *     wetransform GmbH <http://www.wetransform.to>
  */
 
-package eu.esdihumboldt.hale.io.json.internal;
+package eu.esdihumboldt.hale.io.json.writer;
```
stempler, line 16: Any specific reason for moving this class and many others?

Like this one, many of the classes moved are not only relevant for writing Json but are used for reading Json as well, and were used there before any of the code to write Json was added.

io/plugins/eu.esdihumboldt.hale.io.json.test/src/eu/esdihumboldt/hale/io/json/test/JsonInstanceCollectionTest.groovy
```diff
@@ -58,29 +60,29 @@ class JsonInstanceCollectionTest {
...
-        "type": "LineString",
-        "coordinates": [
-          [102.0, 0.0], [103.0, 1.0], [104.0, 0.0], [105.0, 1.0]
-        ]
-      },
       "properties": {
         "id": "line",
-        "name": "Line feature"
+        "name": "Line feature",
+		"geometry": {
```
stempler, line 66: In principle there can be a geometry inside a property, but in GeoJson the geometry is primarily inside the `geometry` field next to the `properties`. Please have a look at the GeoJson spec.

io/plugins/eu.esdihumboldt.hale.io.json/src/eu/esdihumboldt/hale/io/json/JsonInstanceReader.java
```diff
@@ -58,17 +86,61 @@ protected IOReport execute(ProgressIndicator progress, IOReporter reporter)
 		progress.begin("Creating " + getDefaultTypeName() + " parser", ProgressIndicator.UNKNOWN);
 
 		try {
-			boolean expectGeoJson = true; // currently defaults to true, no
-											// major difference in functionality
+			boolean expectGeoJson = false;
```
stempler, line 89: Not sure what you want to achieve with all the changes here. If it is about supporting using the information in `@type`, then that would be better suited to be done in `JsonToInstance` for each individual instance.

io/plugins/eu.esdihumboldt.hale.io.json/src/eu/esdihumboldt/hale/io/json/JsonInstanceReader.java
```diff
@@ -58,17 +86,61 @@ protected IOReport execute(ProgressIndicator progress, IOReporter reporter)
...
+			progress.setCurrentTask("Extracting json instances");
+
+			String featureCollection = "FeatureCollection";
+			String typeFeatureCollection = "type";
+
+			String targetField = "@type";
+			ObjectMapper objectMapper = new ObjectMapper();
+
+			try {
+				JsonNode rootNode = objectMapper.readTree(new File(getSource().getLocation()));
```
stempler, line 99: This reads the whole Json document into memory, that is definitely not something we want to do.
Ideal would be to only use the streaming API, but that is hard to do because there is no guaranteed order the fields will appear in. So the compromise I settled for was only reading a whole object/feature into memory at a time.

stempler (commented):
I would again like to encourage adding/extending unit tests when working on implementing a certain functionality. This is important for avoiding regressions and also greatly helps for the review.

PR #1048, opened 2023-08-04
feat: export valid shapefiles for mappings including feature types with no geometry

emanuelaepure10:
Export valid shapefiles for feature types with geometries. 
Displays a warning that feature types with no geometry will not be exported to shapefile.

ING-3974

### DIFF
--- a/io/plugins/eu.esdihumboldt.hale.io.shp/src/eu/esdihumboldt/hale/io/shp/writer/ShapefileInstanceWriter.java
+++ b/io/plugins/eu.esdihumboldt.hale.io.shp/src/eu/esdihumboldt/hale/io/shp/writer/ShapefileInstanceWriter.java
@@ -118,7 +118,8 @@ public class ShapefileInstanceWriter extends AbstractGeoInstanceWriter {
 					return file.toURI();
 				}).collect(Collectors.toList());
 
-				// Reset the target property so that a caller can find out which
+				// Reset the target property so that a caller can find out
+				// which
 				// files were created.
 				setTarget(new MultiLocationOutputSupplier(uris));
 			}
@@ -152,7 +153,7 @@ public class ShapefileInstanceWriter extends AbstractGeoInstanceWriter {
 	 * 3. create features <br>
 	 * 4. write the feature data to the Shapefile.
 	 * 
-	 * @param instances instance to write to.
+	 * @param instanceCollection instance to write to.
 	 * @param progress the progress indicator.
 	 * @param reporter the reporter.
 	 * @param location file path URI.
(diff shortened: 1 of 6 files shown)


io/plugins/eu.esdihumboldt.hale.io.shp.ui/src/eu/esdihumboldt/hale/io/shp/ui/SimpleTargetCRSConfigurationPage.java
```diff
@@ -0,0 +1,190 @@
...
+
+/**
+ * Configuration page for the target CRS.
+ * 
+ * @param <W> the concrete I/O wizard type
+ * @param <P> the {@link IOProvider} type used in the wizard
+ * 
+ * @author Simon Templer
+ */
+public class SimpleTargetCRSConfigurationPage<P extends GeoInstanceWriter, W extends IOWizard<P>>
```
stempler, line 48: Instead of copying the whole class I would prefer an approach that does not duplicate so much code.

This seems to be the main change:
image

What about adding an optional constructor argument to the original class that either gets a boolean if the message should be shown or a string for the message text and depending on the value shows the message.
Then this class here would just implement a default constructor that calls the super constructor with the respective parameter.

io/plugins/eu.esdihumboldt.hale.io.shp/src/eu/esdihumboldt/hale/io/shp/writer/ShapefileInstanceWriter.java
```diff
@@ -226,13 +227,15 @@ private Map<String, Map<String, SimpleFeatureType>> createFeatureType(
...
-			reporter.error(
-					"Cannot create Shapefile for Schema: " + localPart + " as no Geometry found!!");
-		}
+
 		// create SimpleFeatureType from SimpleFeatureTypeBuilder.
 		for (Entry<String, Map<String, SimpleFeatureTypeBuilder>> schemaEntry : schemaBuilderMap
 				.entrySet()) {
+
+			if (missingGeomsForSchemas.contains(schemaEntry.getKey())) {
+				continue;
```
stempler, line 236: Maybe log a warning in the reporter that the type was skipped? That would IMHO be better than just silently skipping it.

io/plugins/eu.esdihumboldt.hale.io.shp.ui/META-INF/MANIFEST.MF
```diff
@@ -24,5 +24,9 @@ Import-Package: com.google.common.io;version="9.0.0",
  eu.esdihumboldt.util
 Require-Bundle: org.eclipse.core.runtime;bundle-version="3.7.0",
  org.eclipse.ui;bundle-version="3.7.0",
- eu.esdihumboldt.hale.io.shp;bundle-version="2.5.0"
+ eu.esdihumboldt.hale.io.shp;bundle-version="2.5.0",
+ org.geotools,
```
stempler, line 28: Are all these additional imports necessary?

stempler (approved):

io/plugins/eu.esdihumboldt.hale.io.shp.ui/META-INF/MANIFEST.MF
emanuelaepure10: No need. I just update the file.

PR #1051, opened 2023-08-07
fixup! feat: download wfs files to tmp folder for bucket service

kapil-agnihotri:
Refactor to rename downloaded file to not start with a "." and to have name format as
<datasetId>_<numberOfTotalFeatureProcessed>

ING-3609


stempler (approved):
LGTM

PR #1053, opened 2023-08-16
extend Json reader functionality

stempler:
- support more simple types (still no complex properties supported though)
- support using `@type` field to determine instance type
- support reading GeoJson geometries
- support different kind of source file structures (configuration parameter)
- support configuration of default type to use for read instances, by default applies if type cannot be detected in another way, but can be forced to be used via the configuration
- basic support for reading complex properties


stempler:
retest this please

PR #1054, opened 2023-08-21
feat: GeoJSON/JSON wizard additional settings page

emanuelaepure10:
Add 3 parameters for the reader that should be configurable via the UI:
-A “read mode” that controls what kind of document structure can be parsed
-A type name (QName string representation similar to how it is used for Shapefile and CSV readers) for the default type to associate to read instances
-A setting to force the above default type to be used

ING-3128


io/plugins/MANIFEST.MF
```diff
@@ -0,0 +1,7 @@
+Manifest-Version: 1.0
```
stempler, line 1: Was this file created by mistake?

io/plugins/eu.esdihumboldt.hale.io.json.ui/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,28 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: ui
+Bundle-SymbolicName: eu.esdihumboldt.hale.io.json.ui;singleton:=true
+Bundle-Version: 1.0.0.qualifier
```
stempler, line 5: Please adapt the version to the same as other hale bundles.

io/plugins/eu.esdihumboldt.hale.io.json.ui/src/eu/esdihumboldt/hale/io/json/ui/util/EnumJSONAutoOrderTypes.java
```diff
@@ -0,0 +1,59 @@
...
+package eu.esdihumboldt.hale.io.json.ui.util;
+
+/**
+ * Enumeration for different types of winding orders for Geometry. Selected type
+ * will be applied geometry. In case of Polygon/MultiPolygon selected type will
+ * be applied to the shell and the reversed type is applied to the holes.
+ * 
+ * @author Arun
+ */
+public enum EnumJSONAutoOrderTypes {
```
stempler, line 25: This replicates the [JsonReadMode](https://github.com/halestudio/hale/blob/master/io/plugins/eu.esdihumboldt.hale.io.json/src/eu/esdihumboldt/hale/io/json/internal/JsonReadMode.java) enum already defined and used by the reader. It should only be defined once.

io/plugins/eu.esdihumboldt.hale.io.json.ui/src/eu/esdihumboldt/hale/io/json/ui/TypeSelectionPage.java
```diff
@@ -0,0 +1,272 @@
...
+				@Override
+				public String getText(Object element) {
+					if (element instanceof EnumJSONAutoOrderTypes) {
+						return ((EnumJSONAutoOrderTypes) element).getJsonModeOrder();
+					}
+					return super.getText(element);
+				}
+
+			});
+			readModeOrderCombo.setInput(EnumJSONAutoOrderTypes.class);
```
stempler, line 126: I don't see the initial selection being set on the combo. The initial selection should reflect the default for the reader (`auto`).

io/plugins/MANIFEST.MF
emanuelaepure10: Yes. I have removed it

io/plugins/eu.esdihumboldt.hale.io.json.ui/src/eu/esdihumboldt/hale/io/json/ui/TypeSelectionPage.java
```diff
@@ -0,0 +1,272 @@
...
+ * 
+ * All rights reserved. This program and the accompanying materials are made
+ * available under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation, either version 3 of the License,
+ * or (at your option) any later version.
+ * 
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this distribution. If not, see <http://www.gnu.org/licenses/>.
+ * 
+ * Contributors:
```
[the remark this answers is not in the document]
emanuelaepure10: I have replaced it by 

--------------
/*
 * Copyright (c) 2021 wetransform GmbH
 * 
 * All rights reserved. This program and the accompanying materials are made
 * available under the terms of the GNU Lesser General Public License as
 * published by the Free Software Foundation, either version 3 of the License,
 * or (at your option) any later version.
 * 
 * You should have received a copy of the GNU Lesser General Public License
 * along with this distribution. If not, see <http://www.gnu.org/licenses/>.
 * 
 * Contributors:
 *     wetransform GmbH <http://www.wetransform.to>
 */
 ------------------
 but the same I have used seems to be used 1591 times in the project and the above one just 570 times. 
 Shouldn't use just one header everywhere?

io/plugins/eu.esdihumboldt.hale.io.json.ui/src/eu/esdihumboldt/hale/io/json/ui/TypeSelectionPage.java
```diff
@@ -0,0 +1,272 @@
...
+	protected void onShowPage(boolean firstShow) {
+		super.onShowPage(firstShow);
+
+		page.setLayout(new GridLayout(2, false));
+		GridData layoutData = new GridData();
+		layoutData.widthHint = 200;
+		
+		if (firstShow) {
+
+			Group autodetectOrderGroup = new Group(page, SWT.NONE);
```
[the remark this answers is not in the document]
emanuelaepure10: I was also in doubt where to create it.
In CSV is created in `createContent`, instead in shp in `onShowPage`. 
I have moved it.

io/plugins/eu.esdihumboldt.hale.io.json.ui/src/eu/esdihumboldt/hale/io/json/ui/TypeSelectionPage.java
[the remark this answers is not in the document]
stempler: Well the year when the code was created is no constant and mentioning Fraunhofer and the data harmonisation panel as contributors is not correct. If you create a new class it will automatically add the header as configured in the current template.
The license is usually the same, though there are exceptions for some extensions that have a different license.

stempler (approved):

PR #1057, opened 2023-08-25
fix: fix endless loop when skipping instances in paginated WFS request

stempler:
Relates to https://github.com/wetransform/bucket-service/pull/276


kapil-agnihotri (approved):
Should I revert the change back to skip instances in bucket-service?

stempler:
> Should I revert the change back to skip instances in bucket-service?

@kapil-agnihotri I guess for that this change here would also need to be on the hale 4 maintenance branch as well

Issue #878, opened 2021-01-29
Hale Studio freeze while WFS Import


nicolasdiedrich:
Hale studio freezes when importing WFS data.

- Occurs after the occurrence of an error during WFS import when import is repeated.
- The process continues endlessly and aborting is not possible. 
- An error message does not occur again
- Only solution: shut down Hale studio via task manager

![Freeze](https://user-images.githubusercontent.com/75833042/106264220-aebf0b80-6225-11eb-94b5-4fabc3cecc8a.PNG)

dhdeangelis:
I wonder if this issue is solved or any ideas have emerged for a workaround. I am using Hale 4.1.0 (released one week ago) and I am experiencing a similar problem with one particular WFS service: the GUI stalls while loading the source data. Even a small subset, 100-500 features, cannot finish loading.
This particular WFS exposes a total of 1.2 million features but I am not loading all of it at once of course. Is this perhaps related to paging issues? This service works OK in for example QGIS. However, tests with another service (exposing around 85000 features) does not show this problem. IS this possibly showing problems with the WFS service itself?

Thoughts appreciated

/Hernán


![image](https://user-images.githubusercontent.com/51515911/143219676-0728384d-7096-480d-821b-8e9588179c7f.png)

With a small subset I can get the following error log in a reasonable time (several minutes):

`java.lang.IllegalStateException: Could not open instance input
	at eu.esdihumboldt.hale.io.gml.reader.internal.GmlInstanceCollection$GmlInstanceIterator.<init>(GmlInstanceCollection.java:125)
	at eu.esdihumboldt.hale.io.gml.reader.internal.GmlInstanceCollection.iterator(GmlInstanceCollection.java:736)
	at eu.esdihumboldt.hale.io.gml.reader.internal.wfs.WfsBackedGmlInstanceCollection$WfsBackedGmlInstanceIterator.createNextIterator(WfsBackedGmlInstanceCollection.java:421)
	at eu.esdihumboldt.hale.io.gml.reader.internal.wfs.WfsBackedGmlInstanceCollection$WfsBackedGmlInstanceIterator.<init>(WfsBackedGmlInstanceCollection.java:384)
	at eu.esdihumboldt.hale.io.gml.reader.internal.wfs.WfsBackedGmlInstanceCollection.iterator(WfsBackedGmlInstanceCollection.java:266)
[...]

PR #1058, opened 2023-09-01
fix: fix endless loop when skipping instances in paginated WFS request

kapil-agnihotri:
cherry picked commits from master
relates to https://github.com/wetransform/bucket-service/pull/276


JohannaOtt:
@kapil-agnihotri Any chance that this fixes https://github.com/halestudio/hale/issues/878 ?

kapil-agnihotri:
@JohannaOtt they could be related. Although, not very sure if when storing instances in DB they were also skipped. This can also be tested in latest hale studio (master) as the [changes were merged](https://github.com/halestudio/hale/pull/1057) last week.

JohannaOtt:
@kapil-agnihotri Thanks for the input.
Unfortunately no one seems to recall a WFS for which #878 had been occurring. So I cannot test if it is fixed with the changes you made. I assume we will learn if it is persisting if more users report about it happening.

PR #1059, opened 2023-09-01
fixup: feat: add multiple selection of feature types to be exported to CSV

emanuelaepure10:
Same approach as for XSL has been implemented for the CSV writer. In the wizard of selecting a feature type has been implemented the multi selection table and the same number of .csv files as the selection are exported.

ING-3987


io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/writer/internal/CSVInstanceWriter.java
```diff
@@ -138,12 +176,10 @@ protected IOReport execute(ProgressIndicator progress, IOReporter reporter)
 
 			// header is only finished if all properties have been processed
 			// insert header to temp file and write it to output
-			insertHeader(tempFile, getTarget().getOutput(), headerRow);
+			insertHeader(tempFile, new FileOutputStream(filenameWithType), headerRow);
```
stempler, line 179: This breaks compatibility with the previous functionality, because there it was possible to use something as a target that is not a file.
It would be great if that behavior could be kept at least for the case where only one type is selected and thus your `changeName` variable set to `false`.

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/writer/internal/CSVInstanceWriter.java
```diff
@@ -76,27 +79,64 @@ public boolean isCancelable() {
 	protected IOReport execute(ProgressIndicator progress, IOReporter reporter)
 			throws IOProviderConfigurationException, IOException {
 
-		boolean solveNestedProperties = getParameter(
-				InstanceTableIOConstants.SOLVE_NESTED_PROPERTIES).as(Boolean.class, false);
-		// XXX what does "solve nested properties" mean?
+		URI location = getTarget().getLocation();
+
+		String filenameWithType = location.getPath();
+		String filePath = Paths.get(location).getParent().toString();
```
stempler, line 85: An URI is not necessarily a file URI. Also, there is the possibility that the location is null, for cases where only an output stream is provided.
There are some writers that can only write to files (e.g. Geopackage), but the CSV writer so far could also handle other targets (see also comment further below).

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/writer/internal/CSVInstanceWriter.java
emanuelaepure10: Would you mind telling me how to have a target which is not a file to test?
Thank you

stempler: If you want to test this in a unit test you could use a `ByteArrayOutputStream` as target.

emanuelaepure10: I would guess that what you meant is that at the [line](https://github.com/halestudio/hale/blob/ce740b79fa0f04d371abe6f9c61776f62534c0c5/io/plugins/eu.esdihumboldt.hale.io.csv.test/src/eu/esdihumboldt/hale/io/csv/writer/CSVInstanceWriterTest.java#L331C12-L331C12) could be send a supplier of a file `FileIOSupplier `or a `ByteArrayOutputStream`. 
Any tips on how to convert a `ByteArrayOutputStream` into a `FileIOSupplier `? 
Thank you

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/writer/internal/CSVInstanceWriter.java
```diff
@@ -76,27 +79,61 @@ public boolean isCancelable() {
...
+
+					String filePath = Paths.get(location).getParent().toString();
+					String baseFilename = Paths.get(location).getFileName().toString();
+					baseFilename = baseFilename.substring(0, baseFilename.lastIndexOf("."));
+					String filenameWithType = filePath + FileSystems.getDefault().getSeparator()
+							+ baseFilename + InstanceTableIOConstants.UNDERSCORE
+							+ selectedTypeName.getLocalPart()
+							+ InstanceTableIOConstants.CSV_EXTENSION;
+
+					addCSVFileByQName(reporter, new FileOutputStream(filenameWithType), instances);
```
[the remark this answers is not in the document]
stempler: If the stream is closed somewhere else it would be good to at least document that here.

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/writer/internal/CSVInstanceWriter.java
stempler: It shouldn't be a `FileIOSupplier`. I assume it is about [setting the writer target](https://github.com/halestudio/hale/blob/ce740b79fa0f04d371abe6f9c61776f62534c0c5/common/plugins/eu.esdihumboldt.hale.common.core/src/eu/esdihumboldt/hale/common/core/io/ExportProvider.java#L42)?
There the type is `LocatableOutputSupplier`. You could check what existing implementations are there that would fit and if there aren't any create one by implementing the interface.

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/writer/internal/CSVInstanceWriter.java
```diff
@@ -76,27 +79,71 @@ public boolean isCancelable() {
...
+					String filenameWithType = filePath + FileSystems.getDefault().getSeparator()
+							+ baseFilename + InstanceTableIOConstants.UNDERSCORE
+							+ selectedTypeName.getLocalPart()
+							+ InstanceTableIOConstants.CSV_EXTENSION;
+
+					File outputFile = new File(filenameWithType);
+					try {
+						FileOutputStream fos = new FileOutputStream(outputFile);
+						addCSVFileByQName(reporter, fos, instances);
+						fos.close();
```
stempler, line 115: The stream should also be closed when an error occurs. Usually this is done in a `finally` block, or even better in a try-with-resources as mentioned earlier:

> For resources like streams you should always make sure to properly close them. If this is not done this can for instance lead to open file handles that may prevent deleting such a file.
> 
> You could for instance use the [try-with-resources](https://www.baeldung.com/java-try-with-resources) that was introduced in Java 7.

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/writer/internal/CSVInstanceWriter.java
```diff
@@ -76,27 +79,71 @@ public boolean isCancelable() {
...
+					} catch (IOException e) {
+						throw new IOException(
+								"An error occurred while trying to close the CSV file: "
+										+ e.getMessage());
+					}
+				}
+				else {
+					addCSVFileByQName(reporter, getTarget().getOutput(), instances);
```
stempler, line 123: As far as I can see closing the output stream is still not handled here.

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/writer/internal/CSVInstanceWriter.java
stempler: You can omit the explicit call to `close()` since that is done by the try-with-resources block automatically.

stempler (approved):

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/writer/internal/CSVInstanceWriter.java
emanuelaepure10: I  know that is optional to have it or not. I can remove it.

PR #1060, opened 2023-09-04
fixup! fix: For Shapefile writer recognize geometries for feature types

emanuelaepure10:
Shapefile files are correctly generated by the sending the CRS to the Geometry Finder.

ING-3974

### DIFF
--- a/io/plugins/eu.esdihumboldt.hale.io.shp/src/eu/esdihumboldt/hale/io/shp/writer/ShapefileInstanceWriter.java
+++ b/io/plugins/eu.esdihumboldt.hale.io.shp/src/eu/esdihumboldt/hale/io/shp/writer/ShapefileInstanceWriter.java
@@ -26,6 +26,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashMap;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
@@ -128,7 +129,9 @@ public class ShapefileInstanceWriter extends AbstractGeoInstanceWriter {
 				String cpgFileName = filePath + "/" + f + ShapefileConstants.CPG_EXTENSION;
 				writeCodePageFile(cpgFileName);
 			}
-
+			if (filesWritten.isEmpty()) {
+				reporter.warn("No file has been exported because there was no geometry found.");
+			}
 			reporter.setSuccess(true);
 		} catch (Exception e) {
 			reporter.error(new IOMessageImpl(e.getMessage(), e));
@@ -206,7 +209,7 @@ public class ShapefileInstanceWriter extends AbstractGeoInstanceWriter {
 
 		Map<String, Map<String, SimpleFeatureTypeBuilder>> schemaBuilderMap = new HashMap<String, Map<String, SimpleFeatureTypeBuilder>>();
 
-		List<String> missingGeomsForSchemas = new ArrayList<String>();
+		LinkedHashSet<String> missingGeomsForSchemas = new LinkedHashSet<String>();
 		try (ResourceIterator<Instance> it = instances.iterator()) {
 			while (it.hasNext() && !progress.isCanceled()) {
 				Instance instance = it.next();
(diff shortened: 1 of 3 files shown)


io/plugins/eu.esdihumboldt.hale.io.shp/src/eu/esdihumboldt/hale/io/shp/writer/ShapefileInstanceWriter.java
```diff
@@ -373,7 +374,7 @@ private String truncatePropertyName(String propName) {
 	private List<GeometryProperty<?>> traverseInstanceForGeometries(Instance instance) {
 		// find geometries in the schema.
 		InstanceTraverser traverser = new DepthFirstInstanceTraverser(true);
-		GeometryFinder geoFind = new GeometryFinder(null);
+		GeometryFinder geoFind = new GeometryFinder(getTargetCRS());
```
stempler, line 377: Looking at the [code](https://github.com/halestudio/hale/blob/01a8d220645900b77cd7a1249aa59ab6a0273984/common/plugins/eu.esdihumboldt.hale.common.instance/src/eu/esdihumboldt/hale/common/instance/geometry/GeometryFinder.java#L83) this will assign the CRS to a geometry encountered w/o CRS.
A CRS assignment should not be done by the writer, but when the data is read. The target CRS is intended to convert the data to, but if the data has not CRS, it can't be converted.

stempler (dismissed):

io/plugins/eu.esdihumboldt.hale.io.shp/src/eu/esdihumboldt/hale/io/shp/writer/ShapefileInstanceWriter.java
stempler: > Than could be possible that the project that we are testing doesn't have a CRS?

Looks like it - I guess this is not contained in a unit test but something you test manually in the UI?
You can check the source data by example in the source data view and can see there also if for a geometry CRS information is included or not.

stempler: Also, only if there is a CRS instances can appear on the map view.

emanuelaepure10: I'm testing it from the UI. In the source data I cannot see any information about CRS. The project attached to the issue is pretty big, maybe is hidden somewhere and I cannot see it.

All the other tests related to shp are passing successfully.

stempler: If you can share your test project somewhere (e.g. Box) I could have a look.

stempler: Provided feedback in the issue

emanuelaepure10: I have keep the missingGeomsForSchemas list just to add into the reporter the warning that the geometry is missing. 
We can also decide to remove it and just leave the warning in the UI that informs the user that "...Feature types with no geometry will not be exported".

io/plugins/eu.esdihumboldt.hale.io.shp/src/eu/esdihumboldt/hale/io/shp/writer/ShapefileInstanceWriter.java
```diff
@@ -234,7 +235,6 @@ private Map<String, Map<String, SimpleFeatureType>> createFeatureType(
 
 			if (missingGeomsForSchemas.contains(schemaEntry.getKey())) {
 				reporter.warn("No geometry found for " + schemaEntry.getKey());
-				continue;
```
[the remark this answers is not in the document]
emanuelaepure10: > 1. if there are some features with geometry and some without for a feature type, only the ones with geometry end up in the target file?

ONLY the features with geometry end up in the target file.

> if there are no features with geometry for a feature type, is an empty target file created for the type, or none at all?

NO file will be created for the features without a geometry.

> or will an error be thrown?

In the last page of the wizard there is a warning `"Warning! Feature types with no geometry will not be exported"` and if there is not geometry presented at all in any of the features that at the end of the export a warning message of the reported `IOReporter `will be shown informing the user that `"No file has been exported because there was no geometry found."`

An example would be 
image

stempler (approved):

PR #1062, opened 2023-10-04
fix: export transformed data leads to a shift of the attribute values

emanuelaepure10:
No shift in the attributes happens when exporting a transformed data to shp-file.

ING-4052

### DIFF
--- a/io/plugins/eu.esdihumboldt.hale.io.shp.test/src/eu/esdihumboldt/hale/io/shp/ShapefileInstanceWriterTest.groovy
+++ b/io/plugins/eu.esdihumboldt.hale.io.shp.test/src/eu/esdihumboldt/hale/io/shp/ShapefileInstanceWriterTest.groovy
@@ -333,6 +333,75 @@ class ShapefileInstanceWriterTest {
 	}
 
 	@Test
+	void testSingleGeometryWithMissingProperties() {
+
+		Schema schema = new SchemaBuilder().schema {
+			city {
+				name(String)
+				region(String)
+				population(Integer)
+				location(GeometryProperty)
+			}
+		}
+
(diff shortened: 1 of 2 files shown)


stempler (commented):
Would be great if you can add a unit test that reproduces the original issue if that is easily done.

Please use `fix:` as prefix for the commit message instead of `bug:`. Thanks!

emanuelaepure10:
I knew that you are going to ask about the test, but I actually wanted to ask you if you have idea how to have some/couple of instances out of all of the test that are not transformed.  

What I mean is this
image
to have 13 instances of "arokstatus" in the source data and none in the transformed data.

Thank you

stempler:
> to have 13 instances of "arokstatus" in the source data and none in the transformed data.

I guess that is a matter of the transformation project used, which transformations functions are used there and if there are any errors during the transformation.

If this is about test data, it is probably easier to just create test data within the test, which is quite easy using the schema and instance builders, e.g. similar to [here](https://github.com/halestudio/hale/blob/5fdc62e86284c9aadd4b4e6d15d0f6c77f1b36ec/io/plugins/eu.esdihumboldt.hale.io.shp.test/src/eu/esdihumboldt/hale/io/shp/ShapefileInstanceWriterTest.groovy#L280).

Were you able to identify what kind of data caused the issue?

Did it happen when

- the first instance to write has null for an attribute value?
- all instance have null for an attribute value?
- any instance has null for an attribute value?

emanuelaepure10:
The project doesn't have an error. It is intended to transform just a number of instances, not all.

The problem arise when in one instance a property has null value.

With all the projects I have tried the exported .shp file has been correctly exported after removing the check on the null value/property.

My question was about creating a test that would replicate the real case.

stempler (approved):

PR #1063, opened 2023-10-09
fix: add schema location for each instance

emanuelaepure10:
Add all schema locations to GML writer data exports containing top level objects from multiple schemas

ING-3890

### DIFF
--- a/io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriter.java
+++ b/io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriter.java
@@ -31,6 +31,7 @@ import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
+import java.util.LinkedHashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
@@ -1151,6 +1152,45 @@ public class StreamGmlWriter extends AbstractGeoInstanceWriter
 
 			}
 
+			// get additional schema location from all the imports
+			// for all the instances
+			ResourceIterator<Instance> resourceIterator = instances.iterator();
+			DefaultInstanceCollection itInstancenew = new DefaultInstanceCollection(
+					new ArrayList<Instance>());
+			try {
+				Set<String> schemasToBeAdded = new LinkedHashSet<String>();
+				while (resourceIterator.hasNext() && !progress.isCanceled()) {
+					Instance instance = resourceIterator.next();
+					itInstancenew.add(instance);
+
+					TypeDefinition type = instance.getDefinition();
+					if (!type.getName().getNamespaceURI().equals(targetIndex.getNamespace())
+							&& !additionalSchemas.containsKey(type.getName().getNamespaceURI())) {
+
+						schemasToBeAdded.add(type.getName().getNamespaceURI());
+					}
+				}
+
+				if (!schemasToBeAdded.isEmpty()) {
+					for (String namespaceToBeAdded : schemasToBeAdded) {
+						targetIndex.getSchemaImports().entrySet().stream()
+								.filter(entry -> namespaceToBeAdded.equals(entry.getValue()))
+								.findFirst().ifPresent(entry -> {
+									try {
+										URI instanceSchemaLoc = stripFragment(
+												new URI(entry.getKey()));
+										addValidationSchema(namespaceToBeAdded,
+												() -> instanceSchemaLoc, null);
+									} catch (URISyntaxException e) {
+										reporter.error("Invalid URI syntax: " + e.getMessage());
+									}
+								});
+					}
+				}
+			} finally {
+				resourceIterator.close();
+			}
+
 			// additional schema namespace prefixes
 			for (Entry<String, String> schemaNs : additionalSchemaPrefixes.entrySet()) {
 				GmlWriterUtil.addNamespace(writer, schemaNs.getKey(), schemaNs.getValue());
@@ -1202,7 +1242,7 @@ public class StreamGmlWriter extends AbstractGeoInstanceWriter
 			writeAdditionalElements(writer, containerDefinition, reporter);
 
 			// write the instances
-			ResourceIterator<Instance> itInstance = instances.iterator();
+			ResourceIterator<Instance> itInstance = itInstancenew.iterator();
 			try {
 				Map<TypeDefinition, DefinitionPath> paths = new HashMap<TypeDefinition, DefinitionPath>();
 
@@ -1988,4 +2028,5 @@ public class StreamGmlWriter extends AbstractGeoInstanceWriter
 	protected EnumWindingOrderTypes getDefaultWindingOrder() {
 		return EnumWindingOrderTypes.counterClockwise;
 	}
+
 }
--- a/io/plugins/eu.esdihumboldt.hale.io.xsd/src/eu/esdihumboldt/hale/io/xsd/reader/XmlSchemaReader.java
+++ b/io/plugins/eu.esdihumboldt.hale.io.xsd/src/eu/esdihumboldt/hale/io/xsd/reader/XmlSchemaReader.java
@@ -21,7 +21,6 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.text.MessageFormat;
 import java.util.Collection;
-import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
@@ -360,8 +359,7 @@ public class XmlSchemaReader extends AbstractSchemaReader {
 		groupCounter = new TObjectIntHashMap<String>();
 
 		// Map schema locations to target namespaces
-		Map<String, String> imports = new HashMap<>();
-		imports.put(location.toString(), namespace);
+		index.getSchemaImports().put(location.toString(), namespace);
 
 		// load XML Schema schema (for base type definitions)
 		try {
@@ -384,7 +382,7 @@ public class XmlSchemaReader extends AbstractSchemaReader {
 			_log.error("Exception while loading XML Schema schema", e);
 		}
 
-		loadSchema(location.toString(), xmlSchema, imports, progress, true);
+		loadSchema(location.toString(), xmlSchema, progress, true);
 
 		groupCounter.clear();
 
@@ -526,7 +524,7 @@ public class XmlSchemaReader extends AbstractSchemaReader {
 	 *            declared here should be flagged mappable
 	 */
 	protected void loadSchema(String schemaLocation, XmlSchema xmlSchema,
-			Map<String, String> imports, ProgressIndicator progress, boolean mainSchema) {
+			ProgressIndicator progress, boolean mainSchema) {
 		String namespace = xmlSchema.getTargetNamespace();
 		if (namespace == null) {
 			namespace = XMLConstants.NULL_NS_URI;
@@ -695,22 +693,29 @@ public class XmlSchemaReader extends AbstractSchemaReader {
 				XmlSchema importedSchema = imp.getSchema();
 				String location = importedSchema.getSourceURI();
 				String targetNamespace = importedSchema.getTargetNamespace();
-				if (!imports.containsKey(location)) { // only add schemas that
-														// were not already
-														// added
-					boolean addedBefore = imports.entrySet().stream().anyMatch(e -> {
-						// Consider two schemas to be identical if their schema
-						// locations differ only in the scheme part (e.g. "http"
-						// and "https") and they have the same target namespace.
-						return targetNamespace.equals(e.getValue())
-								&& schemeIndependentEquals(location, e.getKey());
-					});
+				if (!index.getSchemaImports().containsKey(location)) { // only
+																		// add
+					// schemas
+					// that
+					// were not already
+					// added
+					boolean addedBefore = index.getSchemaImports().entrySet().stream()
+							.anyMatch(e -> {
+								// Consider two schemas to be identical if their
+								// schema
+								// locations differ only in the scheme part
+								// (e.g. "http"
+								// and "https") and they have the same target
+								// namespace.
+								return targetNamespace.equals(e.getValue())
+										&& schemeIndependentEquals(location, e.getKey());
+							});
 
 					if (!addedBefore) {
 						// place a marker in the map to prevent loading the
 						// location in the call to loadSchema
-						imports.put(location, targetNamespace);
-						loadSchema(location, importedSchema, imports, progress,
+						index.getSchemaImports().put(location, targetNamespace);
+						loadSchema(location, importedSchema, progress,
 								mainSchema && imp instanceof XmlSchemaInclude);
 						// is part of main schema if it's a main schema include
 					}
--- a/io/plugins/eu.esdihumboldt.hale.io.xsd/src/eu/esdihumboldt/hale/io/xsd/model/XmlIndex.java
+++ b/io/plugins/eu.esdihumboldt.hale.io.xsd/src/eu/esdihumboldt/hale/io/xsd/model/XmlIndex.java
@@ -67,6 +67,11 @@ public class XmlIndex extends DefaultSchema {
 	private final BiMap<String, String> prefixes = HashBiMap.create();
 
 	/**
+	 * Schema imports
+	 */
+	private final Map<String, String> schemaImports = new HashMap<>();
+
+	/**
 	 * @see DefaultSchema#DefaultSchema(String, URI)
 	 */
 	public XmlIndex(String namespace, URI location) {
@@ -147,4 +152,14 @@ public class XmlIndex extends DefaultSchema {
 		return prefixes;
 	}
 
+	/**
+	 * @return the imports
+	 */
+
+	/**
+	 * @return the schemaImports
+	 */
+	public Map<String, String> getSchemaImports() {
+		return schemaImports;
+	}
 }


emanuelaepure10:
Retest this please

io/plugins/eu.esdihumboldt.hale.io.xsd/src/eu/esdihumboldt/hale/io/xsd/reader/XmlSchemaReader.java
```diff
@@ -709,8 +709,8 @@ else if (item instanceof XmlSchemaNotation) {
 					if (!addedBefore) {
 						// place a marker in the map to prevent loading the
 						// location in the call to loadSchema
-						imports.put(location, targetNamespace);
-						loadSchema(location, importedSchema, imports, progress,
+						index.getImports().put(location, targetNamespace);
```
stempler, line 712: You basically replaced the map of import schema locations to namespace that was previously just kept internally.

Maybe you remember our discussion, that for this task, where we need to identify a location for a namespace, we need to inverse association with the problem to decide on which location would be the best / right one to use. For that we said using the first import encountered as the "main" import for the namespace would be a good heuristic.

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriter.java
```diff
@@ -1156,6 +1157,41 @@ public URI getLocation() {
 				GmlWriterUtil.addNamespace(writer, schemaNs.getKey(), schemaNs.getValue());
 			}
 
+			// get additional schema location from the instances
+			ResourceIterator<Instance> resourceIterator = instances.iterator();
```
stempler, line 1161: As mentioned in the ticket, iterating over all instances twice is something that we want to avoid.
This would break the possibility to stream instances to this writer without either reading the instances twice or using a temporary storage.

Do you have any reason to believe that it is certain, that only a schema location for the instance/feature types namespaces are required? Why would e.g. a property in one of those types that uses another namespace from a base schema not require an additional schema location for validation?
Did you test this? I guess the Buildings2D could be a candidate to test this.

If it is certain that only the feature type namespace is required to be covered and not any others that appear in the file from the features' elements on, then I would make the behavior as you implemented it now an optional behavior, to still keep the possibility to stream instances using the GML writer.

Is there something that speaks in general against including all the "main" import locations of the schema as we had discussed as simplest solution?

emanuelaepure10: I have tried the solution to add just the "main" imports from the schema location...but this doesn't work. Doesn't work just because one can decide to have a feature types with a schemaLocation that is not the "main", but it's a secondary level or more of the main schema location. Actually the example attached to the issue contains exactly this. I haven't committed this solution because I don't think that would cover all the cases. 

I still think that the previous solution was the best (actually the one committed here), in this case we would include exactly the necessary schemaLocation taken from the instances. and this cannot be done without streaming the instances once more.  Streaming 500 instances to get the couples namespace-schemaLocation takes no time. 

If you have any other idea please let me know
Thank you

stempler: > I have tried the solution to add just the "main" imports from the schema location...but this doesn't work. Doesn't work just because one can decide to have a feature types with a schemaLocation that is not the "main", but it's a secondary level or more of the main schema location.

What I referred to in our discussion as main import for a namespace is the schema location that is likely best suited to be used as schema location for the respective namespace. The same namespace may be distributed across multiple files and different files might be referenced from different places in the schema hierarchy.

Collecting all of these would still cover all imported namespaces, with one "main" schema location for each namespace.

> Do you have any reason to believe that it is certain, that only a schema location for the instance/feature types namespaces are required? Why would e.g. a property in one of those types that uses another namespace from a base schema not require an additional schema location for validation?
Did you test this? I guess the Buildings2D could be a candidate to test this.

Do you have any feedback on this? If the problem happens if a namespace that is not part of the schema locations is introduced at feature type level, I would not see a reason why there shouldn't be a similar problem when the same happens at a level further down in the XML structure.

emanuelaepure10: > Do you have any feedback on this? If the problem happens if a namespace that is not part of the schema locations is introduced at feature type level, I would not see a reason why there shouldn't be a similar problem when the same happens at a level further down in the XML structure.
So, you have to comply your data with some feature types, each of it is part of ONE single XML schema. As the feature types is part of a schema, how could be possible to have the feature type part of another schema so another namespace? 
Yes, there are rare cases, such as Building, Building2D and Building3D where all of them contain 2 elements with the same name: Building and BuildingPart, but these are coming from Building and both Building2D and Building3D import the schema of Building, so you will always manage to get the right schema location. 
 
From the feedback I have the namespaces used in the transformed feature types will always be part of the imported schema as a direct import or lower import in the hierarchy of the imports.

emanuelaepure10: I got this answer form @Johanna, just confirming that:

```
It is possible to have target types that are not defined in the target schema directly but in one of the schemas imported in it (or one of the schemas imported in the imported ones and so on).
If the type would not be defined in the schema or one of its imported ones, there would be no way for the application to determine the schema of the type, so that is not possible.
```

stempler: I have the feeling we talk about two different things here.

The whole ticket is about the issue with the validation in the INSPIRE validator and that for some reason, if there are feature elements that are in a namespace that is not explicitly mentioned in the schema location (but is included in a schema import down the line), the validation fails.

So my question is, if this can happen on the level of the feature element, why not also at a level further down, if there is also a new namespace used that has no explicit schema location associated in the document root.

We don't know why the validation requires the extra schema location, so my point is how do we know that it doesn't require it in similar cases on a different level in the XML file.

In my view, if we don't know the exact rules why it does not work, it would be safer to just add explicit locations for all potential namespaces instead of just the ones where we think it is probably enough. The later is what we had done for the root element's namespace, and it was not enough, otherwise there wouldn't be this task.

One example here:

```
<?xml version='1.0' encoding='UTF-8'?>
<wfs:FeatureCollection xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.opengis.net/wfs/2.0 http://schemas.opengis.net/wfs/2.0/wfs.xsd http://www.opengis.net/gml/3.2 http://schemas.opengis.net/gml/3.2.1/gml.xsd http://inspire.ec.europa.eu/schemas/bu-core2d/4.0 https://geodienste.hamburg.de/HH_WFS_INSPIRE_Gebaeude_2D_ALKIS?SERVICE=WFS&amp;VERSION=2.0.0&amp;REQUEST=DescribeFeatureType&amp;OUTPUTFORMAT=application%2Fgml%2Bxml%3B+version%3D3.2&amp;TYPENAME=bu-core2d:BuildingPart&amp;NAMESPACES=xmlns(bu-core2d,http%3A%2F%2Finspire.ec.europa.eu%2Fschemas%2Fbu-core2d%2F4.0)" xmlns:wfs="http://www.opengis.net/wfs/2.0" timeStamp="2023-10-11T15:06:23Z" xmlns:gml="http://www.opengis.net/gml/3.2" numberMatched="unknown" numberReturned="0">
  <!--NOTE: numberReturned attribute should be 'unknown' as well, but this would not validate against the current version of the WFS 2.0 schema (change upcoming). See change request (CR 144): https://portal.opengeospatial.org/files?artifact_id=43925.-->
  <wfs:member>
    <bu-core2d:BuildingPart xmlns:bu-core2d="http://inspire.ec.europa.eu/schemas/bu-core2d/4.0" gml:id="BuildingPart_DEHHALKA8560009i">
      <gml:identifier codeSpace="http://inspire.ec.europa.eu/ids">https://registry.gdi-de.org/id/de.hh/e33c193f-278f-4118-a581-cc8e8420abc3/BuildingPart_DEHHALKA8560009i</gml:identifier>
      <bu-base:beginLifespanVersion xmlns:bu-base="http://inspire.ec.europa.eu/schemas/bu-base/4.0">2022-11-03T10:24:52Z</bu-base:beginLifespanVersion>
      <bu-base:conditionOfConstruction xmlns:bu-base="http://inspire.ec.europa.eu/schemas/bu-base/4.0" xsi:nil="true" nilReason="true"/>
      <bu-base:inspireId xmlns:bu-base="http://inspire.ec.europa.eu/schemas/bu-base/4.0">
```

The `bu-base` prefix and the corresponding namespace are first used on a property of the `BuildingPart`. There is no explicit schema location for that namespace. How would we know that something like that would not be a problem, if the assumption is that if we replace the defined schema location for `bu-core2d` by something that uses a different target namespace but imports `bu-core2d`?

emanuelaepure10: I have the solution, the problem is much simpler than we imagine.

If you have in a file that you want to validate feature type A coming from application schema X and feature type B coming from application schema Y, where Y is an import into X, than  for the validator is enough to add in the schemaLocation just the application schema X BUT ALSO  

**`"http://www.opengis.net/gml/3.2 http://schemas.opengis.net/gml/3.2.1/gml.xsd"`**

this helps the validator to retrieve the root element definition.

Please check [this comment of Fabio Vinci](https://github.com/INSPIRE-MIF/helpdesk-validator/issues/223#issuecomment-588725257)

The `"http://www.opengis.net/gml/3.2 http://schemas.opengis.net/gml/3.2.1/deprecatedTypes.xsd"` should be replace by the above pair because the `gml.xsd `contains `deprecatedTypes.xsd`.

@JohannaOtt would you have anything else to add?

Thank you in advance

JohannaOtt:
The ticket link in the commit message is not correct (ING-380 instead of ING-3980)

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriter.java
JohannaOtt: > So my question is, if this can happen on the level of the feature element, why not also at a level further down, if there is also a new namespace used that has no explicit schema location associated in the document root.
> 
> We don't know why the validation requires the extra schema location, so my point is how do we know that it doesn't require it in similar cases on a different level in the XML file.
> 
> In my view, if we don't know the exact rules why it does not work, it would be safer to just add explicit locations for all potential namespaces instead of just the ones where we think it is probably enough. The later is what we had done for the root element's namespace, and it was not enough, otherwise there wouldn't be this task.

My 2 cents on this: I get your point but I think that this is not an issue as otherwise, none of our Buildings data would ever have validated and my feeling is that we would be aware of that. Also, the ticket is explictely about feature types on top level, so at some point in the testing when writing the ticket, Kate must have come to the conclusion that only schemaLocations for the feature types used on top level must be added and not for types used in a nested structure (and I would assume from that; also not for elements/attributes used in nested structures). But I did not check the validator code or anything the like, so this argument is only based on that we do not get validation issues for that.


> If you have in a file that you want to validate feature type A coming from application schema X and feature type B coming from application schema Y, where Y is an import into X, than for the validator is enough to add in the schemaLocation just the application schema X BUT ALSO
> 
> "http://www.opengis.net/gml/3.2 http://schemas.opengis.net/gml/3.2.1/gml.xsd"
> 
> this helps the validator to retrieve the root element definition.
> 
> Please check https://github.com/INSPIRE-MIF/helpdesk-validator/issues/223#issuecomment-588725257

As discussed in our call last week, I do not think that this will solve our solution. IMO, adding the gml schemaLocation will help in cases where the validator is not able to validate an element using the gml namespace (error message according to [this issue](https://github.com/INSPIRE-MIF/helpdesk-validator/issues/223): "Cannot find the declaration of element 'gml:FeatureCollection'").
In our case, other namespaces & their schemaLocations are relevant - for example the om schemaLocation missing is resulting in an error message "Invalid content was found starting with element '{"http://www.opengis.net/om/2.0":OM_Observation}'. One of '{"http://www.opengis.net/gml/3.2":AbstractFeature}' is expected.".
So for all feature types used (according to the ticket on top level), the validator will complain about missing decplarations/features not being expected with specific error messages.

Maybe interesting: According to [this comment](https://github.com/INSPIRE-MIF/helpdesk-validator/issues/223#issuecomment-698892840), hale studio is adding the schemaLocations when the schemas are explicitely reloaded after the respective types from imported schemas have been added to the mappable target types but I never tested it and I think it is surprising. But if that is true, it might help to find the proper solution. If not, please ignore this part.

emanuelaepure10: For the project we have in the [attached to the issue](https://wetransform.atlassian.net/browse/ING-3980)
I get always the same [errors in the validator](https://inspire.ec.europa.eu/validator/test-run/details.html?id=EID41de63ac-a10a-47c9-8d54-4808af571a3e)
by adding in the schemaLocation:

```
http://inspire.ec.europa.eu/schemas/so/4.0 http://inspire.ec.europa.eu/schemas/so/4.0/Soil.xsd 
http://inspire.ec.europa.eu/schemas/ompr/3.0 http://inspire.ec.europa.eu/schemas/ompr/3.0/Processes.xsd
```
OR
```
http://inspire.ec.europa.eu/schemas/so/4.0 http://inspire.ec.europa.eu/schemas/so/4.0/Soil.xsd 
http://www.opengis.net/gml/3.2 http://schemas.opengis.net/gml/3.2.1/gml.xsd 
```
OR 
```
http://inspire.ec.europa.eu/schemas/so/4.0 http://inspire.ec.europa.eu/schemas/so/4.0/Soil.xsd 
http://inspire.ec.europa.eu/schemas/ompr/3.0 http://inspire.ec.europa.eu/schemas/ompr/3.0/Processes.xsd 
http://www.opengis.net/gml/3.2 http://schemas.opengis.net/gml/3.2.1/gml.xsd 
```

Any idea how to go ahead?
Thank you

JohannaOtt: > I get always the same [errors in the validator](https://inspire.ec.europa.eu/validator/test-run/details.html?id=EID41de63ac-a10a-47c9-8d54-4808af571a3e)

I would say none of the errors are related to missing/wrong schema locations (no error message "Invalid content was found starting with element [...]". IMO, they are related to the CRS missing the required prefix, the alignment project containing errors and the data not being complete.

stempler: @emanuelaepure10 

> I have the solution, the problem is much simpler than we imagine.

I mean if that solves the problem with the validator then that should be easy to test, using the dataset that currently fails with `Invalid content was found starting with element '{"http://www.opengis.net/om/2.0":OM_Observation}'. One of '{"http://www.opengis.net/gml/3.2":AbstractFeature}' is expected.` and changing the schema location.

> The `"http://www.opengis.net/gml/3.2 http://schemas.opengis.net/gml/3.2.1/deprecatedTypes.xsd"` should be replace by the above pair because the gml.xsd contains deprecatedTypes.xsd.

Not really a strong argument though because `deprecatedTypes.xsd` also includes `gml.xsd`.

If this solves the problem however, this would be simple to do by (1) collecting the information on the main schema location for a namespace in the XML index as we had discussed initially and (2) in the logic where we add the location for the root element prefer the main schema location for that namespace as collected before the location where the element is defined.

@JohannaOtt 

> I get your point but I think that this is not an issue as otherwise, none of our Buildings data would ever have validated and my feeling is that we would be aware of that. Also, the ticket is explictely about feature types on top level, so at some point in the testing when writing the ticket, Kate must have come to the conclusion that only schemaLocations for the feature types used on top level must be added and not for types used in a nested structure (and I would assume from that; also not for elements/attributes used in nested structures). But I did not check the validator code or anything the like, so this argument is only based on that we do not get validation issues for that.

I don't see why XML schema validation should treat GML feature type elements in a special way, or elements that are at the third level in the XML compared to elements at the fourth level.
Since we don't know what the rules are for the validation to work, my proposal was to just add additional schema locations for every namespace imported from the main schema so we catch all possible cases where it might complain, even if that may mean that we may have some schema locations in the file that are not needed.

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriter.java
```diff
@@ -1151,6 +1152,45 @@ public URI getLocation() {
...
+			// get additional schema location from all the imports
+			// for all the instances
+			ResourceIterator<Instance> resourceIterator = instances.iterator();
+			DefaultInstanceCollection itInstancenew = new DefaultInstanceCollection(
+					new ArrayList<Instance>());
+			try {
+				Set<String> schemasToBeAdded = new LinkedHashSet<String>();
+				while (resourceIterator.hasNext() && !progress.isCanceled()) {
+					Instance instance = resourceIterator.next();
+					itInstancenew.add(instance);
```
stempler, line 1164: Like this every instance will be kept in memory, which will be a problem if there is a lot of data.

emanuelaepure10: If you refer to schemasToBeAdded from what I've seen till now this list could have maximum couple of elements. 
If you refer to the resourceIterator, I'm closing it after using it.
I would love to use the instances.iterator() twice also not to have the itInstancenew, but somehow for the tests accessing it second time returns always null. Instead from the HS works as it sholud.
Do you have any better idea?
Thank you

stempler: I refer to all instances being added to a collection during the process and thus remaining in memory.

If the instances need to be traversed multiple times, then the approach as before iterating two times is the correct one, but the implementation needs to state that it can't process the instances in one iteration (see `isPassthrough`method).
However, as mentioned in the other discussion if we do the two iterations here, I would want that behavior to not be the default and enabled with a configuration parameter, to by default keep the possibility to stream the results:

> If it is certain that only the feature type namespace is required to be covered and not any others that appear in the file from the features' elements on, then I would make the behavior as you implemented it now an optional behavior, to still keep the possibility to stream instances using the GML writer.

stempler:
@emanuelaepure10 Can't this be closed, as it turns out that the extra information in the XML file that this would add is not required after all?

PR #1068, opened 2023-10-25
feat: support loading schema from Json data

stempler:
First version that comes with some known restrictions:

- Parameters for the reader are not configurable in the UI, so e.g. the read mode can't be selected and only Json arrays and GeoJson FeatureCollections are supported in the UI at the moment.
- The schema reader only considers the first 50 objects encountered in the Json data. This is currently not configurable.
- Nested arrays (array of arrays) are not supported

ING-4046


stempler:
Merged so I can use the changes downstream, feedback naturally still welcome!

PR #1075, opened 2023-11-02
feat: configure decimal places in JsonInstanceWriter

kapil-agnihotri:
This configuration can be added in
`<export-config>` element of `.halex`
project file as
`<setting name="numOfDecimalPlaces">14</setting>`
ING-3332

### DIFF
--- a/io/plugins/eu.esdihumboldt.hale.io.json/plugin.xml
+++ b/io/plugins/eu.esdihumboldt.hale.io.json/plugin.xml
@@ -30,6 +30,18 @@
          <contentType
                ref="eu.esdihumboldt.hale.io.json.gzip">
          </contentType>
+         <providerParameter
+               description="Number of decimal places to display."
+               label="Number of decimal places"
+               name="eu.esdihumboldt.hale.io.json.numOfDecimalPlaces">
+            <parameterBinding
+                  class="java.lang.Integer">
+            </parameterBinding>
+            <valueDescriptor
+                  default="7"
+                  defaultDescription="Default decimal places to 7">
+            </valueDescriptor>
+         </providerParameter>
       </provider>
       <provider
             class="eu.esdihumboldt.hale.io.json.GeoJSONInstanceWriter"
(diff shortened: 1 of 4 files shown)


io/plugins/eu.esdihumboldt.hale.io.json/src/eu/esdihumboldt/hale/io/json/JsonInstanceWriter.java
```diff
@@ -71,6 +70,14 @@ public boolean isCancelable() {
 	@Override
 	protected IOReport execute(ProgressIndicator progress, IOReporter reporter)
 			throws IOProviderConfigurationException, IOException {
+		Integer numOfDecimalPlaces = getParameter("numOfDecimalPlaces").as(Integer.class);
```
stempler, line 73: The new parameter should also be added to the provider's definition in the extension point to document it.

kapil-agnihotri: Thank you Simon for the feedback. I have updated provider's definition.

io/plugins/eu.esdihumboldt.hale.io.json/plugin.xml
```diff
@@ -30,6 +30,19 @@
          <contentType
                ref="eu.esdihumboldt.hale.io.json.gzip">
          </contentType>
+         <providerParameter
+               default="2"
```
stempler, line 34: This looks strange - I don't think there is a `default` attribute on that element. Did you use the Eclipse editor to add this?

stempler (approved):

io/plugins/eu.esdihumboldt.hale.io.json/plugin.xml
kapil-agnihotri: True, I initially edited the xml  but then later modified it from the editor. Looks like this left the initial changes unaltered. Thank you Simon.

PR #1076, opened 2023-11-03
Support for reading multiple sheets from an Excel file at once

stempler:
A new setting was introduced that enables loading multiple sheets when reading the Excel file. Type names are tried to be determined based on match with the sheet name, if no other information is provided.

An additional new setting for the Excel reader was introduced to be able to control in detail how an individual sheet should be loaded.
    
Currently support is only provided in headless mode. In the hale studio UI it is not yet supported to load multiple sheets.

Also includes fixes:
    
- XLSInstanceWriter did not state correctly if it requires multiple iterations over the instance collection
- reading .xlsx files with hale studio, which did not work before because always .xls format was assumed
- test for writing multiple sheets

ING-4062


io/plugins/eu.esdihumboldt.hale.io.xls/src/eu/esdihumboldt/hale/io/xls/reader/SheetSettings.java
```diff
@@ -0,0 +1,159 @@
...
+	private Integer skipLines;
+
+	/**
+	 * Create sheet settings
+	 * 
+	 * @param identifiedByName optional sheet name to identify a sheet
+	 * @param identifiedByIndex optional sheet index to identify a sheet
+	 */
+	public SheetSettings(String identifiedByName, Integer identifiedByIndex) {
+		super();
```
kapil-agnihotri, line 46: There is no parent class. Do we need to call `super()` here as well?

io/plugins/eu.esdihumboldt.hale.io.xls/src/eu/esdihumboldt/hale/io/xls/reader/XLSInstanceReader.java
```diff
@@ -83,83 +74,121 @@ public boolean isCancelable() {
...
+			}
 
-			skipNlines = getParameter(CommonSchemaConstants.PARAM_SKIP_N_LINES).as(Integer.class);
+			if (type == null) {
+				// try display name (since this is used when writing to Excel)
+				getSourceSchema().getMappingRelevantTypes().stream()
+						.filter(t -> flName.equals(t.getDisplayName())).findFirst().orElse(null);
+			}
 
+			if (type == null) {
```
kapil-agnihotri, line 143: `type` is not set in the above `filter()` operation it would still be `null` here if it was `null` above. Also is it intended to leave the result of this `filter()` operation?

io/plugins/eu.esdihumboldt.hale.io.xls/src/eu/esdihumboldt/hale/io/xls/reader/XLSInstanceReader.java
```diff
@@ -83,83 +74,121 @@ public boolean isCancelable() {
...
-		if (skipType == null) {
+			// exact name
+			if (qname != null) {
+				type = getSourceSchema().getType(qname);
+			}
 
-			skipNlines = getParameter(CommonSchemaConstants.PARAM_SKIP_N_LINES).as(Integer.class);
+			if (type == null) {
+				// try display name (since this is used when writing to Excel)
+				getSourceSchema().getMappingRelevantTypes().stream()
```
kapil-agnihotri, line 139: The result of this expression is not collected in any variable. Is it expected?

io/plugins/eu.esdihumboldt.hale.io.xls/src/eu/esdihumboldt/hale/io/xls/writer/XLSInstanceWriter.java
```diff
@@ -125,12 +126,52 @@ else if (getContentType().getId().equals("eu.esdihumboldt.hale.io.xls.xlsx")) {
...
+		if (exportTypes != null) {
+			List<TypeDefinition> types = new ArrayList<TypeDefinition>();
+			String[] splitExportType = exportTypes.split(",");
+			for (String featureType : splitExportType) {
+				QName typeName = QName.valueOf(featureType);
+				TypeDefinition type = getTargetSchema().getType(typeName);
+
+				if (type == null) {
+					// try matching local name
+					type = getTargetSchema().getMappingRelevantTypes().stream()
```
kapil-agnihotri, line 144: Maybe move this common code from reader and writer class to a common utility class?

io/plugins/eu.esdihumboldt.hale.io.xls/src/eu/esdihumboldt/hale/io/xls/writer/XLSInstanceWriter.java
```diff
@@ -125,12 +126,52 @@ else if (getContentType().getId().equals("eu.esdihumboldt.hale.io.xls.xlsx")) {
...
+							.filter(t -> typeName.getLocalPart().equals(t.getName().getLocalPart()))
+							.findFirst().orElse(null);
+				}
+
+				if (type == null) {
+					// try matching display name
+					type = getTargetSchema().getMappingRelevantTypes().stream()
```
kapil-agnihotri, line 151: Maybe move this common code from reader and writer class to a common utility class?

io/plugins/eu.esdihumboldt.hale.io.xls/src/eu/esdihumboldt/hale/io/xls/reader/XLSInstanceReader.java
stempler: Good catch! Thanks

io/plugins/eu.esdihumboldt.hale.io.xls/src/eu/esdihumboldt/hale/io/xls/reader/SheetSettings.java
stempler: Eclipse added it when generating the constructor. I guess there is also not really a reason to remove it.

io/plugins/eu.esdihumboldt.hale.io.xls.test/src/eu/esdihumboldt/hale/io/xls/test/reader/XLSReaderTest.java
```diff
@@ -339,6 +348,165 @@ public void testPropertyType() throws Exception {
...
+	 * 
+	 * @throws Exception if an error occurs
+	 */
+	@Test
+	public void testReadMultiSheetTypeSettingsNamed() throws Exception {
+		Schema schema = XLSReaderTestUtil.createMultiSheetExampleSchema();
+
+		ValueList settings = new ValueList();
+
+		// Note: we swap the tables to make sure auto-detection based on the
```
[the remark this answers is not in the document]
stempler: The auto-detection based on the name is not used, instead we say that sheet "city" should be associated with the "person" type and vice versa. We do this swap to be sure that the result of the matching is not just using the automated detection.

stempler (commented):
Thanks a lot for the feedback, will create a follow-up PR.

PR #1080, opened 2023-11-14
hale: Enable exporting feature types with no data

emanuelaepure10:
Users can select feature types without data for both: 

> Custom export configuration for hale-connect using Excel as output type 
> Export of transformed data as Excel from hale-studio

For the feature types without data there is the indication in the label that _[no data]_  is present. Has been reintroduced the button "**Exclude feature types without data**"

ING-4107

### DIFF
--- a/io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/writer/AbstractTableInstanceWriter.java
+++ b/io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/writer/AbstractTableInstanceWriter.java
@@ -147,6 +147,34 @@ public abstract class AbstractTableInstanceWriter extends AbstractInstanceWriter
 	} // close method
 
 	/**
+	 * Iterates over properties of the type definition and creates a map of the
+	 * given properties
+	 * 
+	 * @param definition given type to check
+	 * @param headerRow the current header row of the table
+	 * @return a map of properties with string of localpart of the QName of the
+	 *         property as key
+	 */
+	protected Map<String, Object> getPropertyMap(TypeDefinition definition,
+			List<String> headerRow) {
+		Iterable<QName> allProperties = DefinitionUtil.getAllProperties(definition).stream()
+				.map(def -> def.getName()).collect(Collectors.toList());
+		Map<String, Object> row = new HashMap<String, Object>();
+		for (QName qname : allProperties) {
+
+			String cellValue = "";
+			Object property = null;
+			if (shouldBeDisplayed(property)) {
+				cellValue = qname.getLocalPart();
+			}
+			addProperty(headerRow, row, property, cellValue);
+
+		}
+
+		return row;
+	}
+
+	/**
 	 * 
 	 * @param instance the actual instance
 	 * @param qNameOfTheInstance the qname of the instance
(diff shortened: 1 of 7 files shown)


io/plugins/eu.esdihumboldt.hale.io.xls.ui/src/eu/esdihumboldt/hale/io/xls/ui/XLSInstanceExportConfigurationPage.java
```diff
@@ -155,48 +153,30 @@ public void checkStateChanged(CheckStateChangedEvent event) {
...
-				public boolean isGrayed(Object element) {
-					if (!(element instanceof TypeDefinition))
-						return false;
-					return checkboxState(element);
-				}
+	/**
+	 * @param element element to be checked
+	 * @return true if the button cannot be selected
+	 */
+	private boolean checkboxState(Object element) {
```
stempler, line 165: Maybe rename the method to better reflect its new function (e.g. `hasData`?).
I think it would be better to change the type of `element` to `TypeDefinition` to avoid accidental wrong use.

io/plugins/eu.esdihumboldt.hale.io.xls/src/eu/esdihumboldt/hale/io/xls/writer/XLSInstanceWriter.java
```diff
@@ -170,14 +174,18 @@ private void addSheetByQName(boolean solveNestedProperties, InstanceCollection i
 			try {
 				instance = instanceIterator.next();
 			} catch (NoSuchElementException e) {
+				if (!ignoreEmptyFeaturetypes) {
+					Sheet sheet = workbook.createSheet(selectedTypeName.getDisplayName());
+					Row headerRow = sheet.createRow(0);
+					writeHeaderRow(headerRow, headerRowStrings);
```
stempler, line 180: Where is `headerRowStrings` populated for this type?
To me it looks like it will only contain something if a previous sheet contained instances and then it will be the header of that sheet repeated and not a header based on the type for this sheet.

emanuelaepure10: Very true indeed. 
I'll dig into it further to see how can I get the properties of a TypeDefinition without having an instance associated.

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/writer/AbstractTableInstanceWriter.java
```diff
@@ -146,6 +146,45 @@ else if (useSchema) {
...
+	 * @param useSchema <code>true</code> if properties should be defined from
+	 *            the schema, otherwise <code>false</code> and properties are
+	 *            defined from the Instances, but for the type cannot be taken
+	 *            into consideration as there is no instance available
+	 * @return a map of properties with string of localpart of the QName of the
+	 *         property as key
+	 */
+	protected Map<String, Object> getPropertyMap(TypeDefinition definition, List<String> headerRow,
+			boolean useSchema, boolean solveNestedProperties) {
+		// properties of current instance
```
stempler, line 166: Comments need to be updated because there is no instance being looked at?

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/writer/AbstractTableInstanceWriter.java
```diff
@@ -146,6 +146,45 @@ else if (useSchema) {
...
+	 *            should be solved, otherwise <code>false</code>
+	 * @param useSchema <code>true</code> if properties should be defined from
+	 *            the schema, otherwise <code>false</code> and properties are
+	 *            defined from the Instances, but for the type cannot be taken
+	 *            into consideration as there is no instance available
+	 * @return a map of properties with string of localpart of the QName of the
+	 *         property as key
+	 */
+	protected Map<String, Object> getPropertyMap(TypeDefinition definition, List<String> headerRow,
+			boolean useSchema, boolean solveNestedProperties) {
```
stempler, line 165: `useSchema` - this is not really a variable here, because you work based on the schema in any case.
`solveNestedProperties` - similar here, since you only look at top level properties and not what kind of nested properties there might be. Which is perfectly fine for the use case.

stempler (approved):
LGTM

PR #1082, opened 2023-11-16
fix: Fix TopoJSON encoding issue

emanuelaepure10:
Use Jackson's ObjectMapper to parse the JSON string into a JsonNode, then recursively traverses the JsonNode to replace "null" values with null. Finally, it converts the modified JsonNode back to a JSON string.

ING-3979

### DIFF
--- a/io/plugins/eu.esdihumboldt.hale.io.topojson.test/src/eu/esdihumboldt/hale/io/topojson/test/TopoJsonInstanceWriterTest.groovy
+++ b/io/plugins/eu.esdihumboldt.hale.io.topojson.test/src/eu/esdihumboldt/hale/io/topojson/test/TopoJsonInstanceWriterTest.groovy
@@ -16,10 +16,14 @@
 package eu.esdihumboldt.hale.io.topojson.test
 
 import static org.junit.Assert.assertEquals
+import static org.junit.Assert.assertNull
 import static org.junit.Assert.assertTrue
 
 import java.nio.file.Files
 import java.nio.file.Path
+import java.time.Instant
+import java.time.LocalDate
+import java.time.format.DateTimeFormatter
 import java.util.function.Consumer
 
 import org.junit.Test
@@ -56,6 +60,8 @@ class TopoJsonInstanceWriterTest {
 		SimpleType {
 			id(String)
 			name(String)
+			label(String)
+			date(LocalDate)
 			geometry(GeometryProperty)
 		}
 	}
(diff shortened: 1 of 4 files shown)


io/plugins/eu.esdihumboldt.hale.io.topojson/src/eu/esdihumboldt/hale/io/topojson/TopoJsonInstanceWriter.java
```diff
@@ -171,8 +177,60 @@ private void convertShapefileToTopoJson(String shapefilePath,
...
+
+			if (compressOutput) {
+				Toolbox.writeFile(targetPath, Compress.compressB64(modifiedJsonString));
+			}
+			else {
+				Toolbox.writeFile(targetPath, modifiedJsonString);
+			}
+
+		} catch (Exception e) {
+			e.printStackTrace();
```
stempler, line 201: The exception should not be dropped silently but end up in the report and ideally make it fail. You should be able to achieve that by letting this method throw it.

io/plugins/eu.esdihumboldt.hale.io.topojson/src/eu/esdihumboldt/hale/io/topojson/TopoJsonInstanceWriter.java
```diff
@@ -171,8 +177,60 @@ private void convertShapefileToTopoJson(String shapefilePath,
...
+	/**
+	 * @param node to replace the unwanted characters
+	 * @return the clean node
+	 */
+	public static JsonNode replaceUnwantedCharacters(JsonNode node) {
+		if (node.isObject()) {
+			ObjectNode objectNode = (ObjectNode) node;
+			objectNode.fields().forEachRemaining(entry -> {
+				JsonNode value = entry.getValue();
+				if (value.isTextual() && value.textValue().contains("\u0000")) {
```
stempler, line 214: What about the case with the null date?

emanuelaepure10: That's missing at the moment. 
I'm uncertain about the best approach to resolve the date issue.

io/plugins/eu.esdihumboldt.hale.io.topojson/src/eu/esdihumboldt/hale/io/topojson/TopoJsonInstanceWriter.java
```diff
@@ -169,10 +181,110 @@ private List<URI> writeIntermediateShapefiles(InstanceCollection instances,
...
+									objectNode.put(entry.getKey(), textValue);
+								}
+
+							}
+							else {
+								objectNode.put(entry.getKey(), textValue);
+							}
+
+						} catch (Exception e) {
+							objectNode.put(entry.getKey(), (String) null);
```
stempler, line 261: I think it would be good to log this exception, since it seems to be unexpected that it may happen (since date parsing errors are handled somewhere else).

io/plugins/eu.esdihumboldt.hale.io.topojson.test/src/eu/esdihumboldt/hale/io/topojson/test/TopoJsonInstanceWriterTest.groovy
```diff
@@ -56,6 +57,7 @@ class TopoJsonInstanceWriterTest {
 		SimpleType {
 			id(String)
 			name(String)
+			label(String)
```
stempler, line 60: Having an additional unit test to test the case with the null date would be great.

io/plugins/eu.esdihumboldt.hale.io.topojson.test/src/eu/esdihumboldt/hale/io/topojson/test/TopoJsonInstanceWriterTest.groovy
```diff
@@ -144,8 +155,9 @@ class TopoJsonInstanceWriterTest {
 			assertEquals(1, json.objects.Topology.geometries[1].arcs.size())
 			assertEquals(3, json.objects.Topology.geometries[1].arcs[0].size())
 			assertEquals([-1, 2, 3], json.objects.Topology.geometries[1].arcs[0])
-		}
 
+			assertNull(json.objects.Topology.geometries[1].'properties'.label)
+			assertNull(json.objects.Topology.geometries[1].'properties'.date)
```
stempler, line 160: Please also check the case where a value should be present, to test that that value is not replaced by null unintentionally.

stempler (approved):

PR #1083, opened 2023-11-28
fix: TopoJSON encoding null values

emanuelaepure10:
Convert all the "null" String values into null value.
Update test for date values.

ING-3979

### DIFF
--- a/io/plugins/eu.esdihumboldt.hale.io.topojson.test/src/eu/esdihumboldt/hale/io/topojson/test/TopoJsonInstanceWriterTest.groovy
+++ b/io/plugins/eu.esdihumboldt.hale.io.topojson.test/src/eu/esdihumboldt/hale/io/topojson/test/TopoJsonInstanceWriterTest.groovy
@@ -21,9 +21,7 @@ import static org.junit.Assert.assertTrue
 
 import java.nio.file.Files
 import java.nio.file.Path
-import java.time.Instant
 import java.time.LocalDate
-import java.time.format.DateTimeFormatter
 import java.util.function.Consumer
 
 import org.junit.Test
@@ -109,10 +107,8 @@ class TopoJsonInstanceWriterTest {
 
 	@Test
 	public void testWriteTopoJson() {
-		def aDate = LocalDate.of(2017, 10, 3)
-		def aTimestamp = Instant.parse('2023-04-02T00:00:00Z')
-		// Create a DateTimeFormatter with the desired pattern
-		def formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss'Z'");
+		def aDate = LocalDate.of(2023, 11, 28)
+		def badDate = null
 
 		InstanceCollection instances = new InstanceBuilder(types: schema).createCollection {
 			SimpleType {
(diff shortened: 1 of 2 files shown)


io/plugins/eu.esdihumboldt.hale.io.topojson.test/src/eu/esdihumboldt/hale/io/topojson/test/TopoJsonInstanceWriterTest.groovy
```diff
@@ -138,6 +134,14 @@ class TopoJsonInstanceWriterTest {
 				date null
 				geometry(createGeometry('POLYGON ((20 10, 40 10, 40 20, 20 20, 20 10))', 4326))
 			}
+
+			SimpleType {
+				id '4'
+				name 'Area 4\u0000'
+				label 'null'
+				date badDate
```
stempler, line 142: Why did you remove the test with the valid date? I think it would be still good to test a valid date as well. The value is already null on the other instances, so how the null date is handled can be tested on them.

io/plugins/eu.esdihumboldt.hale.io.topojson/src/eu/esdihumboldt/hale/io/topojson/TopoJsonInstanceWriter.java
```diff
@@ -268,11 +273,11 @@ public static JsonNode replaceUnwantedCharacters(JsonNode node) {
 								}
 								else {
 									// Create a DateTimeFormatter object with
-									// the desired format
-									DateTimeFormatter formatter = DateTimeFormatter
-											.ofPattern("yyyy-MM-dd'T'HH:mm:ss'Z'");
+//									// the desired format
+//									DateTimeFormatter formatter = DateTimeFormatter
```
stempler, line 277: Is this no longer done because there can be a date or date and time?

stempler: Was just wondering why it was removed but still left as comment. I would say to use the original text value is fine and in line with the previous behavior.

io/plugins/eu.esdihumboldt.hale.io.topojson.test/src/eu/esdihumboldt/hale/io/topojson/test/TopoJsonInstanceWriterTest.groovy
```diff
@@ -109,10 +107,8 @@ class TopoJsonInstanceWriterTest {
 
 	@Test
 	public void testWriteTopoJson() {
-		def aDate = LocalDate.of(2017, 10, 3)
-		def aTimestamp = Instant.parse('2023-04-02T00:00:00Z')
-		// Create a DateTimeFormatter with the desired pattern
-		def formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss'Z'");
+		def aDate = LocalDate.of(2023, 11, 28)
```
emanuelaepure10, line 110: @stempler I have used this one as valid date.

stempler: Ah, missed that it was still used in the first instance.

stempler (approved):

PR #1088, opened 2023-11-30
fix: TopoJSON: Convert String integer/number and String boolean into integer/number or boolean

emanuelaepure10:
convert String integer/number and String boolean into integer/number or boolean

ING-3979


stempler (approved):
LGTM

PR #1090, opened 2023-12-13
build: hale: GeoTools update

emanuelaepure10:
Update hale»studio with GeoTools 29.1

ING-4083


emanuelaepure10:
Please review the comments in [this section](https://wetransform.atlassian.net/browse/ING-4083)  where unresolved issues have been documented. 
Your insights and assistance in addressing these concerns would be greatly appreciated. 
Thank you

stempler:
> Missing requirement: eu.esdihumboldt.hale.common.instance 5.1.0.qualifier requires 'java.package; org.geotools 29.1.0.combined' but it could not be found

Downloading the [Geotools bundle from your referenced commit](https://gitlab.wetransform.to/emanuelaepure/hale-build-support/-/blob/c869ebf7a62caf1f6e10e78eaed7d970ae391d23/updatesites/platform/plugins/org.geotools_29.1.0.combined-i20231214.jar) I can't find a package `org.geotools`, only subpackages. I would guess that the package structure changed and the `org.geotools` package just was removed/refactored and the dependency to it can be removed.
Does [this](https://github.com/emanuelaepure10/hale/blob/387146ebb8511427210ac084baae3d8fd12597a8/common/plugins/eu.esdihumboldt.hale.common.instance/META-INF/MANIFEST.MF#L43) not lead to an error in Eclipse?

stempler:
Build error:

```
Cannot resolve project dependencies:
  Software being installed: eu.esdihumboldt.hale.common.feature.core.feature.group 5.1.0.qualifier
  Missing requirement: eu.esdihumboldt.hale.common.feature.core.feature.group 5.1.0.qualifier requires 'org.eclipse.equinox.p2.iu; commons-httpclient 0.0.0' but it could not be found
```

emanuelaepure10:
I haven't encountered this error in my local environment.

Upon inspecting the [hale-platform repository for commons-httpclient](https://github.com/halestudio/hale-platform/blob/1e7e00c9684c9043c98a857338fd1ccfee05541b/version-history.json#L1252), it appears to be used solely in the version history.

In hale studio, it seems to be utilized as shown in the image below:
image
However, during the validation of Test.product, no error is reported, even though it contains the 'commons-httpclient' plugin. 

Surprisingly, this plugin doesn't appear in the list of installed plugins.

@stempler Do you have any suggestions on how to address this issue?

stempler:
@emanuelaepure10 If the plugin is no longer used (I guess then it is probably replaced by one with a different name), then remove it from the feature and the products.

emanuelaepure10:
[error_log.log](https://github.com/halestudio/hale/files/13872391/error_log.log)
I am receiving warnings from the attached files locally, and as a result, the product fails to start.

stempler:
```
Caused by: java.lang.ClassNotFoundException: si.uom.NonSI cannot be found by org.geotools_29.1.0.combined-i202401081457
        at org.eclipse.osgi.internal.loader.BundleLoader.findClass(BundleLoader.java:516)
        at org.eclipse.osgi.internal.loader.ModuleClassLoader.loadClass(ModuleClassLoader.java:171)
        at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:520)
        ... 40 more
```

The problem looks quite clear - did you check if there is a bundle that provides that package/class?

emanuelaepure10:
image
The plugin is there

emanuelaepure10:
this

```
> ./build.sh clean
> ./build.sh product HALE -o windows -a x86_64
```

save me a LOT of time. I wish I would have known before about them.

Locally, works for me
image

emanuelaepure10:
image
So many missing constraints at the moment, which are all new :-( 
I feel that somehow there is something wrong somewhere...

emanuelaepure10:
Locally I still get 19 errors
but almost all of them have this error
java.nio.file.FileSystemException: C:\Users\EMANUE~1\AppData\Local\Temp\new10352906332041916538.gpkg: The process cannot access the file because it is being used by another process
	at java.base/sun.nio.fs.WindowsException.translateToIOException(WindowsException.java:92)
	at java.base/sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:103)
	at java.base/sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:108)
	at java.base/sun.nio.fs.WindowsFileSystemProvider.implDelete(WindowsFileSystemProvider.java:275)
	at java.base/sun.nio.fs.AbstractFileSystemProvider.delete(AbstractFileSystemProvider.java:105)
	at java.base/java.nio.file.Files.delete(Files.java:1152)
	at eu.esdihumboldt.hale.io.geopackage.GeopackageInstanceWriterTest.withNewGeopackage(GeopackageInstanceWriterTest.groovy:98)
	OR
	java.lang.RuntimeException: java.nio.file.FileSystemException: C:\Users\EMANUE~1\AppData\Local\Temp\tmp12857606802334693251.gpkg: The process cannot access the file because it is being used by another process
	at eu.esdihumboldt.hale.io.geopackage.GeopackageApiTest.withTestFile(GeopackageApiTest.java:73)
	at eu.esdihumboldt.hale.io.geopackage.GeopackageApiTest.testLoadFile(GeopackageApiTest.java:82)
	
	and I was wondering if also Jenkins has the same.

stempler:
@emanuelaepure10 The error you observe could be specific to when Windows file systems are used and since it is related to Geopackage likely unrelated to the Geotools update. Could well be that these errors don't occur on Jenkins, though that is currently stuck on a dependency issue building the test product as you can see in the logs.
I would not see that as a blocker for merging the Geotools update, if tests succeed on the build server, though we should create an issue for solving the problem, as this may be a problem for Windows users.

emanuelaepure10:
@stempler This is my latest test output [hale.log](https://github.com/halestudio/hale/files/13960794/hale.log)

Out of 19 test failed there are:

-  _14 tests failed for_ 

```
Caused by: java.nio.file.FileSystemException: PATH\AppData\Local\Temp\tmp12857606802334693251.gpkg: The process cannot access the file because it is being used by another process
	at java.base/sun.nio.fs.WindowsException.translateToIOException(WindowsException.java:92)
	at java.base/sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:103)
	at java.base/sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:108)
	at java.base/sun.nio.fs.WindowsFileSystemProvider.implDelete(WindowsFileSystemProvider.java:275)
	at java.base/sun.nio.fs.AbstractFileSystemProvider.delete(AbstractFileSystemProvider.java:105)
	at java.base/java.nio.file.Files.delete(Files.java:1152)
	at eu.esdihumboldt.hale.io.geopackage.GeopackageApiTest.withTestFile(GeopackageApiTest.java:68)
```

- _3 tests failed for_ 

```
java.io.IOException: Read Channel is not a File Channel so this is not possible.
	at org.geotools.data.shapefile.fid.IndexedFidReader.goTo(IndexedFidReader.java:249)
	at org.geotools.data.shapefile.ShapefileFeatureReader.buildFeatureId(ShapefileFeatureReader.java:291)
	at org.geotools.data.shapefile.ShapefileFeatureReader.buildFeature(ShapefileFeatureReader.java:259)
	at org.geotools.data.shapefile.ShapefileFeatureReader.hasNext(ShapefileFeatureReader.java:195)
	at org.geotools.data.store.ContentFeatureCollection$WrappingFeatureIterator.hasNext(ContentFeatureCollection.java:145)
```

- and last  _2 tests failing that I should say that should be solved are_ 

```
java.lang.ClassNotFoundException
Failed to load class eu.esdihumboldt.hale.ui.codelist.service.internal.config.CodeListAssociationsTypeTest from fragment hosts
Caused by: java.lang.ClassNotFoundException: An error occurred while automatically activating bundle eu.esdihumboldt.hale.ui.codelist (232).
	at org.eclipse.osgi.internal.hooks.EclipseLazyStarter.postFindLocalClass(EclipseLazyStarter.java:126)
```

 
and 

```
java.lang.ClassNotFoundException
Failed to load class eu.esdihumboldt.hale.ui.util.groovy.ast.ASTGraphUtilTest from fragment hosts
Caused by: java.lang.ClassNotFoundException: An error occurred while automatically activating bundle eu.esdihumboldt.hale.ui.util.groovy (241).
```

Do you have any suggesiton on the last 2?
Thank you

emanuelaepure10:
@stempler No local changes observed with the latest updates.

stempler:
@emanuelaepure10 The three remaining test failures seem to be caused by a change in Geotools that requires that Shapefiles have to be loaded from the file system and other mechanisms (e.g. from a URL that points to a Jar entry) are not supported.

emanuelaepure10:
@stempler I noticed that the Jenkins build was automatically aborted due to a timeout. Can I initiate a new build?

stempler:
retest this please

emanuelaepure10:
@stempler Do you have any insights into why I'm not encountering the three failing tests locally that Jenkins seems to be reporting?
[hale.log](https://github.com/halestudio/hale/files/13982637/hale.log)

stempler:
> Do you have any insights into why I'm not encountering the three failing tests locally that Jenkins seems to be reporting?

See

> The three remaining test failures seem to be caused by a change in Geotools that requires that Shapefiles have to be loaded from the file system and other mechanisms (e.g. from a URL that points to a Jar entry) are not supported.

Likely when running from Eclipse the resources are read directly from the file system and on Jenkins with the built test product they are contained in Jars.

emanuelaepure10:
@stempler @florianesser 
I could use a helping hand here.

As I'm unable to locally test the code responsible for extracting files from a JAR, I find myself writing this with fingers crossed, hoping it works.

Would you mind lending me your expertise on this matter?

Thank you in advance.

emanuelaepure10:
We finally have a successful build! :-)

stempler:
@emanuelaepure10 Can you please create a respective merge request for hale-build-support? Thanks!

common/plugins/eu.esdihumboldt.hale.common.filter.test/src/eu/esdihumboldt/hale/common/filter/FilterTest.java
```diff
@@ -186,6 +195,58 @@ public void simpleSchemaTestCQL() throws Exception {
...
+	 */
+	private Path getTempZip() throws IOException {
+		// Specify the path of the resource
+		String resourcePath = "/testdata/GN_Point/GN_Point.zip";
+
+		// Get the input stream for the resource
+		InputStream inputStream = FilterTest.class.getResourceAsStream(resourcePath);
+
+		// Create a temporary directory
+		Path tempDirectory = Files.createTempDirectory("ShapefileTest");
```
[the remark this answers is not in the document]
emanuelaepure10: @stempler Thank you for the hint.

emanuelaepure10:
@stempler Done.
I hope that you have received the mail from GitLab for the PR :-)

common/plugins/eu.esdihumboldt.hale.common.filter.test/src/eu/esdihumboldt/hale/common/filter/FilterTest.java
```diff
@@ -129,61 +133,82 @@ public void simpleFilterTestCQL() throws CQLException {
...
+		Path tempDirectory = java.nio.file.Files.createTempDirectory("extracted-zip");
 
-		schemaReader.validate();
-		IOReport report = schemaReader.execute(null);
-		assertTrue(report.isSuccess());
-
-		Schema schema = schemaReader.getSchema();
+		try {
+			Path shpTempFile = IOUtils.withTemporaryExtractedZipResource(resourcePath,
+					FilterTest.class, FilterTest::processContents, tempDirectory);
```
stempler, line 142: You don't need to use a method for the functional parameter, you can define it inline.

I would suggest to add the whole logic that requires the temporary files to be present to add to the function and leave the creation and the cleanup of the temporary folder to the `withTemporaryExtractedZipResource` method. Otherwise you could have just used `IOUtils.extract` directly.

For example:

```
IOUtils.withTemporaryExtractedZipResource(resourcePath, FilterTest.class, (Path tempFolder) -> {
  //TODO process extracted Shapefile
});
```

emanuelaepure10: @stempler 
I'm facing a situation where I cannot include the temporary files in the `IOUtils.withTemporaryExtractedZipResource` because the folder needs to be cleaned after the extracted files are used. The lines below illustrate how the source is set using the temporary file:
`schemaReader.setSource(new DefaultInputSupplier(shpTempFile.toUri()));`
`instanceReader.setSource(new DefaultInputSupplier(shpTempFile.toUri()));`

However, using only `IOUtils.extract` is not an option either, as it would require spreading the logic across three different places. 

What would you recommend or prefer in this scenario?

stempler: The idea with using the function as parameter as suggested was to do all the processing that requires the files to be present in the passed-in function/lambda, so that `withTemporaryExtractedZipResource` can handle all the hassle with creating and deleting the temporary folder transparently.

emanuelaepure10: I understand. Are you suggesting that the logic should be relocated back to each individual test?

common/plugins/eu.esdihumboldt.hale.common.filter.test/src/eu/esdihumboldt/hale/common/filter/FilterTest.java
```diff
@@ -129,61 +131,33 @@ public void simpleFilterTestCQL() throws CQLException {
...
+	}
 
+	/**
+	 * Process the contents of the extracted directory.
+	 *
+	 * @param extractedPath The path to the extracted directory.
+	 * @return The processed result (in this case, the path to the processed
+	 *         file).
+	 */
+	public static Path processContents(Path extractedPath) {
```
stempler, line 159: I guess this is no longer used?

io/plugins/eu.esdihumboldt.hale.io.shp.test/src/eu/esdihumboldt/hale/io/shp/ShapefileInstanceWriterTest.groovy
```diff
@@ -224,47 +226,65 @@ class ShapefileInstanceWriterTest {
...
 
+	/**
+	 * Process the contents of the extracted directory.
+	 *
+	 * @param extractedPath The path to the extracted directory.
+	 * @return The processed result (in this case, the path to the processed file).
+	 * @throws IOException If an I/O error occurs.
+	 */
 
+	def processContents(Path extractedPath) {
```
stempler, line 285: I guess this is no longer used?

stempler (approved):
Thanks a lot for the changes!
There are only two minor things on methods I think are no longer used.

io/plugins/eu.esdihumboldt.hale.io.shp.test/src/eu/esdihumboldt/hale/io/shp/ShapefileInstanceWriterTest.groovy
emanuelaepure10: Exactly. I have removed them

common/plugins/eu.esdihumboldt.hale.common.filter.test/src/eu/esdihumboldt/hale/common/filter/FilterTest.java
emanuelaepure10: Exactly. I have removed them. Thanks

PR #1093, opened 2024-01-08
fixup!: limited list of export formats is removed

Kate-Lyndegaard:
This pull request updates the list of available export formats.

### DIFF
--- a/doc/plugins/eu.esdihumboldt.hale.doc.user/html/tasks/data/save_transformation.html
+++ b/doc/plugins/eu.esdihumboldt.hale.doc.user/html/tasks/data/save_transformation.html
@@ -8,7 +8,7 @@
 	<h1>Save the transformation result</h1>
 	<p>
 		If you have defined a mapping and loaded source data in the
-		application, <i>hale»studio</i> performs the transformation derived from the
+		application, <i>haleï¿½studio</i> performs the transformation derived from the
 		mapping on the source data. This happens on every change to the
 		mapping, to instantly show the user what consequences his changes
 		have.
@@ -31,11 +31,10 @@
 		work based on a XML target schema.
 	</p>
 	<p>
-		Supported data export formats in <i>hale»studio</i> are:
+		Supported data export formats in <i>haleï¿½studio</i> can be viewed on the following page:
 	</p>
 	<ul>
-		<li><a href="../../reference/export/xml_data.html">XML</a></li>
-		<li><a href="../../reference/export/gml_data.html">GML</a></li>
+		<li><a href="http://help.halestudio.org/latest/nav/0_7_2_1">Supported formats for export</a></li>
 	</ul>
 	<p>When specifying the target file you also have the possibility to
 		select a validator that is executed after the file is written, to
@@ -50,7 +49,7 @@
 			<a href="PLUGINS_ROOT/eu.esdihumboldt.hale.doc.user/html/concepts/what_is_instance.html">What is an Instance?</a>
 		</p>
 	</div>
-	
+
 	<!-- related tasks: load target schema, load source data -->
 	<div>
 		<br> <img
--- a/doc/plugins/eu.esdihumboldt.hale.doc.user/html/tasks/data/save_transformation.html
+++ b/doc/plugins/eu.esdihumboldt.hale.doc.user/html/tasks/data/save_transformation.html
@@ -8,7 +8,7 @@
 	<h1>Save the transformation result</h1>
 	<p>
 		If you have defined a mapping and loaded source data in the
-		application, <i>hale�studio</i> performs the transformation derived from the
+		application, <i>hale»studio</i> performs the transformation derived from the
 		mapping on the source data. This happens on every change to the
 		mapping, to instantly show the user what consequences his changes
 		have.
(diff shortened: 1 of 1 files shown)


doc/plugins/eu.esdihumboldt.hale.doc.user/html/tasks/data/save_transformation.html
```diff
@@ -8,7 +8,7 @@
 	<h1>Save the transformation result</h1>
 	<p>
 		If you have defined a mapping and loaded source data in the
-		application, <i>hale�studio</i> performs the transformation derived from the
+		application, <i>hale�studio</i> performs the transformation derived from the
```
JohannaOtt, line 11: The special character (or its encoding) between "hale" and "studio" seems to have been replaced with the changes done. Any chance to check if it will still display correctly in the online help?

Kate-Lyndegaard: @stempler 

I can see the encoding issue in both the original version and in the new changes:

grafik

Is this expected? Is it a browser or GitHub display issue?

stempler: To me it shows correct in GitHub for the old version:

image

Since currently the encoding is correct on the website it would probably be safest to leave it as it was before.

emanuelaepure10: @stempler If you refresh the page, you'll notice that both the `conversation `and `file changed` will be displayed with incorrect encoding.

stempler: For me there is no difference when refreshing.
It may be shown correctly with the change, but we don't really know (maybe the new encoding is ISO-8859-1 as stated in the HTML file, but maybe that is actually not respected for the final help).
So just saying, since it's working right now in the final product, it would be the safe bet to not change it.

emanuelaepure10 (approved):
LGTM

doc/plugins/eu.esdihumboldt.hale.doc.user/html/tasks/data/save_transformation.html
```diff
@@ -8,7 +8,7 @@
 	<h1>Save the transformation result</h1>
 	<p>
 		If you have defined a mapping and loaded source data in the
-		application, <i>hale�studio</i> performs the transformation derived from the
+		application, <i>hale»studio</i> performs the transformation derived from the
```
JohannaOtt, line 11: There still seems to be some kind of encoding change as the line is marked as changed. I cannot assess if this could cause issues in how it is displayed, or not.

doc/plugins/eu.esdihumboldt.hale.doc.user/html/tasks/data/save_transformation.html
```diff
@@ -31,11 +31,10 @@ <h1>Save the transformation result</h1>
 		work based on a XML target schema.
 	</p>
 	<p>
-		Supported data export formats in <i>hale�studio</i> are:
+		Supported data export formats in <i>hale»studio</i> can be viewed on the following page:
 	</p>
 	<ul>
-		<li><a href="../../reference/export/xml_data.html">XML</a></li>
-		<li><a href="../../reference/export/gml_data.html">GML</a></li>
+		<li><a href="http://help.halestudio.org/latest/nav/0_7_2_1">Supported formats for export</a></li>
```
stempler, line 37: Downside with linking to the online help is that there may be a different selection of formats than available in the local version (e.g. when the local version includes the pro plugins).
I could not find any information in the [Eclipse Help](https://help.eclipse.org/latest/topic/org.eclipse.platform.doc.isv/guide/ua_help.htm?cp=2_0_19_1) though, how to link to navigation topics.

I asked ChatGPT about it and it [suggests that there is a notation that allows linking to a topic without help page given its id](https://chat.openai.com/share/f2b61921-f032-4e92-af22-cb35c2293633), but I don't know if it just made that up (I could not find any documentation/example of that).

If that does not work a workaround could be adding an HTML file to the Export topic (which currently only has children but no HTML page), and linking to that.
According to the documentation [here](https://help.eclipse.org/latest/topic/org.eclipse.platform.doc.isv/guide/ua_help_content_child_links.htm?cp=2_0_19_1_1_10) it should be possible to generate links to that child pages for that HTML page, for it to be similar to the current auto-generated representation.

Though to properly test all these things it would be required to run hale studio locally from Eclipse.

emanuelaepure10: @stempler Would you please provide the information on how to re-build the documentation. Can I find anywhere documented how to build it? 
I have added the `<!--INSERT_CHILD_LINKS-->` and I would like to try it out.
Thank you

stempler: @emanuelaepure10 Easiest way is to run hale studio locally from Eclipse and launch the help from the menu (Help -> Help)

emanuelaepure10: @stempler In this way the files are not "forced" to be recreated, so the `<!--INSERT_CHILD_LINKS-->` is not considered.

stempler:
Closing in favor of https://github.com/halestudio/hale/pull/1109

PR #1096, opened 2024-01-17
fix: hale: Removal of project variable is not persisted on save

emanuelaepure10:
Removing of a variable via settings dialog is persisted in the project file upon save.

ING-4169


ui/plugins/eu.esdihumboldt.hale.ui/src/eu/esdihumboldt/hale/ui/service/project/internal/ProjectVariablesPreferencePage.java
```diff
@@ -205,6 +205,11 @@ public void widgetSelected(SelectionEvent e) {
 						c.dispose();
 					}
 					variables.remove(varName);
+
+					ProjectService ps = PlatformUI.getWorkbench().getService(ProjectService.class);
+					ps.getConfigurationService().setProperty(
+							ProjectVariables.PROJECT_PROPERTY_VARIABLES, Value.complex(variables));
```
stempler, line 211: This will directly store the removed variable in the project service, when a user clicked on remove. This should only happen though when a user actually applies the changes from the preference page.

To me it looks like what is missing here could be a `varControls.remove(varName);`.

stempler (changes requested):

stempler (approved):

PR #1101, opened 2024-02-02
fix/feat:  add GUI parameter to select date formatter for the Excel/CSV

emanuelaepure10:
The issue has been extended to CSV as well, as the code is/might be common in some points. 

Add a combo box with different patterns for the date formatter + a custom text. This pattern is then used when importing cells/rows of type Date.

ING-4165

### DIFF
--- a/io/plugins/eu.esdihumboldt.hale.io.csv.ui/src/eu/esdihumboldt/hale/io/csv/ui/TypeSelectionPage.java
+++ b/io/plugins/eu.esdihumboldt.hale.io.csv.ui/src/eu/esdihumboldt/hale/io/csv/ui/TypeSelectionPage.java
@@ -44,7 +44,6 @@ import eu.esdihumboldt.hale.common.schema.model.TypeDefinition;
 import eu.esdihumboldt.hale.io.csv.reader.CSVConstants;
 import eu.esdihumboldt.hale.io.csv.reader.CommonSchemaConstants;
 import eu.esdihumboldt.hale.io.csv.reader.internal.CSVConfiguration;
-import eu.esdihumboldt.hale.io.xls.reader.ReaderSettings;
 import eu.esdihumboldt.hale.ui.common.definition.selector.TypeDefinitionSelector;
 import eu.esdihumboldt.hale.ui.io.instance.InstanceReaderConfigurationPage;
 import eu.esdihumboldt.hale.ui.service.schema.SchemaService;
@@ -59,7 +58,11 @@ public class TypeSelectionPage extends InstanceReaderConfigurationPage implement
 	/**
 	 * Parameter for the the custom date label
 	 */
-	private static final String CUSTOM_FORMAT_LABEL = "Custom format";
+	public static final String CUSTOM_FORMAT_LABEL = "Custom format";
+	/**
+	 * Parameter for combo box for an empty selection
+	 */
+	public static final String EMPTY_SELECTION = "";
 
 	private TypeDefinitionSelector sel;
 	private Spinner skipNlinesSpinner;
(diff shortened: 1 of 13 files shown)


io/plugins/eu.esdihumboldt.hale.io.csv.ui/src/eu/esdihumboldt/hale/io/csv/ui/TypeSelectionPage.java
```diff
@@ -202,11 +202,11 @@ public boolean updateConfiguration(InstanceReader provider) {
 				Value.of(skipNlinesSpinner.getSelection()));
 
 		if (customFormat.isVisible()) {
-			provider.setParameter(ReaderSettings.PARAMETER_DATE_FORMAT,
-					Value.of(customFormat.getText()));
+			provider.setParameter(CSVConstants.PARAMETER_DATE_FORMAT, Value.of(customFormat
+					.getText().replace("mm", "MM").replace("DD", "dd").replace("YYYY", "yyyy")));
```
stempler, line 206: Why are minutes replaced by months?

io/plugins/eu.esdihumboldt.hale.io.csv.ui/src/eu/esdihumboldt/hale/io/csv/ui/TypeSelectionPage.java
```diff
@@ -202,11 +202,11 @@ public boolean updateConfiguration(InstanceReader provider) {
 				Value.of(skipNlinesSpinner.getSelection()));
 
 		if (customFormat.isVisible()) {
```
stempler, line 204: It seems to be not possible to not specify a value in the UI, but I think at least for CSV it should be the default to not specify the parameter because here the mechanism only works on guessing which values might be dates and the data may be modified unintentionally.

Also I think it would be good to make the description of the parameter more detailed for what the behavior is for Excel and CSV.

Suggestions:

Excel: "Format for imported date values (values are represented as strings)"

CSV: "Reformatting of dates (date values are tried to be detected automatically)"

io/plugins/eu.esdihumboldt.hale.io.csv.ui/src/eu/esdihumboldt/hale/io/csv/ui/TypeSelectionPage.java
emanuelaepure10: @stempler Good question. That come from Kate that has been adding date as **dd.mm-yyyy** as a `custom format`.
I would then propose to remove this `replace `and add in the interface a text informing on how to custom format should be composed.

emanuelaepure10:
@stempler would you have time to review this? Thank you

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/reader/internal/CSVInstanceCollection.java
```diff
@@ -174,12 +182,72 @@ public Instance next() {
...
 			}
 
+			try {
+				String dateTime = reader.getParameter(CSVUtil.PARAMETER_DATE_FORMAT)
+						.as(String.class);
+				if (dateTime != null && !part.isEmpty()) {
+					part = parseDate(part, dateTime);
+				}
+			} catch (Exception e) {
+				// Handle the exception appropriately, but no need for this
```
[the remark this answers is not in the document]
stempler: So ignoring the exception is intended?
Maybe adapt the comment then?
"Handle the exception appropriately" sounds like a TODO

io/plugins/eu.esdihumboldt.hale.io.csv/plugin.xml
```diff
@@ -68,6 +68,19 @@
                   sampleDescription="Any integer equal or larger than 0">
             </valueDescriptor>
          </providerParameter>
+         <providerParameter
+               description="Date Time Formatter specifying how String values representing a Date should be imported"
```
stempler, line 72: Maybe align to the text on the UI label?

stempler (approved):

PR #1102, opened 2024-02-05
fix-adv: add code mapping EPSG:7837

emanuelaepure10:
Activate the code mapping EPSG:7837 for AdvCrsAuthorityFactory

ING-4083


florianesser (approved):

florianesser:
One comment on the commit message: `fix-adv:` is not a valid commit type according to the Developer Guidelines. To specify `adv` as a scope, you could use `fix(adv):`.

PR #1103, opened 2024-02-06
Alignment Merger: apply conditions to individually joined types

stempler:


stempler:
Note: waiting for customer feedback

PR #1105, opened 2024-02-08
fix: fix check if an object is identifiable

stempler:
Due to the wrong check, the index based Merge and Join handlers were not used at all. Also, no instances were added to the index service.


florianesser (approved):
LGTM

PR #1110, opened 2024-02-20
Fixes and refactorings for the instance index

florianesser:
- add workaround for the edge case where no target instance was created for merges that don't have a key property
- fix that a retransformation did not immediately use the updated instance index after an alignment change but had to be manually retriggered to work correctly


stempler (approved):
LGTM

PR #1111, opened 2024-02-21
fix: export shp file

emanuelaepure10:
Fix the export of a shape file which was failing with the error “can handle 7 attributes only, index is 7.“

ING-4189

### DIFF
--- a/io/plugins/eu.esdihumboldt.hale.io.shp/src/eu/esdihumboldt/hale/io/shp/writer/ShapefileInstanceWriter.java
+++ b/io/plugins/eu.esdihumboldt.hale.io.shp/src/eu/esdihumboldt/hale/io/shp/writer/ShapefileInstanceWriter.java
@@ -569,16 +569,16 @@ public class ShapefileInstanceWriter extends AbstractGeoInstanceWriter {
 				TypeDefinition type = instance.getDefinition();
 				String localPart = type.getName().getLocalPart();
 				if (schemaFtMap.containsKey(localPart)) {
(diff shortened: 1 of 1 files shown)


emanuelaepure10:
@stempler thank you
you were faster them me asking for it :-D

io/plugins/eu.esdihumboldt.hale.io.shp/src/eu/esdihumboldt/hale/io/shp/writer/ShapefileInstanceWriter.java
```diff
@@ -635,17 +634,16 @@ private void writePropertiesInstanceData(
...
 						.findChildren(prop.getName().getLocalPart()).value();
-				List<GeometryProperty<?>> geoms = traverseInstanceForGeometries(instance);
-				// add value by traversing geometryType from instance
-				for (GeometryProperty<?> geoProp : geoms) {
-					if (geoProp.getGeometry() != null) {
-						String geometryType = geoProp.getGeometry().getGeometryType();
-						if (schemaFbMap.get(localPart) != null
-								&& schemaFbMap.get(localPart).get(geometryType) != null) {
-							schemaFbMap.get(localPart).get(geometryType).add(value);
+				if (schemaFbMap.get(localPart) != null && value != null) {
```
kapil-agnihotri, line 637: @emanuelaepure10 as discussed I think a null check for `value` here might have impact on attributes not present in the export. Also, this might not prevent the IndexOutOfBounds exception if all the additional values are not null. I believe we should analyse the data and try to pass in all the attributes during schema creation in step 2.

PR #1113, opened 2024-02-22
Updated application versions for 5.1.0 release

emanuelaepure10:


stempler (changes requested):
Here I would expect a Merge from master before doing any additional changes. Now all changes including the changes from master are present in one commit and the original history is lost.

stempler:
> Following this guide
https://github.com/halestudio/hale/edit/master/RELEASE.md
I don't see that 2 commit should be done.

The merge is this step https://github.com/halestudio/hale/blob/6ee44b91c078559820b7054a28e05caaabfe6192/RELEASE.md?plain=1#L21
That should result in a merge commit.
The commit that is created afterwards should only cover the changes done afterwards.

PR #1120, opened 2024-02-27
refactor: add feature for unsafe IO providers

stempler:
...to easily be able to include them when running hale studio from Eclipse.


florianesser (approved):
LGTM

PR #1121, opened 2024-02-28
fix:  functions not working for elements with value and attribute

emanuelaepure10:
Try to find a solution to show values also for the cases where now is shown a String as eu.esdihumboldt.hale.common.instance.orient.storage.DatabaseHandle$InstanceHandle@1008ac5f

ING-4156


common/plugins/eu.esdihumboldt.hale.common.align/src/eu/esdihumboldt/hale/common/align/model/AlignmentUtil.java
```diff
@@ -1045,13 +1046,23 @@ public static void addValues(Group group, List<ChildContext> path,
...
+						}
+					}
+					else if (value instanceof Group) {
 						addValues((Group) value, subPath, collectedValues, onlyValues);
 					}
 					else if (subPath.isEmpty()) {
 						// normal value and at the end of the path
 						if (value != null) {
-							collectedValues.add(value);
+							if (value instanceof InstanceDecorator) {
```
stempler, line 1060: If the `subPath` is empty, then an `Instance` should actually be handled by the `value instanceof Group` condition above, where `addValues` is called again. There, with an empty path it then depends on the value of the parameter `onlyValues` if the `Instance` itself is added or its value extracted.

common/plugins/eu.esdihumboldt.hale.common.align/src/eu/esdihumboldt/hale/common/align/model/AlignmentUtil.java
```diff
@@ -1045,13 +1046,23 @@ public static void addValues(Group group, List<ChildContext> path,
 
 				// check all values
 				for (Object value : values) {
-					if (value instanceof Group) {
+					if (value instanceof InstanceDecorator) {
+						if (((InstanceDecorator) value).getValue() != null) {
+							collectedValues.add(((InstanceDecorator) value).getValue());
```
stempler, line 1051: Looking at the method signature it seems to me that the `onlyValues` parameter is what is used here to control if it is OK to return an `Instance` or rather it's value.
See here: https://github.com/halestudio/hale/blob/5cec368a3da576f45c6b100adc378611d12b54cc/common/plugins/eu.esdihumboldt.hale.common.align/src/eu/esdihumboldt/hale/common/align/model/AlignmentUtil.java#L1002

Do you have a hint from where `addValues` is called in your case and how it is determined what value is set for the `onlyValues` parameter?

common/plugins/eu.esdihumboldt.hale.common.align/src/eu/esdihumboldt/hale/common/align/model/AlignmentUtil.java
```diff
@@ -1045,13 +1046,23 @@ public static void addValues(Group group, List<ChildContext> path,
 
 				// check all values
 				for (Object value : values) {
-					if (value instanceof Group) {
+					if (value instanceof InstanceDecorator) {
```
stempler, line 1049: Directly extracting the value here seems to ignore the `path` parameter that is passed in. It looks to me like that could lead to wrong results.

stempler (commented):
Were you able to set up a transformation test demonstrating the problem?
That is also something that would make it easier for others to support in the investigation.

stempler:
@emanuelaepure10 Are you sure it is not just that in the function implementations there are `Instance`s passed in but then worked with as if they are normal values? I could imagine this for instance happening [here](https://github.com/halestudio/hale/blob/5e88ffee6ff00f4710708d83d290b91135f1a2b8/cst/plugins/eu.esdihumboldt.cst.functions.core/src/eu/esdihumboldt/cst/functions/core/FormattedString.java#L65).

emanuelaepure10:
> Are you sure it is not just that in the function implementations there are Instances passed in but then worked with as if they are normal values? I could imagine this for instance happening [here](https://github.com/halestudio/hale/blob/5e88ffee6ff00f4710708d83d290b91135f1a2b8/cst/plugins/eu.esdihumboldt.cst.functions.core/src/eu/esdihumboldt/cst/functions/core/FormattedString.java#L65).

Doesn't seem to be this line to convert it to a String. For a DatabaseHandle$InstanceHandle trying to convert it to a string in the above line results into an error and goes into catch so the correct value is then assigned to the value.

PR #1122, opened 2024-02-29
build: migrate build to GitHub Actions

emanuelaepure10:
Add check for all the incoming pull requests and for building the master branch.

ING-4164

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -1,43 +1,43 @@
-name: Check
+name: Pull Request Workflow
 
-# XXX disabled for now because we use Jenkins still, but at the time this was tested it was functional
-# on:
-#   pull_request:
-#     branches: [ master ]
+on:
+  pull_request:
+    branches:
+      - '*'  # Trigger on all branches for pull requests
 
 jobs:
-  check:
+  build:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v3
+      - name: Checkout code
+        uses: actions/checkout@v4
 
-    # Note: Actually we don't need Maven CLI, only Java 17, but this may have additional benefits due to the Maven cache
-    - name: Setup Maven
-      uses: s4u/setup-maven-action@v1.5.1
-      with:
-        java-version: 17
-        java-distribution: temurin
-        maven-version: 3.8.6
+      - name: Setup Maven
+        uses: s4u/setup-maven-action@v1.5.1
+        with:
+          java-version: 17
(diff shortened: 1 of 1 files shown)


.github/workflows/check.yml
```diff
@@ -1,43 +1,87 @@
...
-      if: always() # always run even if the previous step fails
-      with:
-        report_paths: 'build/target/testReports/*.xml'
+      - name: Publish Test Report
+        uses: mikepenz/action-junit-report@v3
+        if: always() # always run even if the previous step fails
+        with:
+          report_paths: 'build/target/testReports/*.xml'
 
+  build_master:
```
stempler, line 46: I would be easier to have a separate workflow for what is to be done when something is pushed to master.

Right now this job looks much like a copy of the other job, but it should include everything that is currently done in the [Jenkins task](https://builds.wetransform.to/job/hale/job/hale%7Epublish(master)/).

.github/workflows/check.yml
```diff
@@ -1,43 +1,87 @@
 name: Check
 
-# XXX disabled for now because we use Jenkins still, but at the time this was tested it was functional
-# on:
-#   pull_request:
-#     branches: [ master ]
+on:
+  pull_request:
```
stempler, line 4: Tip: To test the on pull request trigger it's easiest if you push your changes to a branch in the main repository instead of your fork. Then it will already consider your changes to the workflow.

.github/workflows/check.yml
```diff
@@ -1,43 +1,87 @@
...
+    branches:
+      - '*'  # Trigger on all branches for pull requests
+  push:
```
stempler, line 7: As mentioned in another comment I would suggest to move this trigger and the related actions to another workflow.
If you push your changes to a specific branch in the main repository, you can test the workflow by adding the branch you are using to the list of branches.

.github/workflows/check.yml
```diff
@@ -1,43 +1,87 @@
...
+      - name: Setup Maven
+        uses: s4u/setup-maven-action@v1.5.1
+        with:
+          java-version: 17
+          java-distribution: temurin
+          maven-version: 3.8.6
+          notify-failure: false
+          skip-build: false
+          expect-tests: false
+          submodules: 'recursive'
```
stempler, line 28: I don't see the input parameters you added as supported parameters for the [action](https://github.com/marketplace/actions/setup-maven-action).

stempler:
Closing this as it seems to be superseded by #1124

PR #1128, opened 2024-03-13
fix: don't close PrintStream when using printCell

stempler:
...to avoid closing System.out which is used in most cases as target for printing a Cell in tests, which lead to incomplete output from tests.


emanuelaepure10 (approved):
LGTM

PR #1130, opened 2024-03-13
fix: allow filtering for joins

stempler:
...when selecting a source type in the alignment view.

Before no results would be shown because the internal check expects all types to match while in the view only one can be specified.


emanuelaepure10 (approved):

florianesser (approved):
LGTM

PR #1159, opened 2024-04-18
feat: add trivy vulnerability check

emanuelaepure10:
Trivy vulnerability check is added the GitHub Actions workflows.

ING-4183


trivy-results.sarif
```diff
@@ -0,0 +1,185 @@
+{
```
stempler, line 1: I don't think this file should be added, rather added to `.gitignore`?

.github/workflows/trivy.yml
```diff
@@ -0,0 +1,72 @@
+# This workflow uses actions that are not certified by GitHub.
+# They are provided by a third-party and are governed by
+# separate terms of service, privacy policy, and support
+# documentation.
+
+name: Trivy vulnerability scanner
```
stempler, line 6: Since we need a built hale studio to scan in `rootfs` mode I think it would make more sense to integrate the scan into a workflow that anyway already creates a hale studio artifact.

emanuelaepure10: sure. But this can be done as well after we understood which tool we like the most, right?

stempler: Do you refer to trivy and codeql?
In my view both serve different purposes so both should be included.

emanuelaepure10: Also in my view Trivy and CodeQL have different focuses, they complement each other very well and using them both we can cover both dependency vulnerabilities and code-level issues,  but I thought that trivy is favourite and we care less about the rest :-)

stempler: @emanuelaepure10 So can we proceed in that you adapt the trivy scan to scan the product, or what would be your plan?

.github/workflows/trivy.yml
```diff
@@ -0,0 +1,87 @@
...
+          #vuln-type: 'os,library'
+          #severity: 'CRITICAL,HIGH,MEDIUM'
+          #template: '/sarif.tpl'
+          #output: 'trivy-results.sarif'
+          #skip-dirs: "ignored-dir"
+          #trivy-config: trivy.yaml
+
+      - name: Run Trivy scan
+        run: |
+          trivy fs --target /home/runner/work/hale/hale --debug --format sarif --output trivy-results.sarif
```
emanuelaepure10, line 56: @stempler could you please help me understand which is the correct directory to scan? Thank you

stempler: We had discussed in a call around May 8th that a built product should be scanned, why I suggested [here](https://github.com/halestudio/hale/pull/1159/files#r1594056229) to integrate the scan into a workflow where products are built anyway.
There is also an [earlier discussion](https://wetransform.slack.com/archives/C06JBBH0CTA/p1714399025710769?thread_ts=1714383569.075639&cid=C06JBBH0CTA) where I asked about what you are scanning and referred to a [comment on a previous ticket](https://wetransform.atlassian.net/browse/ING-3934?focusedCommentId=80935) that refers to [how we did scans manually in the past](https://wetransform.slack.com/archives/C1GEA01EE/p1687179420688989?thread_ts=1687178913.974469&cid=C1GEA01EE). Did you take a look at these resources? Do you have specific open questions?

.github/workflows/check.yml
```diff
@@ -110,6 +110,27 @@ jobs:
...
+          sudo apt-get update
+          sudo apt-get install -y trivy
+          trivy --version
+
+      - name: Run Trivy scan
+        run: |
+          trivy fs --severity CRITICAL,HIGH build/target/hale-studio-*linux*.tar.gz --format sarif --output trivy-results.sarif
+
+
+      - name: Upload Trivy scan results to file
```
emanuelaepure10, line 128: @stempler Do you maybe see why the file is not updated with the results?

.github/workflows/check.yml
```diff
@@ -110,6 +110,27 @@ jobs:
           path: build/target/hale-studio-*macosx*.dmg
           retention-days: 14
 
+      - name: Set up Trivy
```
stempler, line 113: Why don't you use the trivy GitHub action as before?

.github/workflows/check.yml
```diff
@@ -110,6 +110,27 @@ jobs:
...
+          sudo apt-get install -y wget apt-transport-https gnupg lsb-release
+          wget -qO - https://aquasecurity.github.io/trivy-repo/deb/public.key | sudo apt-key add -
+          echo deb https://aquasecurity.github.io/trivy-repo/deb $(lsb_release -sc) main | sudo tee -a /etc/apt/sources.list.d/trivy.list
+          sudo apt-get update
+          sudo apt-get install -y trivy
+          trivy --version
+
+      - name: Run Trivy scan
+        run: |
+          trivy fs --severity CRITICAL,HIGH build/target/hale-studio-*linux*.tar.gz --format sarif --output trivy-results.sarif
```
stempler, line 125: Did you test if it works on the archive or would it need to be extracted first?

.github/workflows/check.yml
stempler: It seems the upload works but there are are just no results - I would guess maybe because the product archive is not extracted.

Would the goal not be to use the `github/codeql-action/upload-sarif` action again to integrate with GitHub code scanning?

.github/workflows/check.yml
emanuelaepure10: If I don't set up/install trivy  I get an error as:
```
Run trivy fs --severity CRITICAL,HIGH build/target/hale-studio-linux-trivy --format sarif --output trivy-results.sarif
/home/runner/work/_temp/1ea9265f-f612-4cca-8[7](https://github.com/halestudio/hale/actions/runs/9449564750/job/26026128237?pr=1159#step:10:8)d1-b6e3a263f8f2.sh: line 1: trivy: command not found
```
See [here](https://github.com/halestudio/hale/actions/runs/9454619259/job/26042573968?pr=1159)

.github/workflows/check.yml
emanuelaepure10: I've tried but I got the same result.

emanuelaepure10:
@stempler I'm unsure how to proceed with this PR and would greatly appreciate your assistance. Thank you!

stempler:
> I'm unsure how to proceed with this PR and would greatly appreciate your assistance. Thank you!

@emanuelaepure10  From my perspective the open tasks would be:

- use the trivy GitHub action (as you had done in the beginning) instead of installing it and running it via the CLI
- I suspect the `fs` scan does not yield any results, at least that was the case for me [when I tested this](https://wetransform.slack.com/archives/C1GEA01EE/p1687179420688989?thread_ts=1687178913.974469&cid=C1GEA01EE), so I suspect `rootfs` has to be used instead
- looking at the code scanning section it seems that in the end, the trivy scan should be running for the master branch and not only for the PR, but this can be done once the process is working

How would one usually find the code scanning results for a PR (e.g. is there a link somewhere?)
I found it not so easy to find the [info for this PR](https://github.com/halestudio/hale/security/code-scanning?query=is%3Aopen+pr%3A1159).

emanuelaepure10:
@stempler I have updated the PR and I completely agree with you that would be so much nicer to have a link directly from the PR to the code scanning results. At the moment when you go to the security tab -> code scanning you could select the PR or the branch you are interested in
image
but working with forks you would probably need to select always the PR instead of a branch. 
Another option could be to create a PR with the vulnerabilities found. 
What would you prefer?

stempler (dismissed):

stempler:
Replaced by #1204

PR #989, opened 2022-09-29
feat: add Throwable class to whitelist

JohannaOtt:
Currently, it is not possible to add the message of an exception to the transformation's logs as the class is not whitelisted

SVC-1190

PR #1166, opened 2024-04-24
feat: bind lookup service to Groovy constant for use in scripts

florianesser:
ING-4256

### DIFF
--- a/cst/plugins/eu.esdihumboldt.cst.functions.groovy/src/eu/esdihumboldt/cst/functions/groovy/internal/GroovyUtil.java
+++ b/cst/plugins/eu.esdihumboldt.cst.functions.groovy/src/eu/esdihumboldt/cst/functions/groovy/internal/GroovyUtil.java
@@ -43,6 +43,7 @@ import eu.esdihumboldt.hale.common.align.transformation.report.TransformationLog
 import eu.esdihumboldt.hale.common.core.io.Text;
 import eu.esdihumboldt.hale.common.core.io.project.ProjectInfoService;
 import eu.esdihumboldt.hale.common.core.report.SimpleLog;
+import eu.esdihumboldt.hale.common.core.service.DelegateServiceProvider;
 import eu.esdihumboldt.hale.common.instance.groovy.InstanceBuilder;
 import eu.esdihumboldt.hale.common.instance.index.InstanceIndexService;
 import eu.esdihumboldt.hale.common.instance.index.spatial.SpatialIndexService;
@@ -304,6 +305,9 @@ public class GroovyUtil implements GroovyConstants {
 
 		binding.setVariable(BINDING_INSTANCE_INDEX,
 				executionContext.getService(InstanceIndexService.class));
+
+		binding.setVariable(BINDING_SERVICE_PROVIDER,
+				new DelegateServiceProvider(executionContext));
 
 		return binding;
 	}
--- a/common/plugins/eu.esdihumboldt.hale.common.core/plugin.xml
+++ b/common/plugins/eu.esdihumboldt.hale.common.core/plugin.xml
@@ -225,4 +225,11 @@
             implementation="eu.esdihumboldt.hale.common.core.io.internal.AsValueMetaClass">
       </metaClass>
    </extension>
+   <extension
+         point="eu.esdihumboldt.util.groovy.sandbox">
+      <allow
+            allowAll="false"
+            class="eu.esdihumboldt.hale.common.core.service.DelegateServiceProvider">
+      </allow>
+   </extension>
 </plugin>
--- a/common/plugins/eu.esdihumboldt.hale.common.core/src/eu/esdihumboldt/hale/common/core/service/DelegateServiceProvider.java
+++ b/common/plugins/eu.esdihumboldt.hale.common.core/src/eu/esdihumboldt/hale/common/core/service/DelegateServiceProvider.java
@@ -1,0 +1,44 @@
+/*
+ * Copyright (c) 2024 wetransform GmbH
+ * 
+ * All rights reserved. This program and the accompanying materials are made
+ * available under the terms of the GNU Lesser General Public License as
+ * published by the Free Software Foundation, either version 3 of the License,
+ * or (at your option) any later version.
+ * 
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this distribution. If not, see <http://www.gnu.org/licenses/>.
+ * 
+ * Contributors:
+ *     wetransform GmbH <http://www.wetransform.to>
+ */
+
+package eu.esdihumboldt.hale.common.core.service;
+
+/**
+ * A DelegateServiceProvider delegates service requests to an underlying
+ * ServiceProvider. It implements the ServiceProvider interface and forwards
+ * getService calls to the specified underlying ServiceProvider.
+ * 
+ * @author EmanuelaEpure
+ */
+public class DelegateServiceProvider implements ServiceProvider {
+
+	private final ServiceProvider serviceProvider;
+
+	/**
+	 * Constructs a new DelegateServiceProvider with the specified
+	 * ServiceProvider.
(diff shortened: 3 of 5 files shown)


cst/plugins/eu.esdihumboldt.cst.functions.groovy/src/eu/esdihumboldt/cst/functions/groovy/internal/GroovyUtil.java
```diff
@@ -305,6 +306,9 @@ public static Binding createBinding(InstanceBuilder builder, Cell cell, Cell typ
 		binding.setVariable(BINDING_INSTANCE_INDEX,
 				executionContext.getService(InstanceIndexService.class));
 
+		binding.setVariable(BINDING_LOOKUP_SERVICE,
+				executionContext.getService(LookupService.class));
```
stempler, line 310: Do you think a specific binding is preferable over the more generic approach to add a binding for the `ServiceProvider`? (see [ING-4256](https://wetransform.atlassian.net/browse/ING-4256))

florianesser: No, if we can find a solution that avoids having to add individual bindings for every  service class that we want to use in Groovy scripts, that would be better IMO. Happy to close this PR in favour of a more generic approach!

emanuelaepure10: Would it be alright if we squash the last two commits?

florianesser: > Would it be alright if we squash the last two commits?

@emanuelaepure10 Yes, no need to keep my initial commit.

cst/plugins/eu.esdihumboldt.cst.functions.groovy/src/eu/esdihumboldt/cst/functions/groovy/internal/GroovyUtil.java
```diff
@@ -305,6 +306,9 @@ public static Binding createBinding(InstanceBuilder builder, Cell cell, Cell typ
 		binding.setVariable(BINDING_INSTANCE_INDEX,
 				executionContext.getService(InstanceIndexService.class));
 
+		binding.setVariable(BINDING_SERVICE_PROVIDER,
+				executionContext.getService(ServiceProvider.class));
```
florianesser, line 310: @emanuelaepure10 I think it must be

```
		binding.setVariable(BINDING_SERVICE_PROVIDER, executionContext);
```

instead as `ExecutionContext` implements the `ServiceProvider` interface. That way, one can do e.g. the following in Groovy scripts to get an instance of `LookupService`:

```
def lookupService = _serviceProvider.getService(LookupService.class)
```

stempler (commented):
I think to be usable this has to go in line with adding exceptions to the Groovy restrictions. At least for being able to call `getService` on the service provider.

cst/plugins/eu.esdihumboldt.cst.functions.groovy/src/eu/esdihumboldt/cst/functions/groovy/internal/GroovyUtil.java
```diff
@@ -305,6 +306,13 @@ public static Binding createBinding(InstanceBuilder builder, Cell cell, Cell typ
 		binding.setVariable(BINDING_INSTANCE_INDEX,
 				executionContext.getService(InstanceIndexService.class));
 
+		try {
+			binding.setVariable(BINDING_SERVICE_PROVIDER, executionContext);
```
emanuelaepure10, line 310: @stempler I'm not very sure if by "adding exceptions to the Groovy restrictions." is this what you meant.

florianesser: @emanuelaepure10 Simon refers to the Groovy restrictions that prevent using classes and methods in Groovy scripts that are not explicitly whitelisted. Check the class [RestrictiveGroovyInterceptor](https://github.com/halestudio/hale/blob/master/util/plugins/eu.esdihumboldt.util.groovy.sandbox/src/eu/esdihumboldt/util/groovy/sandbox/internal/RestrictiveGroovyInterceptor.java).

[This is an example for a PR](https://github.com/halestudio/hale/pull/989) where a class was whitelisted. As Simon suggested, in this case it might make sense to investigate if only calling `getService(...)` should be allowed in contrast to whitelisting access to all methods of the `ServiceProvider`.

stempler: To add to that likely the check is done against the concrete implementation and not the interface (`ServiceProvider`) where the method is defined.
One solution for that could be creating a delegate class that implements the interface, then the delegate can be whitelisted and passed in here, regardless of what the actual implementation class of the execution context is here and what other methods it provides.

util/plugins/eu.esdihumboldt.util.groovy.sandbox/META-INF/MANIFEST.MF
```diff
@@ -9,6 +9,7 @@ Require-Bundle: groovy;bundle-version="2.5.19",
 Export-Package: eu.esdihumboldt.util.groovy.sandbox,
  eu.esdihumboldt.util.groovy.sandbox.internal;x-internal:=true
 Import-Package: de.fhg.igd.eclipse.util.extension,
+ eu.esdihumboldt.hale.common.core.service,
```
stempler, line 12: Instead of adding a dependency here, the allowed class should be added via the [extension point](https://github.com/halestudio/hale/blob/85dfae5fc2979cc885fd511f267874a49045fe44/util/plugins/eu.esdihumboldt.util.groovy.sandbox/plugin.xml#L4) in the bundle where the `DelegateServiceProvider` class is defined.

See [here](https://github.com/halestudio/hale/blob/d56e083333fd46f309c9088eee4820de9f44ae54/common/plugins/eu.esdihumboldt.hale.common.align/plugin.xml#L639) for an example.

stempler (changes requested):

stempler (approved):
LGTM, but I could imagine that for the use case with the lookup tables it could be necessary to add additional classes to be excluded from the Groovy restrictions, e.g  the [LookupServiceImpl](https://github.com/halestudio/hale/blob/master/common/plugins/eu.esdihumboldt.hale.common.lookup/src/eu/esdihumboldt/hale/common/lookup/internal/LookupServiceImpl.java).
Did you test if usage of the service is possible with active Groovy restrictions?

emanuelaepure10:
I have test it with and without `-trustGroovy`

emanuelaepure10:
Do you have any example of how to exclude a class from the Groovy restrictions?
My research bring me nowhere. 
Thank you

stempler:
> Do you have any example of how to exclude a class from the Groovy restrictions?
My research bring me nowhere.

What you did for the `DelegateServiceProvider` [here](https://github.com/halestudio/hale/pull/1166/files#diff-f6127e936c7c31af075fa32374dc76ff485bec3d857ab60c327c993969346bbcR228-R234).

emanuelaepure10:
But... the `DelegateServiceProvider` I add it to the Groovy restrictions, I didn't **exclude it**.

stempler:
> But... the DelegateServiceProvider I add it to the Groovy restrictions, I didn't exclude it.

By adding it to the allowed classes, you excluded it from the restrictions when executing Groovy. It can be accessed without lifting the restrictions.

PR #1170, opened 2024-05-07
ci: pass Artifactory user and password to deployArtifacts task

florianesser:
It looks like the Artifactory secrets are not passed to the `deployArtifacts` task, so adding them to `env` explicitly here. If this solves the problem that the `uploadEclipse` step is currently not executed remains to be seen.


stempler (approved):

emanuelaepure10:
Probably same change should be as well done in the `relasese.yml`.

PR #1171, opened 2024-05-07
fix: use appropriate methods to convert the numbers in XML

emanuelaepure10:
Conversion of the numbers using appropriate methods of each different type.

ING-3965

### DIFF
--- a/io/plugins/eu.esdihumboldt.hale.io.gml.test/src/data/shiporder/shiporder.xml
+++ b/io/plugins/eu.esdihumboldt.hale.io.gml.test/src/data/shiporder/shiporder.xml
@@ -14,11 +14,16 @@ xsi:schemaLocation="http://www.example.com shiporder.xsd">
     <title>Empire Burlesque</title>
     <note>Special Edition</note>
     <quantity>1</quantity>
-    <price>10.90</price>
+    <price>0E-11</price>
   </item>
   <item>
     <title>Hide your heart</title>
     <quantity>1</quantity>
     <price>9.90</price>
+    <priceInteger>123</priceInteger>
+    <priceDouble>12.3</priceDouble>
+    <priceFloat>123.456</priceFloat>
+    <priceLong>1234567890123456789</priceLong>
+    <priceShort>12345</priceShort>
   </item>
 </shiporder>
--- a/io/plugins/eu.esdihumboldt.hale.io.gml.test/src/eu/esdihumboldt/hale/io/gml/reader/internal/GmlInstanceCollectionTest.java
+++ b/io/plugins/eu.esdihumboldt.hale.io.gml.test/src/eu/esdihumboldt/hale/io/gml/reader/internal/GmlInstanceCollectionTest.java
@@ -22,6 +22,7 @@ import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
 
 import java.io.IOException;
+import java.math.BigDecimal;
 import java.net.URI;
 import java.util.Collection;
 
@@ -160,10 +161,12 @@ public class GmlInstanceCollectionTest {
 			Instance instance = it.next();
 			assertNotNull(instance);
 
-			Object[] orderid = instance.getProperty(new QName("orderid")); // attribute
-																			// form
-																			// not
-																			// qualified
+			Object[] orderid = instance.getProperty(new QName("orderid"));
+			// attribute
+			// form
+			// not
+			// qualified
+
 			assertNotNull(orderid);
 			assertEquals(1, orderid.length);
 			assertEquals("889923", orderid[0]);
@@ -213,6 +216,11 @@ public class GmlInstanceCollectionTest {
 			assertEquals(1, title1.length);
 			assertEquals("Empire Burlesque", title1[0]);
 
+			Object[] price1 = ((Instance) item1).getProperty(new QName(ns, "price"));
+			assertNotNull(price1);
+			BigDecimal bigDecimal = new BigDecimal("0.00000000000");
+			assertEquals(bigDecimal, price1[0]);
+
 			// item 2
 			Object item2 = items[1];
 			assertTrue(item2 instanceof Instance);
@@ -221,6 +229,35 @@ public class GmlInstanceCollectionTest {
 			assertNotNull(title2);
 			assertEquals(1, title2.length);
 			assertEquals("Hide your heart", title2[0]);
+
+			Object[] price2 = ((Instance) item2).getProperty(new QName(ns, "price"));
+			assertNotNull(price2);
+			bigDecimal = new BigDecimal("9.90");
+			assertEquals(bigDecimal, price2[0]);
+
+			Object[] priceInteger = ((Instance) item2).getProperty(new QName(ns, "priceInteger"));
+			int intNumber = 123;
+			assertEquals(intNumber, priceInteger[0]);
+
+			// Double
+			Object[] priceDouble = ((Instance) item2).getProperty(new QName(ns, "priceDouble"));
+			double doubleNumber = 12.3;
+			assertEquals(doubleNumber, priceDouble[0]);
+
+			// Float
+			Object[] priceFloat = ((Instance) item2).getProperty(new QName(ns, "priceFloat"));
+			float floatNumber = 123.456F;
+			assertEquals(floatNumber, priceFloat[0]);
+
+			// Long
+			Object[] priceLong = ((Instance) item2).getProperty(new QName(ns, "priceLong"));
+			long longNumber = 1234567890123456789L;
+			assertEquals(longNumber, priceLong[0]);
+
+			// short
+			Object[] priceShort = ((Instance) item2).getProperty(new QName(ns, "priceShort"));
+			short shortNumber = 12345;
+			assertEquals(shortNumber, priceShort[0]);
 
 			// only one object
 			assertFalse(it.hasNext());
--- a/io/plugins/eu.esdihumboldt.hale.io.gml.test/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriter2Test.groovy
+++ b/io/plugins/eu.esdihumboldt.hale.io.gml.test/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriter2Test.groovy
@@ -381,4 +381,151 @@ class StreamGmlWriter2Test {
 
 		result
 	}
+
+	@Test
+	void testNumbers() throws Exception {
+		/**test DOUBLE numbers*/
+		testNumber("doubleNumber", 12345.123456789, null)
+		testNumber("doubleNumber", 12345.123456789, "12345.123456789")
+		double doubleNumber = 123456789.123456789
+		double expected = java.lang.Double.parseDouble(doubleNumber.toString())
+		testNumber("doubleSimpleNumber", doubleNumber, expected)
+		testNumber("doubleSimpleNumber", doubleNumber, expected.toString())
+		testFloatingPointNumber(doubleNumber, "doubleNumber")
+
+		/**test INT, SHORT numbers*/
+		testNumber("integerNumber", 12345, null)
+		testNumber("integerNumber", 12345, "12345")
+		testNumber("shortNumber", 12345 as short, null)
+		testNumber("shortNumber", 12345 as short, "12345")
+
+		float smallDecimal = 0.12345F
+		float largeDecimal = 12345.6789F
+		float verySmall = 0.0000001F
+		float veryLarge = 12345678.9F
+		float negative = -98765.4321F
+		float scientific = 1.23E4F
+		float maxFloat = Float.MAX_VALUE
+		float minFloat = Float.MIN_VALUE
+
+		// for Float and Double we are expecting to use using scientific notation
+		testFloatingPointNumber(smallDecimal, "floatNumber")
+		testNumber("floatNumber", smallDecimal, "0.12345")
+		testFloatingPointNumber(largeDecimal, "floatNumber")
+		testNumber("floatNumber", largeDecimal, "12345.679")
+		testFloatingPointNumber(verySmall, "floatNumber")
+		testNumber("floatNumber", verySmall, "1.0E-7")
+		testFloatingPointNumber(veryLarge, "floatNumber")
+		testNumber("floatNumber", veryLarge, "1.2345679E7")
+		testFloatingPointNumber(negative, "floatNumber")
+		testFloatingPointNumber(scientific, "floatNumber")
+		testFloatingPointNumber(maxFloat, "floatNumber")
+		testFloatingPointNumber(minFloat, "floatNumber")
+
+		/**test LONG numbers*/
+		long smallPositive = 12345L
+		long largePositive = 1234567890123456789L
+		long smallNegative = -12345L
+		long largeNegative = -1234567890123456789L
+		long maxLong = Long.MAX_VALUE
+		long minLong = Long.MIN_VALUE
+		long zero = 0L
+		long powerOfTwo = 1024L
+		long negativePowerOfTwo = -1024L
+		long nearMaxLong = 9223372036854775806L
+		long nearMinLong = -9223372036854775807L
+
+		testNumber("longNumber", smallPositive, "12345")
+		testNumber("longNumber", largePositive, "1234567890123456789")
+		testNumber("longNumber", smallNegative, null)
+		testNumber("longNumber", largeNegative, null)
+		testNumber("longNumber", maxLong, null)
+		testNumber("longNumber", minLong, null)
+		testNumber("longNumber", zero, null)
+		testNumber("longNumber", powerOfTwo, null)
+		testNumber("longNumber", negativePowerOfTwo, null)
+		testNumber("longNumber", nearMaxLong, null)
+		testNumber("longNumber", nearMinLong, null)
+
+		/**test DECIMAL numbers*/
+		testNumber("decimalNumber", new BigDecimal("1234567890.123456789"), null)
+		testNumber("decimalNumber", new BigDecimal(doubleNumber), null)
+
+		doubleNumber = 1.23456789123456789E8
+		testNumber("decimalNumber", new BigDecimal(doubleNumber), null)
+
+		String numberAsString = "1.23456789123456789E8"
+		testNumber("decimalNumber", new BigDecimal(numberAsString), null)
+		testNumber("decimalNumber", new BigDecimal(numberAsString), "123456789.123456789")
+	}
+
+	void testNumber(String elementName, number, expected) throws Exception {
+		def schema = loadSchema(getClass().getResource("/data/numbers/numbers.xsd").toURI())
+
+		def instance = new InstanceBuilder(types: schema).PrimitiveTestType {
+			"$elementName"(number)
+		}
+
+		def writer = new GmlInstanceWriter()
+		File outFile = writeFile(instance, schema, writer)
+		def xmlFile = outFile.getAbsolutePath()
+		def xml = new XmlSlurper().parse(xmlFile)
+
+		String actualText = xml.featureMember.PrimitiveTest."$elementName".text()
+		if (expected != null) {
+			if (expected instanceof String) {
+				assertEquals(expected.toString(), xml.featureMember.PrimitiveTest."$elementName".text())
+			} else {
+				compareValues(expected, actualText)
+			}
+		} else {
+			compareValues(number, actualText)
+		}
+
+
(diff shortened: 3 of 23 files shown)


io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/internal/simpletype/SimpleTypeUtil.java
```diff
@@ -119,6 +132,31 @@ else if (simpleTypeValue instanceof XmlDateTime) {
...
+					|| value instanceof XmlInteger) {
+				// Return the string value of the number
+				return String.valueOf(numberValue);
+			}
+			else if (value instanceof XmlDecimal) {
+				BigDecimal decimalValue = ((XmlDecimal) value).getBigDecimalValue();
+				return decimalValue.stripTrailingZeros().toPlainString();
+			}
+		}
+
```
stempler, line 159: The intention with the use of the `xmlbeans` simple type classes is to use them to [get the correct string representation](https://github.com/halestudio/hale/blob/c5a92100ad860826d107a13a0dd20084a727f948/io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/internal/simpletype/SimpleTypeUtil.java#L115) to use in the XML.

You already extended the type map with the respective classes, and this is used [here](https://github.com/halestudio/hale/blob/c5a92100ad860826d107a13a0dd20084a727f948/io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/internal/simpletype/SimpleTypeUtil.java#L82).
But instead of the custom handling that you added here now, what is missing are converters that convert the related type bindings to and from the XML simple type classes. Like for example these converters from `Date` to `XmlDateTime` and vice versa:

- [DateToXmlDateTime](https://github.com/halestudio/hale/blob/master/io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/internal/simpletype/converters/DateToXmlDateTime.java)
- [XmlDateTimeToDate](https://github.com/halestudio/hale/blob/master/io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/internal/simpletype/converters/XmlDateTimeToDate.java)

To make converters available there is an existing extension point to register the converters. See example [here](https://github.com/halestudio/hale/blob/c5a92100ad860826d107a13a0dd20084a727f948/io/plugins/eu.esdihumboldt.hale.io.gml/plugin.xml#L1495).

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/internal/simpletype/SimpleTypeUtil.java
```diff
@@ -119,6 +134,19 @@ else if (simpleTypeValue instanceof XmlDateTime) {
 			}
 		}
 
+		if (value instanceof BigDecimal) {
+			// Create a DecimalFormat object without scientific notation
+			DecimalFormat decimalFormat = new DecimalFormat("0.##############");
+			// Format the BigDecimal
+			return decimalFormat.format(value);
+		}
```
stempler, line 142: If the XML element/attribute has a number type, shouldn't it be handled by an xmlbeans type conversion before? I think if we still end up here, we should rather add the missing conversion.

What happens if the number has more decimal places than in the format, will it be truncated?

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/internal/simpletype/SimpleTypeUtil.java
```diff
@@ -119,6 +134,19 @@ else if (simpleTypeValue instanceof XmlDateTime) {
...
+		else
+		// Handle primitive types
+		if (value instanceof Double || value instanceof Float || value instanceof Long
+				|| value instanceof Integer) {
+			return String.valueOf(value);
+		}
```
stempler, line 148: If the XML element/attribute has a number type, shouldn't it be handled by an xmlbeans type conversion before? I think if we still end up here, we should rather add the missing conversion.

Also this representation would basically be what was used originally, which can result in scientific notation.

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/internal/simpletype/SimpleTypeUtil.java
emanuelaepure10: True. It's handled by the new xmlbeans.

stempler (commented):
Generally looks good to me, but I have some doubts if the added converters are sufficient.

As mentioned previously it would be great to have unit tests to accompany your implementation to (1) verify it and (2) prevent future regressions.

Related to the conversions there could be other types of values/bindings for the XML simple types, than what you implemented now.
When we load an XML schema we [here](https://github.com/halestudio/hale/blob/e2c5dd54f31482eff6beee6541b2dd14dfeaee21/io/plugins/eu.esdihumboldt.hale.io.xsd/src/eu/esdihumboldt/hale/io/xsd/reader/internal/XmlTypeUtil.java#L209) use functionality from Geotools to determine the binding. As an example, for the integer simple type a [BigInteger binding](https://github.com/geotools/geotools/blob/38b3f7481a795a60a8bf91b8c0e00bd45e5ab83d/modules/extension/xsd/xsd-core/src/main/java/org/geotools/xs/XSSchema.java#L203) is used. So I think at least the respective binding types from that class need to be supported for the conversion.

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/internal/simpletype/converters/BigDecimalToXmlDecimal.java
```diff
@@ -0,0 +1,46 @@
+/*
```
stempler, line 1: Headers of new files are outdated.

stempler (commented):
Points from my previous review comment are still open - any feedback on those?
See https://github.com/halestudio/hale/pull/1171#pullrequestreview-2045683270

emanuelaepure10:
If you refer to the points about testing, yes there are yet no test done. 
Would you have an example of something similar from where to get an idea on how to create these kind of test?

stempler:
> If you refer to the points about testing, yes there are yet no test done.

Yes, I meant that but also the other mentioned point on the conversions, which should at least cover the conversions between the bindings we use for the XML simple types and the XmlBeans classes. The example I mentioned was that for the integer XML simple type the binding BigInteger is used.
Though it might be possible that this is covered by a transformation chain with intermediate types? If that's the case then this could be verified with the unit tests. Those should ideally include at least one test for every number related simple type.

> Would you have an example of something similar from where to get an idea on how to create these kind of test?

I would propose to look at [`StreamGmlWriter2Test`](https://github.com/halestudio/hale/blob/master/io/plugins/eu.esdihumboldt.hale.io.gml.test/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriter2Test.groovy) for examples of tests writing XML/GML. Though in your case for the validation you would want to check the written XML instead of loading it as instance, since you want to verify how it is represented in XML as text.

emanuelaepure10:
> Yes, I meant that but also the other mentioned point on the conversions, which should at least cover the conversions between the bindings we use for the XML simple types and the XmlBeans classes. The example I mentioned was that for the integer XML simple type the binding BigInteger is used.

Why for an integer cannot be used Integer instead of BigInteger?

> I would propose to look at [StreamGmlWriter2Test](https://github.com/halestudio/hale/blob/master/io/plugins/eu.esdihumboldt.hale.io.gml.test/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriter2Test.groovy) for examples of tests writing XML/GML. Though in your case for the validation you would want to check the written XML instead of loading it as instance, since you want to verify how it is represented in XML as text.

So that means that  [StreamGmlWriter2Test](https://github.com/halestudio/hale/blob/master/io/plugins/eu.esdihumboldt.hale.io.gml.test/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriter2Test.groovy) doesn't help me.  
There is anything around that would check the written XML?

stempler:
@emanuelaepure10 It would be great if you would use fixup commits while working on the PR because that makes reviews much easier.

> Why for an integer cannot be used Integer instead of BigInteger?

I would guess that the reason they did that in Geotools might have been that they wanted to be able to represent numbers that are outside the range supported by integer.

> So that means that [StreamGmlWriter2Test](https://github.com/halestudio/hale/blob/master/io/plugins/eu.esdihumboldt.hale.io.gml.test/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriter2Test.groovy) doesn't help me.

Why?

You can in a similar way

- load an XML schema
- create test data programmatically
- write the data

The main difference would be that you validate the written data in a different way.

> There is anything around that would check the written XML?

An easy way to load and process XML is for instance the Groovy `XmlSlurper` class.

emanuelaepure10:
> It would be great if you would use fixup commits while working on the PR because that makes reviews much easier.

I'll do so. I knew that I should always squash them. 

> Yes, I meant that but also the other mentioned point on the conversions, which should at least cover the conversions between the bindings we use for the XML simple types and the XmlBeans classes.

I think that all of this has been already covered.

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/internal/simpletype/converters/BigIntegerToXmlBigInteger.java
```diff
@@ -0,0 +1,39 @@
...
+
+import java.math.BigInteger;
+
+import org.apache.xmlbeans.XmlInteger;
+import org.springframework.core.convert.converter.Converter;
+
+/**
+ * Convert xs:BigInteger to {@link XmlInteger}
+ */
+public class BigIntegerToXmlBigInteger implements Converter<BigInteger, XmlInteger> {
```
stempler, line 26: `BigIntegerToXmlBigInteger` -> `BigIntegerToXmlInteger`

stempler (changes requested):
Do you plan to still add tests?

Related to the commit message I think this is a `fix`, not a feature.

emanuelaepure10:
@stempler 
I have extended an existing test in GmlInstanceCollectionTest, as it seemed the most appropriate place from my perspective. Please review the solution and let me know if it aligns with your expectations.

Thank you.

io/plugins/eu.esdihumboldt.hale.io.gml.test/src/eu/esdihumboldt/hale/io/gml/reader/internal/GmlInstanceCollectionTest.java
```diff
@@ -162,8 +163,6 @@ public void testLoadShiporder() throws Exception {
 
 			Object[] orderid = instance.getProperty(new QName("orderid")); // attribute
 																			// form
-																			// not
```
stempler, line 165: I think the original comment referred to that no namespace needs to be provided because the attribute form as specified in the XML schema is not qualified (i.e. `unqualified`). If you remove the comment I would say it makes sense to remove it completely.

io/plugins/eu.esdihumboldt.hale.io.gml.test/src/data/shiporder/shiporder.xml
```diff
@@ -14,11 +14,16 @@ xsi:schemaLocation="http://www.example.com shiporder.xsd">
     <title>Empire Burlesque</title>
     <note>Special Edition</note>
     <quantity>1</quantity>
-    <price>10.90</price>
+    <price>0E-11</price>
```
stempler, line 17: The root issue is that this notation is actually not valid and that hale studio should not create such a notation.

Fine for me to add a test for reading it, nice that we can deal with it, but it should definitely be mentioned that this is not valid according to the schema, so this can't result in any misunderstandings.

What we need to test in addition though, is writing data that would have resulted in an invalid representation before your changes, and verify that the notation in the XML is valid (and not what we get when we read it with hale).

So as before my suggestion would be to (in addition) create tests similar to in `StreamGmlWriter2Test`, because they provide a good template for writing test data given an XML schema. You can create an XML schema that is suited for your specific test cases for it and create minimal instance examples as required. As mentioned, `XmlSlurper` would be an easy way to access the content of the created XML file to verify the string representation in the file.

io/plugins/eu.esdihumboldt.hale.io.gml.test/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriter2Test.groovy
```diff
@@ -381,4 +381,50 @@ class StreamGmlWriter2Test {
...
+
+		// Create the instance
+		def instance = new InstanceBuilder(types: schema).PrimitiveTestType {
+			"$elementName"(number)
+		}
+
+		// Write file, validate and load it
+		def writer = new GmlInstanceWriter()
+		def xmlFile = writeFile(instance, schema, writer).getAbsolutePath()
+		def xml = new XmlSlurper().parse(xmlFile)
```
stempler, line 407: The temporary file should be deleted when it is no longer needed, ideally also controlled by `DEL_TEMP_FILES` as in similar logic in this class.

io/plugins/eu.esdihumboldt.hale.io.gml.test/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriter2Test.groovy
```diff
@@ -381,4 +381,50 @@ class StreamGmlWriter2Test {
...
 	}
+
+	@Test
+	void testNumbers() throws Exception {
+		testNumber("integerNumber", 12345)
+		testNumber("doubleNumber", 12345.123456789)
+		testNumber("floatNumber", 12345.6789F)
+		testNumber("longNumber", 1234567890123456789L)
+		testNumber("shortNumber", 12345 as short)
+		testNumber("decimalNumber", new BigDecimal("1234567890.123456789"))
```
stempler, line 392: Would any of these tests before your changes resulted in a scientific notation being used?

In any case I think it would be good to have a test that explicitly uses a scientific notation as input value.

io/plugins/eu.esdihumboldt.hale.io.gml.test/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriter2Test.groovy
```diff
@@ -384,15 +384,25 @@ class StreamGmlWriter2Test {
...
-		testNumber("integerNumber", 12345)
-		testNumber("doubleNumber", 12345.123456789)
-		testNumber("floatNumber", 12345.6789F)
-		testNumber("longNumber", 1234567890123456789L)
-		testNumber("shortNumber", 12345 as short)
-		testNumber("decimalNumber", new BigDecimal("1234567890.123456789"))
+		double doubleNumber = 123456789.123456789
+		double expected = java.lang.Double.parseDouble(doubleNumber.toString())
+
+		testNumber("doubleSimpleNumber", doubleNumber, expected)
```
emanuelaepure10, line 390: I have added the expected result in case the double is a "longer" double.

io/plugins/eu.esdihumboldt.hale.io.gml.test/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriter2Test.groovy
emanuelaepure10: Done.

io/plugins/eu.esdihumboldt.hale.io.gml.test/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriter2Test.groovy
emanuelaepure10: @stempler would you have anything else to add here?

io/plugins/eu.esdihumboldt.hale.io.gml.test/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriter2Test.groovy
```diff
@@ -381,4 +381,72 @@ class StreamGmlWriter2Test {
...
+		testNumber("doubleNumber", 12345.123456789, null)
+		testNumber("floatNumber", 12345.6789F, null)
+		testNumber("longNumber", 1234567890123456789L, null)
+		testNumber("shortNumber", 12345 as short, null)
+		testNumber("decimalNumber", new BigDecimal("1234567890.123456789"), null)
+
+		testNumber("decimalNumber", new BigDecimal(doubleNumber), null);
+
+		doubleNumber = 1.23456789123456789E8
+		testNumber("decimalNumber", new BigDecimal(doubleNumber), null);
```
stempler, line 402: I think an additional test would be great where the scientific notation is passed into the `BigDecimal` constructor as a string.

io/plugins/eu.esdihumboldt.hale.io.gml.test/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriter2Test.groovy
```diff
@@ -381,4 +381,72 @@ class StreamGmlWriter2Test {
 
 		result
 	}
+
+	@Test
+	void testNumbers() throws Exception {
+		double doubleNumber = 123456789.123456789
+		double expected = java.lang.Double.parseDouble(doubleNumber.toString())
```
stempler, line 388: Maybe add a comment here, that this value is then expected to use scientific notation?
Otherwise I think it will be hard for someone else to understand later on why this test is done that way.

io/plugins/eu.esdihumboldt.hale.io.gml.test/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriter2Test.groovy
```diff
@@ -381,4 +381,72 @@ class StreamGmlWriter2Test {
...
+
+		testNumber("doubleSimpleNumber", doubleNumber, expected)
+
+		testNumber("integerNumber", 12345, null)
+		testNumber("doubleNumber", 12345.123456789, null)
+		testNumber("floatNumber", 12345.6789F, null)
+		testNumber("longNumber", 1234567890123456789L, null)
+		testNumber("shortNumber", 12345 as short, null)
+		testNumber("decimalNumber", new BigDecimal("1234567890.123456789"), null)
```
stempler, line 397: I think it would be better to really explicitly use an expected string representation instead of relying on the toString representation. With that we could be sure that the representation is as expected as string and not rely on toString to yield the values as expected. After all the issue was that `toString` was used before and it did not yield the desired values at least for the `decimal` type.

emanuelaepure10: I have removed the `toString` and noticed that [double and float are treated in the same way](https://github.com/halestudio/hale/blob/43771301efa9df1f02e8822f86ae313a63e16308/io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/writer/internal/StreamGmlWriter.java#L1624) I have add more tests for this 2 types.

stempler: I'm afraid I think you misunderstood me. It is important to check the actual text used in the XML representation, so my suggestion is to use concrete strings as expected values, to be sure the notation is as expected.

stempler (approved):

Issue #986, opened 2022-09-22
CSV writer: Reduce multi geometries to single geometries if no multi geometry is contained


JohannaOtt:
Similar to what is implemented for the GML writer, the CSV writer should use single geometries in the CSV files if the data only contains single geometries.
Multi geometries should only be used if there is at least one multi geometry in the data.

Flaminietta:
Can it be one issue for me? @thorsten-reitz @stempler @florianesser?

PR #1173, opened 2024-05-14
feat: CSV, Excel and JSON writer - reduce to single geometries if dat…

emanuelaepure10:
…a does not contain multi geometry

CSV, Excel and JSON writer have been reduced to single geometries if data contains single geometry.

ING-3570
Closes [#986](https://github.com/halestudio/hale/issues/986)


io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/writer/AbstractTableInstanceWriter.java
```diff
@@ -225,20 +234,40 @@ private boolean shouldBeDisplayed(Object obj) {
 	}
 
 	private void addProperty(List<String> headerRow, Map<String, Object> row, Object property,
-			String propertyTypeName) {
+			String propertyTypeName, IOReporter reporter) {
 		if (!headerRow.contains(propertyTypeName)) {
 			headerRow.add(propertyTypeName);
 		}
+
+		if (property instanceof DefaultGeometryProperty<?>) {
```
stempler, line 242: The `instanceof` check and the cast later should be done with the interface `GeometryProperty` and not the implementation class.

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/writer/AbstractTableInstanceWriter.java
```diff
@@ -225,20 +234,40 @@ private boolean shouldBeDisplayed(Object obj) {
...
+			// correct winding order as per right-hand rule, i.e.,
+			// exterior rings are counterclockwise, and holes are
+			// clockwise.
+			setWindingOrder(EnumWindingOrderTypes.counterClockwise);
+			property = unifyGeometry(geometry, reporter,
+					geometryProperty.getCRSDefinition().getCRS());
+
+			if (geometry instanceof GeometryCollection
+					&& ((GeometryCollection) geometry).getNumGeometries() == 1) {
+				property = geometryProperty.getGeometry().getGeometryN(0);
```
stempler, line 257: Like this the CRS information gets lost.

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/writer/AbstractTableInstanceWriter.java
```diff
@@ -225,20 +234,40 @@ private boolean shouldBeDisplayed(Object obj) {
...
+			DefaultGeometryProperty<?> geometryProperty = (DefaultGeometryProperty<?>) property;
+
+			Geometry geometry = geometryProperty.getGeometry();
+
+			// correct winding order as per right-hand rule, i.e.,
+			// exterior rings are counterclockwise, and holes are
+			// clockwise.
+			setWindingOrder(EnumWindingOrderTypes.counterClockwise);
+			property = unifyGeometry(geometry, reporter,
+					geometryProperty.getCRSDefinition().getCRS());
```
stempler, line 253: I don't think for CSV/Excel there is a rule on what the winding order should be. Since you extended `AbstractGeoInstanceWriter` it would make sense to use its configuration parameters.
In that case

- you should use `extractGeometry` to detect and determine geometry and CRS, and apply conversion and the winding order
- extend the writer definitions (in the respective extension) with the additional parameters (target CRS, winding order - not sure about the EPSG prefix - is the CRS information included in the resulting file?)
- extend the UI to make the additional parameters configurable

stempler (changes requested):

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/writer/AbstractTableInstanceWriter.java
emanuelaepure10: @stempler please check the new commit. 
I believe that we should avoid using `extractGeometry` and `setting winding order`. Instead, when writing the property, we should simply check if the geometry is a multi-geometry. If it contains only one geometry, we should write that single geometry. Otherwise, we should follow the existing procedure.
Maybe the aproach that you are describing to create a UI to ask the user the winding order could be done as part of another ticker and extended to all the data export formats. 
In the [InstanceToJson](https://github.com/halestudio/hale/blob/247bc7ce6f33207783639bd178249cfd1413c455/io/plugins/eu.esdihumboldt.hale.io.json/src/eu/esdihumboldt/hale/io/json/internal/InstanceToJson.java#L652) I've see that the  `windingOrder = true` always.

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/writer/AbstractTableInstanceWriter.java
```diff
@@ -229,9 +236,48 @@ private void addProperty(List<String> headerRow, Map<String, Object> row, Object
...
+			if (multiLineString.getNumGeometries() == 1) {
+				return multiLineString.getGeometryN(0);
+			}
+		}
+		else if (geometry instanceof MultiPolygon) {
+			MultiPolygon multiPolygon = (MultiPolygon) geometry;
+			if (multiPolygon.getNumGeometries() == 1) {
+				return multiPolygon.getGeometryN(0);
+			}
+		}
```
stempler, line 277: I think the logic before with the `GeometryCollection` was fine, any specific reason to change it?

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/writer/AbstractTableInstanceWriter.java
```diff
@@ -229,9 +236,48 @@ private void addProperty(List<String> headerRow, Map<String, Object> row, Object
 		if (!headerRow.contains(propertyTypeName)) {
 			headerRow.add(propertyTypeName);
 		}
+
+		if (property instanceof GeometryProperty<?>) {
+			GeometryProperty<?> geometryProperty = (GeometryProperty<?>) property;
+
+			Geometry geometry = geometryProperty.getGeometry();
+			property = simplifyGeometry(geometry);
```
stempler, line 244: Like this the CRS information gets lost, as only the geometry is retained, not the CRS information.

emanuelaepure10: True. Then using the simplified geometry and the CRS, I have created a new DefaultGeometryProperty to be used for writing the files.

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/writer/AbstractTableInstanceWriter.java
stempler: Yes, you changed it. In the original version you did it the same way as in your changes to `InstanceToJson`, just checking for `GeometryCollection` and not these three subtypes. That's why I wondered if there is a specific reason to do it like that here.

stempler:
Is the second commit intended to be a fixup commit?

Then it should start with `fixup! ` followed by the message (first line) of the original commit.
You can also use `git commit --fixup <commit-sha>` to create such a commit.

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/writer/AbstractTableInstanceWriter.java
emanuelaepure10: @stempler the property that comes in the line 242 is a `GeometryProperty<?>` and not a `GeometryCollection`.

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/writer/AbstractTableInstanceWriter.java
emanuelaepure10: @stempler what it was done here was just after taking the geometry from the property, so after `(GeometryProperty<?>) property`. If you tell me that I need to check if the `property is an instanceof GeometryCollection` then that condition it will never be true. 
In the InstanceToJson is another story, already in the method comes a parameter which is a `GeometryProperty<?>`.

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/writer/AbstractTableInstanceWriter.java
emanuelaepure10: Please see the comments above related to your question.

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/writer/AbstractTableInstanceWriter.java
stempler: I'm afraid I don't understand the context of your comment.

> If you tell me that I need to check if the property is an instanceof GeometryCollection then that condition it will never be true

If you refer to my comment [here](https://github.com/halestudio/hale/pull/1173#discussion_r1613705904)

1. this comment was on another place in the code where the JTS geometry is handled, not the geometry property
2. I did not tell you anything, I just asked if there was a specific reason for the change to check for some `GeometryCollection` subclasses instead of the base class

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/writer/AbstractTableInstanceWriter.java
stempler: I don't quite understand. This comment was about using `GeometryProperty` instead of `DefaultGeometryProperty`, which you also did, so what is the open point here from your view?

io/plugins/eu.esdihumboldt.hale.io.csv/src/eu/esdihumboldt/hale/io/csv/writer/AbstractTableInstanceWriter.java
stempler: I can't really follow you. What proposal do you mean?

I see in the current commit you changed the logic back to checking `GeometryCollection` instead of the three subclasses that were checked in the code related to the original comment in this thread. The only thing I did was asking about if there was a specific reason for the change to check the three subclasses instead of only `GeometryCollection` as you did in your first implementation.

emanuelaepure10: No specific reason. In my opinion was a good approach but just using getGeometryN(0) it's enough for all the cases so I dropped the method.

io/plugins/eu.esdihumboldt.hale.io.json/src/eu/esdihumboldt/hale/io/json/internal/InstanceToJson.java
```diff
@@ -646,10 +647,16 @@ protected void writeGeometryValue(JsonGenerator jsonGen, GeometryProperty<?> geo
 			}
 		}
 
+		Geometry geomentry = geomProp.getGeometry();
```
stempler, line 650: Should this be `geometry`?

stempler (approved):
Please rebase locally to combine the commits (since the PR is from a fork, the `/autosquash` command does not work, also the commit message of the fixup commit seems to not reference the original commit correctly)

Issue #1084, opened 2023-11-28
[BUG]: If downloading data via WFS with parameter "resolvedepth" linked objects are not loaded 


annat2022:
**Describe the bug**
If I import source data via WFS with this request [https://geodienste.komm.one:443/ows/services/org.107.7e499bca-5e63-4595-b3c4-eaece8b68608_wfs/org.107.cf1dba4a-9b88-4284-a76f-a2594fe08905?SERVICE=WFS&VERSION=2.0.0&REQUEST=GetFeature&typenames=xplan:BP_Plan&resolvedepth=*](https://geodienste.komm.one:443/ows/services/org.107.7e499bca-5e63-4595-b3c4-eaece8b68608_wfs/org.107.cf1dba4a-9b88-4284-a76f-a2594fe08905?SERVICE=WFS&VERSION=2.0.0&REQUEST=GetFeature&typenames=xplan:BP_Plan&resolvedepth=*)
only one BP_Plan object is loaded but not the linked objects.

**To Reproduce**
1. Add source schema XPlan 5.0
2. Import source data via WFS with this link [https://geodienste.komm.one:443/ows/services/org.107.7e499bca-5e63-4595-b3c4-eaece8b68608_wfs/org.107.cf1dba4a-9b88-4284-a76f-a2594fe08905?SERVICE=WFS&VERSION=2.0.0&REQUEST=GetFeature&typenames=xplan:BP_Plan&resolvedepth=*](https://geodienste.komm.one:443/ows/services/org.107.7e499bca-5e63-4595-b3c4-eaece8b68608_wfs/org.107.cf1dba4a-9b88-4284-a76f-a2594fe08905?SERVICE=WFS&VERSION=2.0.0&REQUEST=GetFeature&typenames=xplan:BP_Plan&resolvedepth=*)
3. Hide unpopulated types -> you'll see that only one `BP_Plan` feature is loaded

**Expected behavior**
Downloading the data with the given GetFeature-Request I'd expect that one `BP_Plan` and one linked `BP_Bereich` and one linked `XP_Rasterdarstellung` is downloaded. 

**Additional context**
Using the same GetFeature-Request in the browser returns a gml-file with a `BP_Plan` objects and  `BP_Bereich`/`XP_Rasterdarstellung` wrapped in a `wfs:additionalObjects` type.
Importing this via a browser downloaded gml-file hale studio can read the feature types in the `additional objects` type correctly. 
[...]

PR #1180, opened 2024-05-21
fix: download linked objects in a WFS using "resolvedepth"

emanuelaepure10:
Using "resolvedepth" when trying to download data via WFS is now downloading the linked objects. As the requested hits to a WFS is better to be setup to UNKNOWN_SIZE, so the size comes as well to be always UNKNOWN_SIZE, the code related to hits and size has been removed and the related methods have been changed accordingly.

ING-4128
Closes [#1084](https://github.com/halestudio/hale/issues/1084)


emanuelaepure10:
@stempler Any idea why my PRs are always failing with the same error in the "Comment with links to the artifacts"?
Thank you

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/reader/internal/wfs/WfsBackedGmlInstanceCollection.java
```diff
@@ -556,9 +519,8 @@ public boolean hasNext() {
 		 *         number of results reported by the WFS.
 		 */
 		protected boolean isFeatureLimitReached() {
-			return (maxNumberOfFeatures != UNLIMITED
-					&& totalFeaturesProcessed >= maxNumberOfFeatures)
-					|| (size != UNKNOWN_SIZE && totalFeaturesProcessed >= size);
+			return maxNumberOfFeatures != UNLIMITED
+					&& totalFeaturesProcessed >= maxNumberOfFeatures;
```
stempler, line 523: Won't we still get problems here because the additional objects are counted as features processed?

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/reader/internal/wfs/WfsBackedGmlInstanceCollection.java
```diff
@@ -312,7 +279,18 @@ public WfsBackedGmlInstanceIterator iterator() {
 	 */
 	@Override
 	public boolean isEmpty() {
-		return size == 0;
+		if (!emptyInitialized) {
+			ResourceIterator<Instance> it = iterator();
```
stempler, line 283: I think a more effective way could be using the hits query that was previously used - it could even be used with `maxFeatures` / `count` set to `1` as we are only interested if there are matches at all.
With that we would avoid the server having to load and transfer a potentially large amount of data.

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/reader/internal/wfs/WfsBackedGmlInstanceCollection.java
emanuelaepure10: If we just setup the `hits == UNKNOWN_SIZE`, that's the code left. It didn't make sense to me to keep it as long as `size `would also be always `UNKNOWN_SIZE`. 
I have used just 2 projects to test, probably it's not enough, but I haven't encountered any problem.

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/reader/internal/wfs/WfsBackedGmlInstanceCollection.java
```diff
@@ -225,41 +227,6 @@ public WfsBackedGmlInstanceCollection(LocatableInputSupplier<? extends InputStre
...
-		case "2.0.2":
-			// The "numberMatched" reported by a 2.0.0/2.0.2 WFS should be
-			// number of features matched by the query. If hits equals
-			// UNKNOWN_SIZE then size is also set to that value
-			this.size = isLimited() ? Math.min(maxNumberOfFeatures, hits) : hits;
-			break;
-		default:
-			this.size = UNKNOWN_SIZE;
-		}
-
```
florianesser, line 262: I'm not sure why this must be removed. I think this can still be used in general (as before) when the query does not contain a `RESOLVEDEPTH` parameter, or am I missing something?

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/reader/internal/wfs/WfsBackedGmlInstanceCollection.java
florianesser: > If we just setup the hits == UNKNOWN_SIZE

> size would also be always UNKNOWN_SIZE

Can you explain why `size` should always be `UNKNOWN_SIZE`? In case the request does not contain a `RESOLVEDEPTH` parameter, the old logic can still be used IMO.

florianesser (changes requested):
Commit message and PR should reference the related GitHub issue #1084

emanuelaepure10:
Or if we just setup the hits to UNKNOWN_SIZE, then the ignoreNumberMatched box doesn't make any sense because wont be used.

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/reader/internal/wfs/WfsBackedGmlInstanceCollection.java
emanuelaepure10: From this code
```
                switch (wfsVersion) {
		case "1.1.0":
			// The "numberOfFeatures" reported by a 1.1.0 WFS may be smaller
			// than the actual number of features matches by the query if the
			// number of features returned per query is limited on the server
			// side. Therefore do not rely on it as a size information here.
			this.size = UNKNOWN_SIZE;
			break;
		case "2.0.0":
		case "2.0.2":
			// The "numberMatched" reported by a 2.0.0/2.0.2 WFS should be
			// number of features matched by the query. If hits equals
			// UNKNOWN_SIZE then size is also set to that value
			this.size = isLimited() ? Math.min(maxNumberOfFeatures, hits) : hits;
			break;
		default:
			this.size = UNKNOWN_SIZE;
		}
```
having hits = UNKNOWN_SIZE, also for the case 2.0.2 the size would come out as UNKNOWN_SIZE, independently if isLimited() is true or false.

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/reader/internal/wfs/WfsBackedGmlInstanceCollection.java
```diff
@@ -228,7 +228,15 @@ public WfsBackedGmlInstanceCollection(LocatableInputSupplier<? extends InputStre
...
-		if (ignoreNumberMatched) {
+		boolean isUsedresolvedepth = false;
+		for (NameValuePair nameValuePair : params) {
+			if ("resolvedepth".equalsIgnoreCase(nameValuePair.getName())) {
+				isUsedresolvedepth = true;
+				break;
+			}
+		}
+
+		if (ignoreNumberMatched || isUsedresolvedepth) {
```
florianesser, line 239: @emanuelaepure10 I think this can be reduced to a single line as `primordialQueryParams` contains all query parameters in upper case ([see here](https://github.com/halestudio/hale/blob/da47d95df419ab8fd1930ed4b56c12ab55ac0b8e/io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/reader/internal/wfs/WfsBackedGmlInstanceCollection.java#L198))

```suggestion
		if (ignoreNumberMatched || primordialQueryParams.contains("RESOLVEDEPTH")) {
```

emanuelaepure10: Thanks! I overlooked `primordialQueryParams`.

florianesser (approved):
LGTM

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/reader/internal/wfs/WfsBackedGmlInstanceCollection.java
```diff
@@ -228,7 +228,8 @@ public WfsBackedGmlInstanceCollection(LocatableInputSupplier<? extends InputStre
 		// Use primordial URI and issue "hits" request to check if the WFS will
 		// return anything at all
 		int hits;
-		if (ignoreNumberMatched) {
+
+		if (ignoreNumberMatched || primordialQueryParams.containsKey("RESOLVEDEPTH")) {
```
stempler, line 232: Maybe add a comment why this is done in case of `RESOLVEDEPTH` being included?

Did you test if the paging works as expected when additionalObjects are returned?
I could imagine a possible impact could be that the request for the second page uses a wrong offset.

emanuelaepure10: > Did you test if the paging works as expected when additionalObjects are returned?
> I could imagine a possible impact could be that the request for the second page uses a wrong offset.

For this I would probably need a bigger project I believe. I will ask the service team.

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/reader/internal/wfs/WfsBackedGmlInstanceCollection.java
```diff
@@ -225,10 +225,16 @@ public WfsBackedGmlInstanceCollection(LocatableInputSupplier<? extends InputStre
 			}
 		}
 
+		// for a query containing RESOLVEDEPTH we disable the pagination
+		if (primordialQueryParams.containsKey("RESOLVEDEPTH")) {
```
emanuelaepure10, line 229: After discussing the issue with @florianesser, we identified a potential solution: disable pagination if the parameters contain "RESOLVEDEPTH".

Another approach could be to disable pagination directly from the UI. I attempted this but haven't found a viable solution yet.

@stempler, your input and ideas would be greatly appreciated to help us find the best solution.

stempler: Problem with that would be that in many cases then it would not be possible to load all data of a service.

- Was your intent to inform the user about this restriction in the UI?
- Might be unexpected when used from another context, e.g. Gradle or CLI
- Did you explore if you would see some way to determine if an instance originated from `additionalObjects` or not? (e.g. see if it is feasible to detect [here](https://github.com/halestudio/hale/blob/d5e541ed65f91227e1c5cf1b10b89b7ebb777ba1/io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/reader/internal/GmlInstanceCollection.java#L169) once entering `additionalObjects` and adding the information to created instances in the [instance metadata](https://github.com/halestudio/hale/blob/d5e541ed65f91227e1c5cf1b10b89b7ebb777ba1/common/plugins/eu.esdihumboldt.hale.common.instance/src/eu/esdihumboldt/hale/common/instance/model/Instance.java#L74))

emanuelaepure10: > Was your intent to inform the user about this restriction in the UI?

Initially, I wanted to disable the checkbox, but I didn't find the way, but if you have any suggestion how to achieve that I would be very happy to hear. 
1. I have tried to understand in the `GmlReaderSettingsPage` if there is any way to get the link and like that be able to check if the "resolvedepth" parameter is there.
2. I tried to extend  `GmlReaderSettingsPage` into the WFS UI package but without success. 
3. Not the best solution would be to add a note in the actual UI and inform the user that that parameter for steps won't be used in case "resolvedepth" has been used. 

> Might be unexpected when used from another context, e.g. Gradle or CLI

True. 

> Did you explore if you would see some way to determine if an instance originated from additionalObjects or not?

I have tried but again without getting the info where the instance comes from. I will definitely try your suggestion.

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/reader/internal/wfs/WfsBackedGmlInstanceCollection.java
```diff
@@ -578,7 +597,31 @@ public Instance next() {
 			}
 
 			Instance instance = iterator.next();
-			return new StreamGmlInstance(instance, totalFeaturesProcessed++);
+
```
emanuelaepure10, line 600: @stempler Could you help me understand how is it possible that for a total or
`totalFeaturesProcessed: 754` and `additionalFeatureProcessed: 1844` processed here with a sum of 2598, 
HS shows the following numbers:
image
Thank you

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/reader/internal/wfs/WfsBackedGmlInstanceCollection.java
```diff
@@ -578,7 +597,31 @@ public Instance next() {
...
+			Iterable<QName> mapPropertiess = instance.getPropertyNames();
+			for (QName qName : mapPropertiess) {
+				if (qName.getLocalPart().equals("id") && qName.getPrefix().equals("gml")) {
```
stempler, line 603: The namespace prefix is nothing that should be checked, because that can be an arbitrary value.
[Here](https://github.com/halestudio/hale/blob/2baec50f5ba9117f09942c13725b031a449c17f0/io/plugins/eu.esdihumboldt.hale.io.wfs/src/eu/esdihumboldt/hale/io/wfs/file/WFSFeatureCollectionWriter.java#L93) is an example where we check for a GML namespace.

Since the GML namespace used should be the same in the same data, I would suggest to only determine the property name once and after that use a more efficient approach than iterating all property names by getting the property directly.

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/reader/internal/wfs/WfsBackedGmlInstanceCollection.java
```diff
@@ -578,7 +597,31 @@ public Instance next() {
...
+					Object[] gmlID = instance.getProperty(qName);
+					String gmlIDToCheck = (String) gmlID[0];
```
stempler, line 605: There should be a `null` check, also rather than risking a `ClassCastException` use `toString()`.

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/reader/internal/wfs/WfsBackedGmlInstanceCollection.java
stempler: I would suggest to add a unit test where a WFS FeatureCollection file with additional objects is read. That doesn't allow to test the pagination, but if you include at least one object with a duplicate GML ID you can use this to test:

1. The correct detection of the "additional objects"
2. If your mechanism for detecting GML ID duplicates works as expected

Once it has been ruled out that a problem is hidden there the paging behavior can be analysed.

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/reader/internal/wfs/WfsBackedGmlInstanceCollection.java
```diff
@@ -563,9 +574,17 @@ public boolean hasNext() {
 		 *         number of results reported by the WFS.
 		 */
 		protected boolean isFeatureLimitReached() {
+			boolean condition = (maxNumberOfFeatures != UNLIMITED
+					&& totalFeaturesProcessed >= maxNumberOfFeatures)
+					|| (size != UNKNOWN_SIZE && totalFeaturesProcessed >= size)
+					|| (totalFeaturesProcessed == size && !iterator.hasNext());
```
stempler, line 580: Why is there a change to the condition needed with this additional term?

emanuelaepure10: Because the additional objects found after the last main instance would be lost.

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/reader/internal/GmlInstanceCollection.java
```diff
@@ -205,6 +210,16 @@ private void proceedToNext() throws XMLStreamException {
 						}
 					}
 					typeStack.push(def);
+					elementStack.push(elementName.getLocalPart());
```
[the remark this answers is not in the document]
emanuelaepure10: I have tried your suggestion but all the features main and additional have the same namespaceURI and in this was I have no way to separate them.

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/reader/internal/wfs/WfsBackedGmlInstanceCollection.java
emanuelaepure10: There is something I don't understand. 
In the GmlInstanceCollection the [TypeDefinition def](https://github.com/halestudio/hale/blob/2baec50f5ba9117f09942c13725b031a449c17f0/io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/reader/internal/GmlInstanceCollection.java#L170) actually have a namespace URI starting with http://www.opengis.net/wfs,
but when arriving to the [WfsBakendGmlInstanceCollection](https://github.com/halestudio/hale/blob/2baec50f5ba9117f09942c13725b031a449c17f0/io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/reader/internal/wfs/WfsBackedGmlInstanceCollection.java#L573) the instance doesn't contain anymore a namespace URI starting with http://www.opengis.net/wfs, but just a QName; id, http://www.opengis.net/gml, gml, so cannot even be compared with a namespace such as http://www.opengis.net/wfs.

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/reader/internal/wfs/WfsBackedGmlInstanceCollection.java
stempler: Thank you. I think it would be very valuable to have that as a comment on that term.

emanuelaepure10:
@stempler does it happened to remember if there is somewhere a test example that contains "additionalObjects"? Thanks

io/plugins/eu.esdihumboldt.hale.io.gml.test/src/eu/esdihumboldt/hale/io/gml/reader/internal/StreamGmlReaderTest.groovy
```diff
@@ -163,4 +198,49 @@ class StreamGmlReaderTest {
...
+		def schemaLocation = getClass().getResource("/data/filter_doubles/filter.xsd").toURI();
+		def gmlLocation = getClass().getResource("/data/filter_doubles/filter.xml").toURI();
+		def paging = 1
+		def expected = 4
+
+		def sourceSchema = loadSchema(schemaLocation)
+		def instances = loadGmlFilter(gmlLocation, sourceSchema, null)
+
+		// Instantiate the class to test
+		DupplicateIDsFilterIterater dupplicateIDsFilterIterater = new DupplicateIDsFilterIterater(instances.iterator());
```
emanuelaepure10, line 213: @stempler could you please help me understand why I cannot import the package for DupplicateIDsFilterIterater? I have added the plugin in the MANIFEST.mf but somehow I cannot import the class.

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/reader/internal/wfs/WfsBackedGmlInstanceCollection.java
```diff
@@ -497,6 +491,9 @@ private void createNextIterator() {
 					sourceSchema, restrictToFeatures, ignoreRoot, strict, ignoreNamespaces,
 					crsProvider, ioProvider);
 			iterator = currentCollection.iterator();
+			if (primordialQueryParams.containsKey("RESOLVEDEPTH")) {
```
emanuelaepure10, line 494: @stempler Could you please help me understand how to initialize the DuplicateIDsFilterIterator that it's not initialize with every new request for pagination? Thank you

emanuelaepure10:
@stempler that should be hopefully the last and complete version. Please let me know if there is anything else to add. 
Thank you for your support

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/reader/internal/GmlInstanceCollection.java
```diff
@@ -568,13 +591,25 @@ public synchronized Instance next() {
...
-						crsProvider, nextType, null, false, ignoreNamespaces, ioProvider);
+				Instance instance = StreamGmlHelper.parseInstance(reader, nextType, elementIndex++,
+						strict, null, crsProvider, nextType, null, false, ignoreNamespaces,
+						ioProvider);
+
+				if (nextTypeDetails != null && !nextTypeDetails.isEmpty()) {
+					for (Map.Entry<TypeDefinition, String> entry : nextTypeDetails.entrySet()) {
+						TypeDefinition key = entry.getKey();
+						String val = entry.getValue();
+						((MutableInstance) instance).setMetaData(ADDITIONAL_OBJECTS, key);
```
stempler, line 602: Why store the type in there and not just a boolean?

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/reader/internal/GmlInstanceCollection.java
```diff
@@ -220,10 +239,14 @@ private void proceedToNext() throws XMLStreamException {
 
 					if (def != null && isAllowedType(def)) {
 						nextType = def;
+						if (isAnAdditionalObject) {
+							nextTypeDetails.put(def, ADDITIONAL_OBJECTS);
```
stempler, line 243: Why using a map?
I think a simple boolean would work as well (e.g. `nextIsAdditionalObject`).

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/reader/internal/wfs/DuplicateIDsFilterIterator.java
```diff
@@ -0,0 +1,159 @@
...
+			}
+		}
+		return true;
+	}
+
+	private boolean isGmlIdProperty(QName propertyName) {
+		return (propertyName.getNamespaceURI().startsWith(GMLConstants.NS_WFS)
+				|| propertyName.getNamespaceURI().startsWith(GMLConstants.GML_NAMESPACE_CORE))
+				&& "id".equals(propertyName.getLocalPart())
+				&& "gml".equals(propertyName.getPrefix());
```
stempler, line 156: I would leave out checking the prefix, since it is not guaranteed to be `gml`. Checking the namespace is enough.

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/reader/internal/wfs/DuplicateIDsFilterIterator.java
```diff
@@ -0,0 +1,159 @@
...
+			for (String string : uniqueIDInstances) {
+				System.out.println(string);
+			}
+			throw new NoSuchElementException();
+		}
+		instranceAlreadyReturned = true;
+		return nextInstance;
+	}
+
+	private boolean isTheInstancePresent(Instance instance) {
```
stempler, line 131: I think it would help to add documentation to the method to easily understand what it is supposed to do.

io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/reader/internal/wfs/DuplicateIDsFilterIterator.java
```diff
@@ -0,0 +1,159 @@
...
+
+/**
+ * Filter the instances by gml:id, storing only those with unique IDs.
+ */
+public class DuplicateIDsFilterIterator implements InstanceIterator {
+
+	private final HashSet<String> uniqueIDInstances = new HashSet<String>();
+	private final InstanceIterator iterator;
+	private Instance nextInstance;
+	private boolean instranceAlreadyReturned = false;
```
stempler, line 36: I think it would help documenting for the variable how it is used and what it represents (so basically that it tells us if `next()` or `skip()` have already been called for the current instance in `nextInstance`).

io/plugins/eu.esdihumboldt.hale.io.gml.test/src/eu/esdihumboldt/hale/io/gml/reader/internal/StreamGmlReaderTest.groovy
```diff
@@ -137,6 +137,41 @@ class StreamGmlReaderTest {
...
+				((InstanceIterator) it).skip()
+				count++
+				if (count % 100 == 0) {
+					println("$count instances skipped")
+				}
+			}
+		}
+
+		println("$count instances skipped")
+		assertEquals(expected, count)
```
stempler, line 172: Any idea why the test does not yield the expected result?

io/plugins/eu.esdihumboldt.hale.io.gml.geometry/src/eu/esdihumboldt/hale/io/gml/geometry/GMLConstants.java
```diff
@@ -38,4 +38,9 @@ public interface GMLConstants {
 	 */
 	public static final String NS_GML_32 = "http://www.opengis.net/gml/3.2";
 
+	/**
+	 * The GML 3.2 namespace
```
stempler, line 42: The constant name and its documentation should reflect that this is only part of a concrete WFS namespace (In other cases I think `prefix` or `core` are used in the name to symbolize this).

stempler (changes requested):

io/plugins/eu.esdihumboldt.hale.io.gml.test/src/eu/esdihumboldt/hale/io/gml/reader/internal/StreamGmlReaderTest.groovy
emanuelaepure10: Because was a wrong result. It should have been 754x3, so 2262.

emanuelaepure10:
It seems that the following three tests are failing:
```
StreamGmlWriterTest.testGeometryPrimitive_32_MultiPolygon_WindingOrder_CW
StreamGmlWriterTest.testGeometryPrimitive_32_MultiPolygon_WindingOrder 
StreamGmlWriterTest.testGeometryPrimitive_32_MultiPolygon_WindingOrder_CCW
```
However, these tests are passing successfully on my local environment.

Do you have any suggestions on what might be causing the failures and how to resolve them?

stempler (approved):

emanuelaepure10:
It appears that [3 tests are failing](https://github.com/halestudio/hale/actions/runs/10468140184). 
When I run the tests locally, all 16 tests in the StreamGmlWriterTest class pass successfully.

Could you provide some guidance on what might be causing these failures and how I can resolve them?

stempler:
> Could you provide some guidance on what might be causing these failures and how I can resolve them?

No clue at the moment what could be the issue. I retriggered the check but same result.
I [ported the change to hale-core](https://github.com/halestudio/hale-core/pull/17), I don't have the result on the CI yet, but locally these three tests also failed.

It seems to be a side effect of other tests running:

- if I only run one of the tests it works
- if I run only this test class it works
- if I run all tests from this project these three tests fail

emanuelaepure10:
test this please

stempler:
To rerun the GitHub actions job when there were failures you can just open the workflow run and select "Rerun jobs" -> "Rerun failed jobs".

io/plugins/eu.esdihumboldt.hale.io.gml.test/src/eu/esdihumboldt/hale/io/gml/reader/internal/GmlInstanceCollectionTest.java
```diff
@@ -229,6 +232,58 @@ public void testLoadShiporder() throws Exception {
...
+			throws IOException, IOProviderConfigurationException, URISyntaxException {
+		GmlInstanceCollection instances = loadInstances(
+				getClass().getResource("/data/filter_doubles/filter.xsd").toURI(),
+				getClass().getResource(pathToXML).toURI(), false);
+
+		FullInstanceIteratorSupport fullInstance = new FullInstanceIteratorSupport(
+				instances.iterator());
+
+		int uniqueInstanceCount = 0;
+		try (DuplicateIDsFilterIterator myIterator = new DuplicateIDsFilterIterator(fullInstance)) {
```
florianesser, line 274: @emanuelaepure10 Code style: Variable names `mySomething` should be avoided. IMO, a better choice here would be `duplicateIdIterator` or simply `it`.

emanuelaepure10:
@stempler @florianesser 
I temporarily removed the test in `StreamGmlReaderTest`', which was already commented as something to be moved when the services were no longer functional. This seems to have resolved the issue. However, I still don’t see how this is related to the failure of the other three tests in `StreamGmlWriterTest`.

io/plugins/eu.esdihumboldt.hale.io.gml.test/src/eu/esdihumboldt/hale/io/gml/reader/internal/GmlInstanceCollectionTest.java
```diff
@@ -229,6 +239,60 @@ public void testLoadShiporder() throws Exception {
...
+	 */
+	private int extractedUniqueInstances(String pathToXML)
+			throws IOException, IOProviderConfigurationException, URISyntaxException {
+
+		GmlInstanceCollection instances = loadInstances(
+				getClass().getResource("/data/filter_doubles/filter.xsd").toURI(),
+				getClass().getResource(pathToXML).toURI(), false);
+
+		FullInstanceIteratorSupport fullInstance = new FullInstanceIteratorSupport(
+				instances.iterator());
```
florianesser, line 279: @emanuelaepure10 Is `FullInstanceIteratorSupport` needed here at all? If yes, it would be great if you could add a brief explanation as a comment. I'm asking because in my experiments the test behaved the same without having `fullInstance` and instantiating `DuplicateIDsFilterIterator` directly with `instances.iterator()`. Thanks!

emanuelaepure10: No need indeed. I have used directly the `instances.iterator()`.

PR #1182, opened 2024-05-22
fix Maven artifact deployment

stempler:


stempler:
Successfully ran a test workflow with the changes [here](https://github.com/halestudio/hale/actions/runs/9188798814?pr=1182).

Issue #1104, opened 2024-02-07
[BUG]: Condition context filter created via "Occurring values" in Properties view is not valid


JohannaOtt:
**Describe the bug**
When creating a condition context on a type using an occurring value from the Properties and editing it, the filter is not valid.

**To Reproduce**
1. Load the dataset attached to hale studio (using the XPlanGML 5.1 schema from the presets)
[3Aenderung.zip](https://github.com/halestudio/hale/files/14192739/3Aenderung.zip)
2. Determine the occurring values of `RP_Landwirtschaft.id`
3. Right click on the value in the Occurring value list and chose "Create condition on parent RP_Landwirtschaft"
4. Right click on the new condition context and chose "Edit condition context"
5. The filter is invalid according to the dialogue 
6. When adding quotation marks around `id`, the filter is valid and still working (filter being `"id" = 'GML_517b396a-2685-11ec-962e-1f829c9bb818'`).

**Expected behavior**
I would expect the filter created by the functionality in the Occurring values list to be valid

**Screenshots**
![grafik](https://github.com/halestudio/hale/assets/34163838/ee58ee3f-13aa-4218-89b9-bc6f0554f4d1)

**Additional context**
The filter created for other attributes is valid. I assume that the original filter is invalid because of "id" being involved. According to the [documentation](http://help.halestudio.org/latest/topic/eu.esdihumboldt.hale.doc.user/html/reference/cql_filter.html?cp=0_7_1), it needs the quotation marks but those should then also be used when creating the filter from the Occurring values.

florianesser:
Can be challenged with [this build](https://github.com/halestudio/hale/pull/1184#issuecomment-2124607748).

JohannaOtt:
Tested with the workflow described above and the filter is not reported invalid in the dialogue any longer.
@florianesser If I understood the discussion in the related PR correctly, the issue will be closed with the PR and I therefore leave it open now?

florianesser:
Thanks for the challenge, @JohannaOtt!

> If I understood the discussion in the related PR correctly, the issue will be closed with the PR and I therefore leave it open now?

Yes, exactly. Now that you've challenged it, I will merge the PR which will close the issue.

PR #1184, opened 2024-05-22
fix: add quotes to automatically generated condition contexts on parent

florianesser:
When generating a condition context on a parent for a property with a name that has special meaning in JQL (like `id`), the generated filter does not work as expected. With this change, the property name is put in quotes so that filters with special property names work correctly.

https://github.com/halestudio/hale/issues/1104


emanuelaepure10 (approved):

stempler (changes requested):
I think for the link to the issue showing up properly in the changelog the reference needs to be prefixed with "closes" or "fixes".

e.g. `Closes #1104`

florianesser:
> I think for the link to the issue showing up properly in the changelog the reference needs to be prefixed with "closes" or "fixes".

@stempler Do you mean in the commit message or in the here in the PR description?

My reasoning for not adding a `closes` reference was to avoid that GitHub closes the issue automatically as soon as the PR is merged. If we're ok with that, I'm happy to add it.

stempler:
@florianesser Not 100% sure, I think it should be at least in the commit. We disussed that recently (see [here](https://wetransform.slack.com/archives/C9FUP0GPM/p1715153189644789?thread_ts=1715088123.417139&cid=C9FUP0GPM)). I think preferable would be if the PR is merged after the issue is challenged, which ideally can be done with the artifacts provided in the PR. What do you think?

emanuelaepure10:
From what I know the keywords that can be used are the following:

- Closes #issue_number
- Fixes #issue_number
- Resolves #issue_number

These can be added as well to our "Developer Guidelines".

florianesser:
Might be a nice enhancement to add a comment with the build artifacts directly to the referenced issue as well. Or was that not done on purpose?

stempler (approved):

stempler:
> From what I know the keywords that can be used are the following:

See [here](https://github.com/conventional-changelog/conventional-changelog/tree/master/packages/conventional-commits-parser#referenceactions) for what is supported by default by the release process.

florianesser:
> Do you mean when the PR is merged?

No, I meant when the PR artifacts are built so that the challenger can see them directly in the issue (I now [mentioned them manually there](https://github.com/halestudio/hale/issues/1104#issuecomment-2124745908)), but...

> For adding a comment to the issue we would also need some logic to determine the issue, not sure if that's easily done.

that explains it. As the PR is referenced anyway if one of the "Closes" keywords is used, I don't think it's necessary to put effort into an automatic comment.

florianesser:
Implementation was challenged by Johanna, will merge as soon as the checks are done.

PR #1187, opened 2024-05-27
feat: allow usage of LookupTableImpl and LookupTableInfoImpl in Groovy

florianesser:
Allow the usage of the classes `LookupTableImpl` and `LookupTableInfoImpl` to make it possible to use `LookupService` in Groovy scripts without having to lift the restrictions.

ING-4256


stempler (approved):
Does that conclude challenging https://github.com/halestudio/hale/pull/1166 ?

emanuelaepure10 (approved):
Thank you Florian for adding them. I just didn't know what else should be added.

florianesser:
> Does that conclude challenging https://github.com/halestudio/hale/pull/1166 ?

Yes, I would say so. I've asked @Kate-Lyndegaard in the ING ticket to re-challenge it with the build from this PR and the example project I attached to the ING ticket.

PR #1189, opened 2024-05-28
ci: add job to check PR label requirements

florianesser:
Add a check to verify that a PR has the `challenged` label. Would need to be added to the "required checks" in the repository. Functionality can be tested with [this PR in my fork](https://github.com/florianesser/hale/pull/8).


.github/workflows/labels.yml
```diff
@@ -0,0 +1,19 @@
+name: Check labels
+
+on:
+  pull_request:
+    branches:
+      - '*'  # Trigger on all branches for pull requests
```
stempler, line 6: Should this be `master`?
I guess in practice it does not make a difference.

.github/workflows/labels.yml
```diff
@@ -0,0 +1,19 @@
...
+jobs:
+  require-challenge:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: mheap/github-action-required-labels@5847eef68201219cf0a4643ea7be61e77837bbce # v5.4.1
+        with:
+          mode: exactly
+          count: 1
+          labels: "challenged"
+          add_comment: true
```
stempler, line 18: The GitHub Actions token won't have write access if the PR is from a fork, do you know if the action handles that gracefully?

stempler (commented):
> Add a check to verify that a PR has the challenged label.

Please add `require-challenge` to the required checks in the `.wetf-repo.yml` file.

.github/workflows/labels.yml
florianesser: I added a `permissions` section now but it looks as if the workflow still doesn't have the necessary permissions to add a comment to the PR. Looks like the same problem that [you had here](https://github.com/halestudio/hale/commit/3149a0217d45aaa85aa34a051bb4b9c5caa2aa01). Any suggestions welcome.

stempler: It's just like that. PR from forks don't get write permissions. You can configure it to allow it, but only for private repos. For public repositories there would only be the way to pass in a separate token that has the required permission (but here you have the same problem because fork PRs don't have access to secrets). So it might be easiest to just not create a comment.

The error you see right now on this PR though probably stems from trying to add a new workflow from a fork PR, which is also not allowed (I think it would usually run the version it finds in the base branch, but that does not exist). So to properly run on this PR, you would need to create the PR from a fork.

florianesser:
Closed in favour of #1190

PR #1197, opened 2024-06-12
fix: Reload and update schemas - saves the last changes

### LINKED ISSUE #956: reload and update schemas - project save function is not working
Hale studio 4.1.0. 

When I run 'File/Reload and update schemas' there is offer to 'Save changes to current project?'.
But id doesn't save changes after hit 'Yes'.

see also https://discuss.wetransform.to/#/thread/3fel/Save-project-after-reload-and-update-schemas

emanuelaepure10:
By using the "Reload and update schemas" button the saving function it's working saving all the changes done to the project.

ING-4307
Closes #956

### DIFF
--- a/ui/plugins/eu.esdihumboldt.hale.ui/src/eu/esdihumboldt/hale/ui/service/project/internal/ProjectServiceImpl.java
+++ b/ui/plugins/eu.esdihumboldt.hale.ui/src/eu/esdihumboldt/hale/ui/service/project/internal/ProjectServiceImpl.java
@@ -651,8 +651,9 @@ public class ProjectServiceImpl extends AbstractProjectService implements Projec
 				}
 
 				String projectName = getProjectInfo().getName();
-				String title = appTitle + " - " + ((projectName == null || projectName.isEmpty())
-						? "Unnamed" : projectName);
+				String title = appTitle + " - "
+						+ ((projectName == null || projectName.isEmpty()) ? "Unnamed"
+								: projectName);
 				if (projectFile == null) {
 					// TODO Use scheme to discover plugin that can provide title
 					// information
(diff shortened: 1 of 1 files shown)


ui/plugins/eu.esdihumboldt.hale.ui/src/eu/esdihumboldt/hale/ui/service/project/internal/ProjectServiceImpl.java
```diff
@@ -884,8 +885,8 @@ public void run() {
 
 	@Override
 	public void update() {
-		// no change check as this is done by clean before a new project is
-		// loaded
+		// check if there are changes, as you are about to load the same project
```
stempler, line 888: Related to the original comment, how is the behavior in this case?

1. open a project
2. modify it
3. new alignment project

emanuelaepure10: You will get this question
image
and here depends by the user and his answer if the changes will be saved or not.

stempler: Also a question related to the issue (#956) - the description says the possibility to save the project was offered, but the changes were not saved. Is that also the behavior you observed?
The main task of the `changeCheck` method seems to be to show that dialog, that was also shown before according to the issue description. So does your change solve this by showing it at an earlier time and the later point where it was already checked does not result in the dialog being shown?

emanuelaepure10: > Also a question related to the issue (#956) - the description says the possibility to save the project was offered, but the changes were not saved. Is that also the behavior you observed?

I do observe the same behaviour. 

> The main task of the changeCheck method seems to be to show that dialog, that was also shown before according to the issue description. So does your change solve this by showing it at an earlier time and the later point where it was already checked does not result in the dialog being shown?

It is exactly how you say. The solution solves the problem showing the dialog earlier. The dialog is not showed anymore later because the changes has ben saved so when checking if there is any change the clean method exit immediately without showing again the dialog.

stempler (approved):

PR #1214, opened 2024-06-26
fix: usage of a broken or risky cryptographic algorithm

emanuelaepure10:
Replace MD5 with SH-256 because MD5 is susceptible to several vulnerabilities, including collision attacks. By switching from MD5 to SHA-256, we enhance the security of your hashing process, making it suitable for modern cryptographic applications.

ING-4368

### DIFF
--- a/ui/plugins/eu.esdihumboldt.hale.ui/src/eu/esdihumboldt/hale/ui/service/groovy/internal/PreferencesGroovyService.java
+++ b/ui/plugins/eu.esdihumboldt.hale.ui/src/eu/esdihumboldt/hale/ui/service/groovy/internal/PreferencesGroovyService.java
@@ -15,8 +15,8 @@
 
 package eu.esdihumboldt.hale.ui.service.groovy.internal;
 
-import java.io.UnsupportedEncodingException;
 import java.net.URI;
+import java.nio.charset.StandardCharsets;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.util.ArrayList;
@@ -317,9 +317,9 @@ public class PreferencesGroovyService extends DefaultGroovyService {
 			// not simply using hashCode, because it would be far to easy to
 			// modify the script in a undetectable way
 			try {
-				MessageDigest md = MessageDigest.getInstance("MD5");
+				MessageDigest md = MessageDigest.getInstance("SHA-256");
 				for (String script : scripts)
-					md.update(script.getBytes("UTF-8"));
+					md.update(script.getBytes(StandardCharsets.UTF_8));
 				byte[] hash = md.digest();
 				StringBuilder sb = new StringBuilder(2 * hash.length);
 				for (byte b : hash) {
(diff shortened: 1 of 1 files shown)


florianesser (commented):
A consequence of this change would be that all alignments that use a hash code for generating IDs will in future generate new IDs which is most likely unexpected. Was there already a discussion on this (I don't see one on the internal ING ticket)?

> Replace MD5 with SH-256 because MD5 is susceptible to several vulnerabilities, including collision attacks.

While it's true that MD5 is considered broken for cryptographic purposes, the use case in hale studio is non-cryptographic, so I would see no risk in continuing to use MD5 for generating hashes.

emanuelaepure10:
From what I understand, MD5 is considered a weak and outdated cryptographic algorithm due to its susceptibility to collision attacks. For most purposes, it is recommended to use more secure algorithms like SHA-256.

However, there hasn't been an internal discussion on this matter. If we choose to continue using MD5, we must accept its vulnerabilities. As long as we avoid using it for hashing passwords or signing data, it should not pose a significant issue.

Given this context, it seems appropriate to close the PR. What are your thoughts?

Or, another idea, we could keep MD5 for the projects which has been saved using MD5 but use SHA-256 for the future.

stempler:
@emanuelaepure10 Maybe you can check where the script hash is used.

By looking at the code via GitHub the only thing that I identified was the functionality in the UI related to asking for permission if Groovy scripts run. If that is the only place then the worst thing that would happen here is that users are again asked for permission when opening a project in a new version. I would not consider that a problem and and in that case would see no argument not to change to the more secure algorithm.

florianesser:
> A consequence of this change would be that all alignments that use a hash code for generating IDs will in future generate new IDs which is most likely unexpected.

@emanuelaepure10 My initial evaluation was not correct. The function that is used in Groovy scripts for generating IDs is `_.crypto.sha256` which is (obviously) unrelated to the change in this PR. I made my original comment believing that the MD5 hash function was the one used in Groovy scripts.

Sorry for the fuss! No objections from my side to merge this.

emanuelaepure10:
Could one of you @stempler @florianesser  please review and accept the PR so I can move the issue ahead in the workflow?
Thank you very much

florianesser:
> Maybe you can check where the script hash is used.

> By looking at the code via GitHub the only thing that I identified was the functionality in the UI related to asking for permission if Groovy scripts run.

@emanuelaepure10 Can you confirm that the functionality is only used for this?

florianesser:
@emanuelaepure10 Ok thanks, last things from my side would be:
- formatting the commit message to not have more than 100 characters in a line
- reword the commit message subject slightly (see below)

```
fix: replace broken or risky cryptographic algorithm

Replace MD5 with SHA-256 because MD5 is susceptible to several vulnerabilities, including
collision attacks. By switching from MD5 to SHA-256, we enhance the security of your hashing
process, making it suitable for modern cryptographic applications.

ING-4368
```

PR #1215, opened 2024-06-26
fix: arbitrary file access during archive extraction ("Zip Slip")

emanuelaepure10:
Ensure that output paths constructed from zip archive entries are validated to prevent writing files to unexpected locations.

ING-4369


stempler (approved):

florianesser:
@emanuelaepure10 One general comment on commit messages: in the commit message subject (= the first line), the part after the commit type - in this case `arbitrary file access during archive extraction ("Zip Slip")` - should start with a verb in imperative mood, e.g. `replace`, `change`, `fix` or `avoid`. So a better message in this case would be:

```
fix: avoid arbitrary file access during archive extraction ("Zip Slip")
```

PR #1216, opened 2024-06-26
fix: added scroll functionality to the CheckboxTableViewer for select…

emanuelaepure10:
…ing export types for CSV and Excel

Implemented scroll functionality in the CheckboxTableViewer used to select export types for CSV and Excel. This enhancement ensures that users can navigate through a potentially long list of export types comfortably, even within constrained dialog sizes.

ING-4333


io/plugins/eu.esdihumboldt.hale.io.csv.ui/src/eu/esdihumboldt/hale/io/csv/ui/InstanceExportConfigurationPage.java
```diff
@@ -122,23 +119,17 @@ public void widgetSelected(SelectionEvent e) {
...
-			Color greyLabel = PlatformUI.getWorkbench().getDisplay()
-					.getSystemColor(SWT.COLOR_DARK_GRAY);
-			separatorLabel.setForeground(greyLabel);
+			featureTypeTable = new CheckboxTableViewer(table);
 
-			table = new Table(chooseFeatureTypes, SWT.CHECK | SWT.MULTI | SWT.SCROLL_PAGE);
-			table.setHeaderVisible(false);
-			table.setLinesVisible(false);
-			table.setBackground(PlatformUI.getWorkbench().getDisplay()
-					.getSystemColor(SWT.COLOR_LIST_BACKGROUND));
```
stempler, line 137: Any specific reason to remove these other settings?

io/plugins/eu.esdihumboldt.hale.io.csv.ui/src/eu/esdihumboldt/hale/io/csv/ui/InstanceExportConfigurationPage.java
```diff
@@ -122,23 +119,17 @@ public void widgetSelected(SelectionEvent e) {
...
-			GridDataFactory groupData = GridDataFactory.fillDefaults().grab(true, false);
-			groupData.applyTo(table);
+			// Configure the layout for the table
+			GridData tableGridData = new GridData(SWT.FILL, SWT.FILL, true, true);
+			tableGridData.heightHint = 300; // Minimum height for the table
+			tableGridData.widthHint = 300; // Minimum width for the table
+			table.setLayoutData(tableGridData);
```
stempler, line 131: Fine for me like that, but I think using `GridDataFactory` would be generally preferable for applying the settings.

io/plugins/eu.esdihumboldt.hale.io.csv.ui/src/eu/esdihumboldt/hale/io/csv/ui/InstanceExportConfigurationPage.java
emanuelaepure10: These lines were removed because they were either redundant or unnecessary given the context:
1. Header visibility: `table.setHeaderVisible(false);`
By default, the table header is not visible unless explicitly set to true. Since the intention was to keep the header invisible, this line was redundant.
2. Lines visibility: `table.setLinesVisible(false);`
Similar to the header visibility, the default setting for table lines visibility is false unless set to true. This line was therefore redundant as well.
3. Background color: `table.setBackground(PlatformUI.getWorkbench().getDisplay().getSystemColor(SWT.COLOR_LIST_BACKGROUND));`
The background color setting was removed because the default background color is typically suitable for the UI design, and it wasn't necessary to explicitly set it in this case.

io/plugins/eu.esdihumboldt.hale.io.csv.ui/src/eu/esdihumboldt/hale/io/csv/ui/InstanceExportConfigurationPage.java
emanuelaepure10: Thank you for the suggestion. I have updated the code accordingly.

Issue #890, opened 2021-03-23
Validation of XPlan writer erroneously throws warnings on missing objects


JohannaOtt:
Version 4.1.0.SNAPSHOT
Build 763247191
Revision df620a7f13f58d8b80d5f5b3f488e3c3a3229e35

When transforming several objects to XPlan and exporting them with the Xplan writer using the functionality to create separate output files for every plan element, the writer throws many warnings all following the same pattern
![grafik](https://user-images.githubusercontent.com/34163838/112179455-001ec200-8bfb-11eb-899d-70fc803c7004.png)
The warning is about objects that are referenced, but are missing. This is not true for the data - the gml files are containing both objects that are referencing each other.
The warning occurs multiple times for each of the exported plans.

Kate-Lyndegaard:
@JohannaOtt Do you have the example file(s) used in the issue description?

JohannaOtt:
I can currently not re-produce the issue. Currently waiting for confirmation if it can be closed.

JohannaOtt:
It is re-producible, checked the wrong logs.
@Kate-Lyndegaard I can provide you with the respective alignment and files.

PR #1218, opened 2024-06-28
fix: Xplan validation erroneously throws warning on missing objects

emanuelaepure10:
Closes #4318

### DIFF
--- a/common/plugins/eu.esdihumboldt.hale.common.instance.graph/src/eu/esdihumboldt/hale/common/instance/graph/reference/ReferenceGraph.java
+++ b/common/plugins/eu.esdihumboldt.hale.common.instance.graph/src/eu/esdihumboldt/hale/common/instance/graph/reference/ReferenceGraph.java
@@ -27,7 +27,6 @@ import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.Queue;
 import java.util.Set;
-import java.util.stream.Collectors;
 
 import javax.annotation.Nullable;
 
@@ -136,7 +135,7 @@ public class ReferenceGraph<T> {
 				List<InstanceReference> instances = getNextAtomicPart();
 				biggestAtom = Math.max(biggestAtom, instances.size());
 
-				if (part.size() + instances.size() > maxObjects) {
+				if (part.size() > 0 && part.size() + instances.size() > maxObjects) {
 					// add to part candidates for later use
 					nextCandidates.add(instances);
 					if (!verticesLeft()) {
@@ -199,7 +198,14 @@ public class ReferenceGraph<T> {
 		 */
 		private List<InstanceReference> getNextAtomicPart() {
 			// select an arbitrary vertex
-			Vertex vtx = graph.someVertex();
+			Vertex vtx = null;
+
+			if (handleFirst != null) {
+				vtx = graph.getVertex(handleFirst);
+			}
+			if (vtx == null) {
+				vtx = graph.someVertex();
+			}
 			if (vtx != null) {
 				// get all vertices associated with that vertex
 				final Set<Vertex> visited = new LinkedHashSet<>();
@@ -236,22 +242,20 @@ public class ReferenceGraph<T> {
 						result.add(ref);
 					}
 					else {
-						Iterable<Vertex> referers = associated.getVertices(Direction.IN);
 						Set<String> ids = new HashSet<>();
-						if (referers != null) {
-							for (Vertex referer : referers) {
-								Object ident = referer.getId();
-								if (ident != null) {
-									ids.add(ident.toString());
-								}
+						for (Vertex referer : associated.getVertices(Direction.IN)) {
+							Object ident = referer.getId();
+							if (ident != null) {
+								ids.add(ident.toString());
 							}
 						}
+
 						if (ids.isEmpty()) {
 							log.warn("Encountered referenced object w/o associated instance: "
 									+ associated.getId());
 						}
 						else {
-							String enumIds = ids.stream().collect(Collectors.joining(", "));
+							String enumIds = String.join(", ", ids);
 							log.warn("Encountered referenced object w/o associated instance: "
 									+ associated.getId() + " - referenced from " + enumIds);
 						}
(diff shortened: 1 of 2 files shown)


common/plugins/eu.esdihumboldt.hale.common.instance.graph/src/eu/esdihumboldt/hale/common/instance/graph/reference/ReferenceGraph.java
```diff
@@ -235,24 +241,24 @@ public Boolean compute(LoopBundle<Vertex> loop) {
 					if (ref != null) {
 						result.add(ref);
 					}
-					else {
+					else if (!vtx.getId().equals(associated.getId())) {
```
stempler, line 244: So there were references from one instance to itself?

common/plugins/eu.esdihumboldt.hale.common.instance.graph/src/eu/esdihumboldt/hale/common/instance/graph/reference/ReferenceGraph.java
```diff
@@ -235,24 +241,24 @@ public Boolean compute(LoopBundle<Vertex> loop) {
...
 						Iterable<Vertex> referers = associated.getVertices(Direction.IN);
 						Set<String> ids = new HashSet<>();
-						if (referers != null) {
-							for (Vertex referer : referers) {
-								Object ident = referer.getId();
-								if (ident != null) {
-									ids.add(ident.toString());
-								}
+
```
stempler, line 247: Any particular reason to remove the null check?

common/plugins/eu.esdihumboldt.hale.common.instance.graph/src/eu/esdihumboldt/hale/common/instance/graph/reference/ReferenceGraph.java
```diff
@@ -235,24 +241,24 @@ public Boolean compute(LoopBundle<Vertex> loop) {
...
+						for (Vertex referer : referers) {
+							Object ident = referer.getId();
+							if (ident != null) {
+								ids.add(ident.toString());
 							}
 						}
+
 						if (ids.isEmpty()) {
-							log.warn("Encountered referenced object w/o associated instance: "
+							log.warn("Encountered referenced object without associated instance: "
```
stempler, line 256: Any particular reason to change the text?

common/plugins/eu.esdihumboldt.hale.common.instance.graph/src/eu/esdihumboldt/hale/common/instance/graph/reference/ReferenceGraph.java
```diff
@@ -294,23 +300,36 @@ public void remove() {
...
 	private boolean partitioned = false;
 
+	private final T handleFirst;
+
 	/**
 	 * Create a new reference graph from the given instance collection.
 	 * 
 	 * @param inspector the instance inspector to use
 	 * @param instances the
+	 * @param handleFirst first to be handle
```
stempler, line 310: Maybe more detailed?
It should be clear that what is passed in here is an ID, and that it is optional and null may be passed in.

common/plugins/eu.esdihumboldt.hale.common.instance.graph/src/eu/esdihumboldt/hale/common/instance/graph/reference/ReferenceGraph.java
emanuelaepure10: Yes and that was the case when actually the warning was given.

common/plugins/eu.esdihumboldt.hale.common.instance.graph/src/eu/esdihumboldt/hale/common/instance/graph/reference/ReferenceGraph.java
emanuelaepure10: I haven't, it's just below.

stempler: No, `referers` is no longer checked for `null`.
Did you check the API and it is guaranteed to never be `null`?

common/plugins/eu.esdihumboldt.hale.common.instance.graph/src/eu/esdihumboldt/hale/common/instance/graph/reference/ReferenceGraph.java
stempler: Did you identify how that can happen? I don't remember self-references from the test data that we looked at and even if there were I would expect the `ref != null` condition to be `true`.
Any idea? Just wondering if something is not correct when building the graph.

common/plugins/eu.esdihumboldt.hale.common.instance.graph/src/eu/esdihumboldt/hale/common/instance/graph/reference/ReferenceGraph.java
stempler: Just curious

common/plugins/eu.esdihumboldt.hale.common.instance.graph/src/eu/esdihumboldt/hale/common/instance/graph/reference/ReferenceGraph.java
emanuelaepure10: The `associated.getVertices(Direction.IN)` will return an empty iterable if there are no vertices.
I actually wanted to used this 
```
Set<String> ids = new HashSet<>();
for (Vertex referer : associated.getVertices(Direction.IN)) {
    Object ident = referer.getId();
    if (ident != null) {
        ids.add(ident.toString());
    }
}
```

common/plugins/eu.esdihumboldt.hale.common.instance.graph/src/eu/esdihumboldt/hale/common/instance/graph/reference/ReferenceGraph.java
emanuelaepure10: Yes, the condition `vtx.getId().equals(associated.getId())` is `true` for the example attached to the issue many times, as well as for the sample example that we have created!

stempler (approved):

stempler:
Related ticket is #890

stempler:
@emanuelaepure10 I proposed a change for you to review in a fixup commit.

We had done a change together to try to have the first atomic part in the partitioning for each plan to include the plan. Only the change to first check the plan vertex in the reference graph was not enough, because that first atomic part was then skipped for later processing, since its size exceeded the maximum size desired for a part.
I modified the behavior to not skip the atomic part if the part that is currently prepared is empty. Like that we make sure that the first part retrieved actually contains the plan, and no other parts are being looked at (and thus we don't get any warnings).

emanuelaepure10 (commented):
LGTM

PR #1219, opened 2024-06-28
fix: use correct representation of date and time when exporting XML/GML

emanuelaepure10:
Resolves issue introduced in 5.2.0 release.
Conversion to XmlDate and XmlDateTime was done but the result not used.

### DIFF
--- a/io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/internal/simpletype/SimpleTypeUtil.java
+++ b/io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/internal/simpletype/SimpleTypeUtil.java
@@ -108,6 +108,7 @@ public class SimpleTypeUtil {
 					GDate gdate = builder.toGDate();
 
 					xmlDate.setGDateValue(gdate);
+					return xmlDate.getStringValue();
 				}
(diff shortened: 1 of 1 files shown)


stempler (commented):
Would be great to have a bit more meaningful message for the changelog.

stempler:
@emanuelaepure10 Thanks a lot for updating the commit message. I think that's already a lot better, but I think we can still improve it in a way that it is more meaningful to a user, e.g. make it clear what functionality is impacted.
Technical details can be mentioned in the message body.

What do you think about this for example?

```
fix: use correct representation of date and time when exporting XML/GML

Resolves issue introduced in 5.2.0 release.
Conversion to XmlDate and XmlDateTime was done but the result not used.
```

emanuelaepure10:
@stempler I have updated the comment, but I don't understand why do you think that this bug was introduced in the 5.2.0?
Are you referring to [this ](https://github.com/halestudio/hale/commit/2baec50f5ba9117f09942c13725b031a449c17f0)commit? 
In the [SimpleTypeUtil](https://github.com/halestudio/hale/blame/master/io/plugins/eu.esdihumboldt.hale.io.gml/src/eu/esdihumboldt/hale/io/gml/internal/simpletype/SimpleTypeUtil.java) the return after the conversion of XMLDate and XMLDateTime was never there.

stempler (approved):

PR #1247, opened 2024-09-05
feat(shp): support writing date fields with other bindings than Date

stempler:
- LocalDate
- LocalDateTime
- Instant

Please note that the time portion gets lost because Geotools by default uses the date only type in DBF.

Ported over from https://github.com/halestudio/hale-core/pull/69


stempler:
FYI @annat2022 Ported this change over to hale studio so there is consistent behavior in hale studio and other products.

PR #1277, opened 2024-11-11
fix: wait until adding instance to sink is complete

stempler:
...to avoid the sink being closed while asynchronous operations are still pending.

ING-4479

Fix ported from https://github.com/halestudio/hale-core/pull/116


florianesser (approved):
LGTM

kapil-agnihotri (approved):

PR #1285, opened 2024-12-06
build: ensure that MSI installer file name starts with hale-studio

florianesser:
Until now, file names of the MSI installer hale»studio followed the pattern

    HALE-<version>-<language>-<arch>.msi

Change this to

    hale-studio-<version>-<language>-<arch>.msi

but do not change the file names of MSI installers for other products.


stempler (approved):
LGTM

PR #1307, opened 2025-01-31
ci: use AWS CLI v2

florianesser:
Use AWS CLI v2 as v1 does not seem to be supported anymore on Ubuntu 24.04 (see also [this issue comment](https://github.com/unfor19/install-aws-cli-action/issues/31#issuecomment-2408688017).

@stempler Do you remember why v1 is used here explicitly? And is there a good way of testing a change like this without merging it to `master` first?


stempler (approved):
No obvious breaking changes documented that would be a problem, giving it a try.

PR #1309, opened 2025-02-03
ci: exclude infocenter from published linux artifact

stempler:
This is necessary because the name of the product was changed in 17d36207583dad9c48b305ebab48f3aad6159db1


florianesser (approved):
LGTM

PR #1510, opened 2026-06-22
fix(geopackage): read boolean fields as Boolean instead of Integer (#1223)

### LINKED ISSUE #1223: [BUG]: For Geopackage containing a boolean field, the value 'true' is read as '1'
**Describe the bug**

**To Reproduce**

**Expected behavior**
A clear and concise description of what you expect to happen.

**Screenshots**

**Additional context**

adamnadolny-wizipisi:
## Summary

Fixes #1223

GeoPackage stores `BOOLEAN` columns as `INTEGER` (0/1) in SQLite. When reading
data back, JDBC `ResultSet.getObject()` returns a `java.lang.Integer`, while the
schema (built by `GeopackageSchemaBuilder`) correctly declares `Boolean.class` as
the binding for such columns. As a result, boolean values were surfaced as `1`/`0`
instead of `true`/`false`.

**Fix:** In `TableInstanceBuilder.groovy`, after reading the raw JDBC value, convert
`Number -> Boolean` when the target property binding is `Boolean.class`.

**Affected file:** `eu.esdihumboldt.hale.io.geopackage/internal/TableInstanceBuilder.groovy`

## Test

Added `testWriteReadBoolean()` to `GeopackageInstanceWriterTest.groovy`:
- Writes two records with `Boolean.TRUE` and `Boolean.FALSE` values
- Verifies the schema binding is `Boolean.class` (not `Integer`)
- Verifies the read-back values are `Boolean` instances with correct `true`/`false` values

---

Author: The Wroclaw Institute of Spatial Information and Artificial Intelligence
(WIZIPISI - Wrocławski Instytut Zastosowań Informacji Przestrzennej i Sztucznej Inteligencji)

### DIFF
--- a/io/plugins/eu.esdihumboldt.hale.io.geopackage.test/src/eu/esdihumboldt/hale/io/geopackage/GeopackageInstanceWriterTest.groovy
+++ b/io/plugins/eu.esdihumboldt.hale.io.geopackage.test/src/eu/esdihumboldt/hale/io/geopackage/GeopackageInstanceWriterTest.groovy
@@ -597,6 +597,65 @@ class GeopackageInstanceWriterTest {
 		}
 	}
 
+	/**
+	 * Test that boolean field values are read back as Boolean, not as Integer.
+	 * Regression test for https://github.com/halestudio/hale/issues/1223
+	 */
+	@Test
+	void testWriteReadBoolean() {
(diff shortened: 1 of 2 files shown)


stempler (approved):
Thanks a lot for the contribution. Looks good to me!

PR #1511, opened 2026-06-22
Fix GeoPackage writer silently appending data when target file is in use

### LINKED ISSUE #1201: [BUG]: geopackage writer overwrite functionality is not working when file is in use
**Describe the bug**
When exporting geopackage from hale»studio to an existing file with the "Overwrite target file if it exists" option enabled, the data is not overwritten but added if the file is in use (for example in QGIS)

**To Reproduce**
1. Export a geopackage from hale»studio
2. Load the geopackage for example in QGIS
3. Export the geopackage again from hale»studio to the same file with the "Overwrite target file if it exists" option enabled
4. Load the file again in QGIS. Every object will be contained twice.

**Expected behavior**
The export functionality should throw an error if it cannot overwrite the existing file instead of adding the data to the file again.

**Additional context**
If the file is not in use, the overwrite functionality is working as expected.
In [this part of the code](https://github.com/florianesser/hale/blob/693b8b8a38782f9686af3a310428515edc3674a6/io/plugins/eu.esdihumboldt.hale.io.geopackage/src/eu/esdihumboldt/hale/io/geopackage/GeopackageInstanceWriter.java#L202), it should be checked if the file can be overwritten, or not.
If not, the export function should throw an error.

adamnadolny-wizipisi:
## Summary

Fixes #1201

When the **Overwrite target file if it exists** option is enabled but the target .gpkg file cannot be deleted (e.g. it is locked by QGIS or another process on Windows), File.delete() returns False silently. The writer then opens the existing file and **appends** data instead of overwriting it — resulting in duplicated records with no error reported to the user.

### Root cause

In GeopackageInstanceWriter.java (around line 202), the return value of File.delete() was not checked:

```java
// before fix
file.delete();  // silently returns false if file is locked
```

### Fix

Check the return value and throw IOException if deletion fails:

```java
if (!file.delete()) {
    throw new IOException("Cannot overwrite existing GeoPackage file (file may be in use): "
            + file.getAbsolutePath());
}
```

The exception is caught by the surrounding 	ry/catch in execute(), which marks the report as failed and records the error — so the user sees a clear error message instead of silently corrupted output.

### Test

Added 	estOverwriteFailsWhenFileInUse() to GeopackageInstanceWriterTest:
- Writes a GeoPackage file
- Simulates "file in use" by holding an open RandomAccessFile handle (mirrors Windows file locking behaviour, e.g. when QGIS has the file open)
- Asserts that the writer reports failure and records an error

---

Submitted by: **The Wroclaw Institute of Spatial Information and Artificial Intelligence**
(WIZIPISI — Wrocławski Instytut Zastosowań Informacji Przestrzennej i Sztucznej Inteligencji)

### DIFF
--- a/io/plugins/eu.esdihumboldt.hale.io.geopackage.test/src/eu/esdihumboldt/hale/io/geopackage/GeopackageInstanceWriterTest.groovy
+++ b/io/plugins/eu.esdihumboldt.hale.io.geopackage.test/src/eu/esdihumboldt/hale/io/geopackage/GeopackageInstanceWriterTest.groovy
@@ -1,4 +1,4 @@
-/*
+﻿/*
  * Copyright (c) 2020 wetransform GmbH
  * 
  * All rights reserved. This program and the accompanying materials are made
@@ -16,6 +16,7 @@
 package eu.esdihumboldt.hale.io.geopackage
 
 import static org.junit.Assert.assertEquals
+import static org.junit.Assert.assertFalse
 import static org.junit.Assert.assertTrue
 
 import java.nio.file.Files
@@ -26,6 +27,7 @@ import java.util.function.Consumer
 
 import javax.xml.namespace.QName
 
+import org.junit.Assume
 import org.junit.Test
 import org.locationtech.jts.geom.Geometry
 import org.locationtech.jts.geom.Point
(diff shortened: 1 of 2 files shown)


stempler (commented):
Thanks a lot for the contribution. Looks good to me so far, now waiting for the workflow runs.
Please make sure to modify the commit message to follow conventional commits (`fix:` commit)

PR #1513, opened 2026-06-22
Reduce log level for noisy XML schema import messages

### LINKED ISSUE #965: Reduce log level to debug for some info messages on xml schema import
Version 4.2.0.SNAPSHOT

Build 202203231441
Revision 1a25d6c351035ec4c3e162fdf7db4dcc1ee4d07c

When importing an xml schema (for example INSPIRE presets), a lot of info messages ("Particle that allows any element is not supported" and "White space facet not supported") are logged and displayed in the properties of the schema import report 
grafik

The same is true for logs of CLI transformations where such messges lead to the logs being longer than needed.

In both cases, these messages tend to confuse users. As long as they do not have an impact on the usablity of a schema, their log level should therefore be reduced to debug.

adamnadolny-wizipisi:
## Summary

Two log messages during XML schema import were emitted at INFO level, causing excessive noise in the log for valid schema constructs that the reader simply skips:

- Unsupported white space facet ? reduced to DEBUG
- Unsupported XmlSchemaAny particle ? reduced to DEBUG

These messages appear for every occurrence and provide no actionable information for most users; debug level is more appropriate.

Fixes #965
Submitted by: The Wroclaw Institute of Spatial Information and Artificial Intelligence
(WIZIPISI — Wrocławski Instytut Zastosowań Informacji Przestrzennej i Sztucznej Inteligencji)

### DIFF
--- a/io/plugins/eu.esdihumboldt.hale.io.xsd/src/eu/esdihumboldt/hale/io/xsd/reader/XmlSchemaReader.java
+++ b/io/plugins/eu.esdihumboldt.hale.io.xsd/src/eu/esdihumboldt/hale/io/xsd/reader/XmlSchemaReader.java
@@ -1111,8 +1111,7 @@ public class XmlSchemaReader extends AbstractSchemaReader {
 		}
 		else if (particle instanceof XmlSchemaAny) {
 			// XXX ignore for now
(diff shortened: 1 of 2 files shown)


stempler (approved):

stempler (changes requested):
Turns out there is no debug level for reporters, so the build fails with compilation errors.

